PR #232, opened 2014-06-10
Previously Published URL Handler implementation

davidjgonzalez:
When dealing w very large numbers of previously published URLs that need to be redirected to the appropriate pages, sling:vanityPaths can reach limits.

This implementation is intended to be hooked into the 404 error page JSP .. ex impl integrated w the ACS EPH

```
<%@page session="false" import="com.adobe.acs.commons.legacyurls.LegacyURLHandler"%><%
LegacyURLHandler legacyURLHandler = sling.getService(LegacyURLHandler.class);

if(!legacyURLHandler.doRedirect(slingRequest, slingResponse)) { 
       // Returned false meaning no legacy URL match could be found
       %><%@include file="/apps/acs-commons/components/utilities/errorpagehandler/404.jsp" %><%
}
%>
```

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/legacyurls/PreviouslyPublishedURLManager.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/legacyurls/PreviouslyPublishedURLManager.java
@@ -1,0 +1,13 @@
+package com.adobe.acs.commons.legacyurls;
+
+import org.apache.sling.api.resource.Resource;
+import org.apache.sling.api.resource.ResourceResolver;
+
+import javax.jcr.RepositoryException;
+
+public interface PreviouslyPublishedURLManager {
+    public void create(ResourceResolver resourceResolver, String path, String... urls) throws RepositoryException;
+    public void update(ResourceResolver resourceResolver, String path, String... urls) throws RepositoryException;
+    public void delete(ResourceResolver resourceResolver, String... urls) throws RepositoryException;
+    public Resource find(ResourceResolver resourceResolver, String url);
+}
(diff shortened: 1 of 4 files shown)


davidjgonzalez:
Even on AEM6/Oak and @asano's enhancements, the powers-that-be apparently don't feel Sling Vanities are the proper tooling for this use case and prefer an approach similar to this.

The determination between Sling Vanities and these urls would be:

* Sling Vanities are marketing short URLs
* Previously Published URLs are just that; URLs that have no intrinsic value other than capturing long-tail traffic.

With the luxury of hindsight, not trying to leverage Sling Vanities for this use-case would've saved a lot of time and consternation in this particular instance.

@justinedelson WDYT about re-opening?

justinedelson:
@davidjgonzalez Sorry, I thought I commented on this yesterday, but I guess I didn't actually hit "Comment" before shutting down. Anyway...

I'm still worried about two things:
* Although I totally hear you that the intention is different than vanities, the implementation (setting a property on the page/asset) would still result in (potentially) authorable content.
* While I see that the fact that the query is only executed on 404, that could still be a lot of queries.

So... WDYT about restructing this around a separate content structure, say /etc/acs-commons/legacyUrls. And then doing a straight node lookup. For example, if there was a request for /oldurl/which/doesnt/exist.html, then we look for /etc/acs-commons/legacyUrls/oldurl/which/doesnt/exist.html and then look for a target property on that node.

Assuming the use of Oak and sling:Folder or oak:Unstructured, flat structures should be mostly OK.

davidjgonzalez:
@justinedelson Agree a lookup table is a good approach - i actually started building that: https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/232/files#diff-e096e9195012574ab18faec04ef60a4dR34

Totally unscientific, but AEM6 Queries should be fast since the index will have ~1 entry per key; AFIACT approx the same as creating a custom node-based map.

Maintaining the custom map has the downside of requiring extra tooling to re/create/refresh/maintain it (ex if the Event Handler responsible for populating it is disabled somehow; there is missing data).

I'm not as worried of having the content be technically "author-able"; we hide a lot of system data on Page nodes that is ACL-modifiable by authors; we just say don't make a dialog to edit it.

Let me revisit this given time/competing features. Just didn't want spend any more time if you were still strongly opposed.

Issue #102, opened 2014-01-16
Replication Status Manager


davidjgonzalez:
Large content migrations are often best done via packages or other means that don't mark resources in AEM as replicated. This can cause confusion/problems/inefficiencies w activation and deactivation.

It would be nice to have a way to mark all these resources as activated (without doing actually activation the resources - i'm looking at your Tree Activate).

v1 could include:
- Selecting trees to mark as replicated

v2 could include:
- Selecting package's whose coverage to mark as replicated.

PR #234, opened 2014-06-12
Package Replication Status Updater (Previously #188)

davidjgonzalez:
Cleaned up #188 commit history; Based on discussions in #102

Replication Status Updater (Activate/Deactivate/Clear) with Package Activation event handler.

ReplicationStatusManager service used to update Replication state of resources. Only resources whose nodeTypes are of the osgi configured node-types are candidates

node-types defaults to: cq:ReplicationStatus, rep:User, re
p:Group, cq:PageContent, dam:AssetContent (the last 2 should actually always match based on cq:ReplicationStatus nodeType check)

JcrPackageReplicationStatusEventHandlerImpl event handler listens for replication events on JcrPackage resources (~ `/etc/packages/.+\.zip`)

Currently get the package contents via package "coverage" (though I think this is different than "contents" which is what is preferable; need to investigate this some more).

OSGi configurable: 
- replicated-at: Current Time or Package Last Modified  
- replicated-by (Defaults to: Package Replication)
- node-types: List of nodetypes that are candidates for to update with replication status

Example folder (/content/dam/geometrixx/banners) that was replicated via package. Event handler was configured to `replicate-at: CURRENT_TIME` 

aem_wcm___banners

adobe_experience_manager_web_console_-_configuration


davidjgonzalez:
Documented here: http://adobe-consulting-services.github.io/acs-aem-commons/features/package-replication-status-updater.html

PR #364, opened 2014-11-19
Adding letterboxing & pillarboxing when aspect ratios of original and final image are different

ramyareddyg:
Adding letterboxing and pillarboxing accordingly to the image when the expected final image aspect ratio is not same as the original image aspect ratio


davidjgonzalez:
@ramyareddyg @bstopp did we decide this should be a different Transformer from the original Resize? I think that makes sense opposed to 1) changing behavior of the original  2) packing too much logic into the Resize transformer.

bstopp:
Getting back into this now: It looks like there's no way to turn _off_ the letter/pill boxing. I agree @davidjgonzalez, a subclass of the resize makes more sense; ensuring backwards compatibility with in-the-wild use cases.

davidjgonzalez:
Closed via #563

PR #396, opened 2014-12-16
Feature/version comparator

steffenrosi:
Hey guys,

this is a UI to support comparing versions of a page. It indicates things like property changed/added/removed, resource added/removed etc. 

There is still some brushing/documenting/writing tests necessary but I would like to get some initial feedback before doing so.

It can be easily tested by navigating to: http://localhost:4502/etc/acs-commons/versioncompare.html?path=/content/geometrixx/de/products/jcr:content
Before doing so, make sure you create a few versions with changes for the page under /content/geometrixx/de/products

Thanks,
Steffen


davidjgonzalez:
@steffenrosi SM only started shipping w AEM 6 right?

If SM 1.1.0 isn't installed, i assume this page crashes out? If it does we should a check to see if AEM has the SM 1.1.0 or later; if they dont, display a message saying "Upgrade to SM 1.1.0+ !!!" 

@justinedelson wdyt? Don't think we've used SM in Commons/Tools yet.

justinedelson:
Thanks for this @steffenrosi . It is very cool.

In terms of the Sling Models dependency, let's just document that 1.1.0 is required (which is there in 6.1 by default)

Issue #139, opened 2014-03-03
Long form text Component


davidjgonzalez:
A component whose purpose is to facilitate the creation and management of long form text on a page, while allowing said long form content to be enriched via CQ Components.

Have something very POC; i like the idea, but good parsys "injection" and component layout may be hard to achieve. Also, using a good inpage WYSIWIG may be preventative due to lack of CQ-specific tooling.

![one_parsys-4](https://f.cloud.github.com/assets/1451868/2306422/f85df480-a284-11e3-8bbb-a3ef9e7eac7b.png)

* Didn't like "Edit components" being default as too many parsys drops zones clutter the page and make it difficult to realize the content.

![one_parsys-6](https://f.cloud.github.com/assets/1451868/2306441/68c9a94e-a285-11e3-8731-fb6f19b98f5e.png)
![one_parsys-3](https://f.cloud.github.com/assets/1451868/2306446/a1ad4e78-a285-11e3-8c5b-9abef8a887ff.png)
![one_parsys-6](https://f.cloud.github.com/assets/1451868/2306451/da2fb3f8-a285-11e3-81b8-142182fd1940.png)

Also needs work handling increase/decrease of parsys injections points based on text content. There are number of ways to handle this, havent given it too much thought to determine the best route (and generally against the clean-up collapse of nodes of GET approach).

https://github.com/davidjgonzalez/acs-aem-commons/tree/feature/long-form-text

davidjgonzalez:
Closing to prefer the merged textedit paradigm as this is what's used throughout the product

PR #408, opened 2015-01-27
Long form text component

davidjgonzalez:
Simplified implementation of #139 

Add the ACS Commons Long Form Text component
long_form_text_demo

Enter long form text into RTE
long_form_text_demo

LFT Component will split the RTE contents on the top HTML elements (p tags) and inject parsys between them
long_form_text_demo

Add Components in between the paragraphs, but manage the text content in its "long form"
long_form_text_demo


davidjgonzalez:
@justinedelson could you take a look at https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/408/files#diff-d43214f8e2d9de42dc919426c2e2bd6aR41 .. and sanity check the HTML > XHTML parsing?

bundle/src/main/java/com/adobe/acs/commons/components/longformtext/LongFormTextComponent.java
```diff
@@ -0,0 +1,40 @@
+package com.adobe.acs.commons.components.longformtext;
```
justinedelson: license header?

bundle/src/main/java/com/adobe/acs/commons/components/longformtext/impl/LongFormTextComponentImpl.java
```diff
@@ -0,0 +1,192 @@
+package com.adobe.acs.commons.components.longformtext.impl;
```
justinedelson: license header?

bundle/src/main/java/com/adobe/acs/commons/components/longformtext/LongFormTextComponent.java
davidjgonzalez: ah yes; needs ProviderType too

content/src/main/content/jcr_root/apps/acs-commons/components/content/long-form-text/dialog.xml
```diff
@@ -0,0 +1,34 @@
...
+        activeTab="0"
+        title="Long Form Text"
+        xtype="tabpanel">
+    <items jcr:primaryType="cq:WidgetCollection">
+        <tab1
+                jcr:primaryType="cq:Widget"
+                title="Text"
+                xtype="panel">
+            <items jcr:primaryType="cq:WidgetCollection">
+                <long-form-text
```
justinedelson: Is there any way (or reason?) to get this to use the generic text configuration feature? Otherwise, users will need to override this dialog whenever they want different RTE options.

davidjgonzalez: @justinedelson not sure i follow; We can get rid of the "rtePlugins" node if thats what youre saying.

bundle/src/main/java/com/adobe/acs/commons/components/longformtext/impl/LongFormTextComponentImpl.java
davidjgonzalez: fixed

content/src/main/content/jcr_root/apps/acs-commons/components/content/long-form-text/dialog.xml
justinedelson: Basically, I'm suggesting that this be replaced with

{code}
<long-form-text
                    jcr:primaryType="cq:Widget"
                    script="rte.long-form.json.jsp"
                    xtype="slingscriptinclude"/>
{code}

so that if a site wanted to define their own RTE Plugins, they could use http://adobe-consulting-services.github.io/acs-aem-commons/features/dynamic-rte-configurations.html and create a node at `/etc/rteconfig/mysite/long-form`

justinedelson: Otherwise, anyone wanting to use this component *and* a different set of plugins would have to extend the component.

davidjgonzalez: Fixed; FWIW I think consumers of our components should ALWAYS extend them with a simple subType "shell" at the very least. This future proofs their content allowing it to be bound to their own flexible impl (they own their impls; changing ACS Commons components requires extra-organizational consensus) and reduces lock-in to ACS Commons

PR #416, opened 2015-02-12
Fix for double quotes in richtext boxes

brennhill:
Richtexts in a multifield widget break when using double quotes.

### DIFF
--- a/content/src/main/content/jcr_root/apps/acs-commons/widgets/source/js/multipanel.js
+++ b/content/src/main/content/jcr_root/apps/acs-commons/widgets/source/js/multipanel.js
@@ -72,7 +72,7 @@ ACS.CQ.MultiFieldPanel = CQ.Ext.extend(CQ.Ext.Panel, {
                 return;
             }
 
-            pData[i.key] = i.getValue();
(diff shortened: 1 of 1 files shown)


content/src/main/content/jcr_root/apps/acs-commons/widgets/source/js/multipanel.js
```diff
@@ -72,7 +72,7 @@ ACS.CQ.MultiFieldPanel = CQ.Ext.extend(CQ.Ext.Panel, {
                 return;
             }
 
-            pData[i.key] = i.getValue();
+            pData[i.key] = i.getValue().replace(/"/g, '&quot;');
```
davidjgonzalez: Do you know if `i.getValue()` is always guaranteed to be a string (or an obj w a `.replace` method).

Otherwise looks good! Thanks!

brennhill: I've updated the code to do a thorough string check and only apply the
replace if it is a string.

-- Sent from my phone, appologies for thumb flubs and bad auto correct.

davidjgonzalez:
Replaced by #418 which allows code to pass JSlint tests

PR #418, opened 2015-02-15
Defect/416 fix for double quotes in rte

davidjgonzalez:
Fixed code to pass JSLint tests ..


djessup:
This patch seems to have caused the regression mentioned in #469

It's not clear to me exactly what is being achieved by this patch. The title mentions an issue with double quotes - is there an example available of the problem being solved? I suspect is may need some changes to either the regex or test conditions for when it is applied.

PR #423, opened 2015-02-27
Base (Sightly) AJAX component

joemaffia:
Sightly version for turning “normal” components into components that
are pulled into the page via AJAX.


justinedelson:
Why is this necessary? A component extending the existing ajax component can be written in Sightly, just by creating the script as ajax.html instead of ajax.jsp.

justinedelson:
See related thread here: http://sling.markmail.org/thread/5sxsywvqhjmdjto2

joemaffia:
@justinedelson that was my initial thought but when I tried wasn't really working so that's the reason I created the Sightly version.

Changed instance and gave it another shot and seems fine

PR #427, opened 2015-03-07
425-bulk workflow manager - auto-resume enhancements

davidjgonzalez:
Enhancement to auto-resume feature;

1) Keeps backwards compatible behavior of auto-resuming.
2) Uses a tree-traversal to walk the Pages under /etc/acs-commons/bulk-workflow-manager to find resumable BWM pages (it does NOT visit the potentially very large batch structure under each page); This traversal is very fast as its walk of a isolated Page tree (which should never get large enough to be problematic).

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/workflow/bulk/BulkWorkflowEngine.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/workflow/bulk/BulkWorkflowEngine.java
@@ -35,6 +35,8 @@ public interface BulkWorkflowEngine {
 
     String SLING_FOLDER = "sling:Folder";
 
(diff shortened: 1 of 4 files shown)


bundle/src/main/java/com/adobe/acs/commons/workflow/bulk/BulkWorkflowEngine.java
```diff
@@ -35,6 +35,8 @@
 
     String SLING_FOLDER = "sling:Folder";
 
+    String BULK_WORKFLOW_MANAGER_PAGE_FOLDER_PATH = "/etc/acs-commons/bulk-workflow-manager";
```
justinedelson: Not a big deal, but does this constant need to be in the API? Would allow avoiding the export version bump.

davidjgonzalez:
@justinedelson sure; thats a problem - fixed

PR #430, opened 2015-03-19
Parameterized (De)Activate Workflow Steps

bstopp:
Adding new feature for parameterized workflow activation steps; Giving the ability to select replication agents instead of triggering a mass activation processes.

We've seen the use case where clients have a preview environment which needs content replication without triggering a default agent. These allow the workflows to have a subset of replication agents be triggered, and subsequent approval step executed, before triggering another subset of agents (this time for the production publish systems).

It's possible to get the arguments correct, on a normal WF process step. But to make life easier for WF editors, i included three custom WF Components. 

* Activate
* Flush
* Deactivate

The Activate/Deactivate only show those replication agents which are *not* cache invalidators. Of course the flush will only list agents which *are* invalidators.

I used an inheritance model to manage not doing a CPD on the dialog, JSON generation; but that means that the dialog titles and other attributes aren't unique to the different types. Let me know if you think it's better to be explicit.

The thread local variables are use because I am pretty sure that the processes can be used by multiple threads; and I needed to keep state (which agents were selected).

Here's some screen shots:

capture

capture

If it's accepted, i'll write up the documentation and do a PR on that repo.

### DIFF
--- a/bundle/pom.xml
+++ b/bundle/pom.xml
@@ -1,8 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project
-    xmlns="http://maven.apache.org/POM/4.0.0"
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd ">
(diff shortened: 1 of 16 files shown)


justinedelson:
In general, I think this looks good. My only real requests are to:
* Minimize/eliminiate the non-relevant formatting changes in the pom files.
* Ensure that that the new dependency versions are correct for 5.6.1

bstopp:
Reverted changes to poms as requested. I think I now have the correct versions of libraries based on 5.6.1.

PR #434, opened 2015-04-06
CQInclude Property Namespace Servlet

davidjgonzalez:
the cqinclude xtype allows for re-use of other dialogs. This supports a nice pattern of composition when creating components; Larger components cant be composed of smaller components, and their editing UI can be centralized into the dialog of the parent component.

OOTB cqinclude is "dumb" (in that it just proxies the data through, and doesnt apply any transformations, etc.) in that renders the sub-component's dialog JSON exactly as it is defined on the sub component. This is problematic when a Component is composed of multiple smaller sub-components and each sub-component uses the same name properties.

For Example a Feature component that is composed of a Title Component, Text Component and Button component. Both the Title component and Button component may store the label value to the property name "./jcr:title". Since both components save to the same property, this is problematic.

This CQ Include Name-Prefixing Servlet allows a "prefixing" node name to be provided during the cqinclude. This servlet is invoked by the addition of the selectors

```
cqinclude.namespace.<namespace>
```

The include for Title component dialog may look like
```
jcr:primaryType=cq:Widget
path=/apps/acme/components/title/dialog/items/tab.cqinclude.namespace.title.json
xtype=cqinclude
```

The include for Button component dialog may look like
```
jcr:primaryType=cq:Widget
path=/apps/acme/components/button/dialog/items/tab.cqinclude.namespace.button.json
xtype=cqinclude
```

The dialog would then read/save the jcr:title from the respective includes as follows

```
./title/./jcr:title
./button/./jcr:title
```

cleanly separating the two.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/json/AbstractJSONObjectVisitor.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/json/AbstractJSONObjectVisitor.java
@@ -1,3 +1,23 @@
+/*
+ * #%L
+ * ACS AEM Commons Bundle
+ * %%
+ * Copyright (C) 2015 Adobe
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


justinedelson:
Rather than using OptingServlet, would it make more sense to have the servlet just return a 404 when the right selector isn't found?

davidjgonzalez:
@justinedelson sure - sounds reasonable.

davidjgonzalez:
Ready to merge if there is no further feedback

PR #443, opened 2015-04-15
add support for QueryBuilder-map based queries in Query Packager

justinedelson:
This is a fairly minor change which adds support to the Query Packager for QueryBuilder queries in key=value form. The key advantage of this is that QueryBuilder supports relative date range predicates (via http://docs.adobe.com/docs/en/cq/5-6-1/javadoc/com/day/cq/search/eval/RelativeDateRangePredicateEvaluator.html) something which if done just with XPath or SQL-2 would require constant updating of the query.


davidjgonzalez:
LGTM! Great addition - QB FTW!

PR #445, opened 2015-04-19
Property Merge POST Processor

davidjgonzalez:
A commons use case is collecting tags via multiple inputs, but using the collection of tags as search criteria. Property Merge allows data to be collected into a set of discrete properties, and then via an event handler be merged into a single property.

```
./primaryTags@PropertyMerge=collectedTags
./secondaryTags@PropertyMerge=collectedTags
<sourceProp>@PropertyMerge=<destinationProp>
```

This implementation always assumes String and does NOT allow duplicates into the destination property


davidjgonzalez:
Am thinking writing should be a job to guarantee work is done. Wdyt about leaving work checks into the event to prevent jobs w no actual work to do being popped. 

 jobs seem a little disingenuous since most often we spawn from events which themselves have no guarantees.

davidjgonzalez:
@justinedelson WDYT about the Post Processor?

Could think of a clean way off the top of my head to pass in Types or Allow Dupes;  Maybe ..

```
<destinationProp>@PropertyMerge.AllowDuplicates=true
<destinationProp>@PropertyMerge.TypeHint=String
```
Not sure how idiomatic the dot notation is for them..?

davidjgonzalez:
```
:aTags@PropertyMerge=abcTags
:bTags@PropertyMerge=abcTags
:cTags@PropertyMerge=abcTags
:absTags@PropertyMerge.AllowDuplicates=true
:absTags@PropertyMerge.TypeHint=String

:dTags@PropertyMerge=deTags
:eTags@PropertyMerge=deTags
:deTags@PropertyMerge.AllowDuplicates=false
:deTags@PropertyMerge.TypeHint=String
```
Adding the above as request params will merge `aTags`, `bTags`, `cTags` into `abcTags` and `dTags`, `eTags` into `deTags`

Note the `:` prefix to ensure these params are ignored and not written to the underlying resource.

@justinedelson WDYT? Look ok?

bundle/pom.xml
```diff
@@ -461,11 +479,16 @@
...
         <dependency>
-            <groupId>org.apache.sling</groupId>
-            <artifactId>org.apache.sling.commons.scheduler</artifactId>
-            <version>2.3.4</version>
-            <scope>provided</scope>
+            <groupId>com.adobe.aem</groupId>
+            <artifactId>aem-api</artifactId>
+            <version>6.0.0.1</version>
         </dependency>
+        <dependency>
```
justinedelson: Two things here:
1) Did you mean to add this dependency twice?
2) AFAICT, this means we are breaking 5.6.1 compatibility. Is this necessary? My preference would be to cut 1.10.0 with 5.6.1 compatibility and then move forward with 2.0 requiring 6.0 as the minimum. WDYT?

davidjgonzalez: @justinedelson let me double check this; i dont think this PR should be dependent on 6.0...

davidjgonzalez:
@justinedelson should be good (removed the aem-6 pom dep); any other thoughts?

PR #448, opened 2015-04-21
Feature/wcm views

davidjgonzalez:
AEM pages can be very customizable with 5's of drop-zones (column controls anyone!) and 10s of components. Each of these editable aspects results in drop-zones or hovers. This can result in an overwhelming experience for AEM Authors (and sometimes slow depending on the browser/client machine!)

WCM Views is a feature that allows sets of Components' (WCMMode) Edit-ability to be toggled on and off on the Page via the Sidekick. The sets are customizable by developers by applying a @wcmViews (String[]) property to EITHER a cq:Component node or directly on a content resource. WCM Views will auto-aggregate the WCM Views available on the page to populate the Sidekick. For WCM Views that cant be easily derived form walking the Page and looking for @wcmViews, the WCM Views Servlet allows for WCM Views to be specified by path (ex. for Inherited Parsys which may not have a resource on the page itself)

WCM Views has a sidekick option to completely display the WCMViews as well as only enable components for Editing that do not have an associated WCM Views (this is called Default View).

wcm-views-test

wcm-views-test

wcm-views-test

wcm-views-test


bstopp:
This looks sick. I'd like to work help with seeing how to make it work using the layers on the Granite/Cooey/Touch UI (whatever you guys are calling it these days).

davidjgonzalez:
Cooey is right!  I was looking at this .. seems like it should be easy enough to make a new Button that has a pop-down w the list of WCM Views..

http://goo.gl/KEPBgH > New Page Action

PR's welcome!

davidjgonzalez:
Capturing 2 ideas from @bstopp 
* Persist the selected wcm-view in a cookie (ala wcmmode)
* Make user group-based views

davidjgonzalez:
@justinedelson you good w/ this? I checked that the SetInterval waiting for the sidekick to become available is necessary.

A TouchUI dropdown can be added in the future as an enhancement.

Issue #458, opened 2015-05-20
Extjs IFramed Dialogs in TouchUI submit issue


andre-inside:
Since installation of acs commons 1.10.0 touchui dialogs which have embedded ExtJS dialog in it can't be submitted anymore (request is issued, but the screen is stuck in the "pending submit" state). How to reproduce:
- Create new geometrixx page
- Open the page in touchui
- Add any component which has no touchui dialog yet (for instance sitemap -> foundation/components/sitemap)
- Open the dialog
- Submit or Cancel -> JS Error -> "Uncaught TypeError: Cannot read property 'addEventListener' of null"

justinedelson:
@schoudry - can you look into this? I would like to hold off on publicizing the release until this can be addressed.

schoudry:
checking....

PR #459, opened 2015-05-20
Extjs IFramed Dialogs in TouchUI submit issue #458

schoudry:


justinedelson:
Does this mean you were able to reproduce the issue?

schoudry:
yes, i think form.cq-dialog is not created when dialog has extjs widgets....

and i am not sure if this is a product bug (or it happened because of some UI extensions-in-development on my cq ) but when i set rootpath in sitemap component and submit dialog, ./rootPath: passed to CQ is empty... (with/without the /apps/acs-commons/touchui-widgets )

justinedelson:
OK. Thanks.

PR #471, opened 2015-06-18
Update codebase to require 6.0 as a minimum platform. Fixes #454

### LINKED ISSUE #454: Remove support for 5.6.1
I think after the release of 1.10, we should remove support for 5.6.1. This primarily means updating dependencies. To make this obvious, the version after 1.10 (modulo bug fixes if necessary) would be 2.0.0.

We should also add a page to the website which lists each version of ACS AEM Commons and the minimum version of AEM required.

justinedelson:
Includes:
* Updating dependencies
* Adding additional method in FileImporter due to updated API
* Fixing minor deprecation issues (more to come)
* Including code from http://www.randelshofer.ch/blog/ as AEM 6.0 uberjar does not include this code (test-scope only)
* Removed two optional imports as they are included in 6.0

Confirmed that all tests pass and package is installable on 6.0 GA.


justinedelson:
Sorry I pushed this to this repo instead of my fork, but it's a PITA to change that now.

justinedelson:
Adios 5.6.1. It's been a fun ride...

Issue #475, opened 2015-06-25
Provide a way to automatically convert multifieldpanel widgets using the Conversion tool


justinedelson:
Using the tool at http://docs.adobe.com/docs/en/aem/6-1/develop/dev-tools/dialog-conversion.html, it should be possible to convert multifieldpanel widgets.

PR #476, opened 2015-06-25
Create a dialog conversion rule for handling multifield panel. fixes …

justinedelson:
…#475. also includes a converted dialog for the definition list component.

Caveat - after conversion, the name properties need to be manually set to the value of the key property on the old widgets.

### DIFF
--- a/bundle/pom.xml
+++ b/bundle/pom.xml
@@ -51,6 +51,7 @@
                         <Bundle-SymbolicName>com.adobe.acs.acs-aem-commons-bundle</Bundle-SymbolicName>
                         <Import-Package>
                             org.scribe*;version="[1.3.0,2)",
+                            com.adobe.cq.dialogconversion;resolution:=optional,
                             *
                         </Import-Package>
                     </instructions>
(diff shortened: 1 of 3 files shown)


justinedelson:
The results seem OK, with the one caveat listed. But the UI seems "squished" to me.

@schoudry can you look at this screenshot and let me know if something else is needed.

converted-dialog

schoudry:
@justinedelson the dialog looks ok when full screen

image

if shown modal, i guess thats the way it is ootb, if we add a custom style with width set, it looks better (the description textarea)

image

justinedelson:
I just noticed on https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/source/touchui-composite-multifield.js#L62 that you have a class="full-width". Is this defined anywhere in our codebase or is it custom?

I imagine it would be common to need this, so if not, perhaps we should add it.

justinedelson:
Or, better yet, would it make sense to just add

    [data-acs-commons-nested] .coral-Form-field {
        width: 100%;
    }

schoudry:
the full-width class is ootb and it exists in 60 sp2 (eg. /libs/foundation/components/list/cq:dialog/content/items/column/items/wellSetStatic/items/pages)... couldn't find it in 61 (so i guess will be ignored if the dialog is opened on 61)

Regd. adding a style, may be its ok to add one, or i guess leave the cosmetic thing to user and leave it as a suggestion, to keep the extension logic minimum possible, not sure....

PR #478, opened 2015-06-26
HttpClientFactory (Review)

kalyanar:
@justinedelson / @davidjgonzalez 

I have created a version based on your inputs. Please review.
Please let me know the test cases I need to write.

### DIFF
--- a/bundle/pom.xml
+++ b/bundle/pom.xml
@@ -88,6 +88,27 @@
 
     <dependencies>
         <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>fluent-hc</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.osgi</groupId>
             <artifactId>org.osgi.compendium</artifactId>
         </dependency>
@@ -477,6 +498,11 @@
                     <groupId>javax.servlet.jsp</groupId>
                 </exclusion>
             </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+            <version>4.3.5</version>
         </dependency>
     </dependencies>
 
--- a/bundle/src/main/java/com/adobe/acs/commons/http/JSONResponseHandler.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/http/JSONResponseHandler.java
@@ -1,0 +1,68 @@
+/*
+ * #%L
+ * ACS AEM Commons Bundle
+ * %%
+ * Copyright (C) 2013 - 2015 Adobe
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+package com.adobe.acs.commons.http;
+
+import org.apache.http.HttpResponse;
+import org.apache.http.client.ClientProtocolException;
+import org.apache.http.client.ResponseHandler;
+import org.apache.sling.commons.json.JSONException;
+import org.apache.sling.commons.json.JSONObject;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.Charset;
+
+/**
+ * Converts response to json
+ *
+ * @param <T>
+ */
+public class JSONResponseHandler<T> implements ResponseHandler<JSONObject> {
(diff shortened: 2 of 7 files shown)


justinedelson:
I'll do more of a code review after vacation, but generally this looks good. Just a few comments:

* is Executor thread safe? I assume not, in which case a new instance should be returned from each call to getExecutor()?
* the parameter to the various methods which return a Request should be named path, as the value gets appended to the base URL to get the full URL.
* please check which of these dependencies come OOTB (with 6.0 and 6.1 both). Anything which isn't OOTB needs to be optionally imported. And we should provide a content package in a separate project for convienence.

kalyanar:
@justinedelson I was googling and as per https://issues.apache.org/jira/browse/HTTPCLIENT-1437, the executor is thread safe. I thought that for the same domain, one executor can be shared. Each executor allows 200 concurrent requests by default. 

I shall fix other items. I shall fix the executor threadsafe issue after your further inputs.

bundle/src/main/java/com/adobe/acs/commons/http/impl/HttpClientFactoryImpl.java
```diff
@@ -0,0 +1,128 @@
...
+import org.apache.http.HttpHost;
+import org.apache.http.client.fluent.Executor;
+import org.apache.http.client.fluent.Request;
+import org.apache.sling.commons.osgi.PropertiesUtil;
+
+import java.util.Map;
+
+
+@Component(
+        label = "Http Components Fluent Executor Factory",
```
davidjgonzalez: label = "ACS AEM Commons - Http Components Fluent Executor Factory",

bundle/src/main/java/com/adobe/acs/commons/http/impl/HttpClientFactoryImpl.java
```diff
@@ -0,0 +1,128 @@
...
+        description = "Http Components Fluent Executor Factory", immediate =
```
davidjgonzalez: immediate is false by default, can leave off

bundle/src/main/java/com/adobe/acs/commons/http/impl/HttpClientFactoryImpl.java
```diff
@@ -0,0 +1,128 @@
...
+        false,
+        metatype = false, configurationFactory = true, policy =
```
davidjgonzalez: We dont want `metatype=true` ? Seems like might be handy for developers to use. Ultimately these configs should be deployed via osgiConfigs, but does it hurt to allow config via Felix?

bundle/pom.xml
```diff
@@ -478,6 +499,11 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
```
davidjgonzalez: This looks like a duplicate dependency?

pom.xml
```diff
@@ -207,6 +207,32 @@
     <dependencyManagement>
         <dependencies>
             <dependency>
+                <groupId>org.apache.httpcomponents</groupId>
```
davidjgonzalez: Are all these dependencies provided by core AEM bundles? Im not sure that they are...
Ex. My AEM 6.0 cannot find locate a bundle that exports `org.apache.http.client.fluent.Request` via depfinder..

The jackson stuff seems to be exported via a AEM Social UGC bundle (TBH the way soco iterates i wouldn't be surprised if that export disappeared in a few versions; almost feels like spillage.)

bundle/src/main/java/com/adobe/acs/commons/http/JSONResponseHandler.java
```diff
@@ -0,0 +1,68 @@
+/*
```
davidjgonzalez: Any chance you can reformat using the projects formatting? It looks like youre at 80 char breaks which makes it rather hard to read.

bundle/src/main/java/com/adobe/acs/commons/http/impl/HttpClientFactoryImpl.java
```diff
@@ -0,0 +1,128 @@
...
+    private static final String USE_SSL = "use.ssl";
+
+    @Property(label = "Disable certificate check", description = "If selected" +
+            " it will disable certificate check for the SSL connection",
+            boolValue = false)
+    private static final String DISABLE_CERTIFICATE_CHECK = "disable" +
+            ".certificate";
+
+    @Property(label = "username", description = "user name", value = "admin")
+    private static final String USERNAME = "sap.username";
```
davidjgonzalez: `sap.username` to `username` ?

bundle/src/main/java/com/adobe/acs/commons/http/POJOResponseHandler.java
```diff
@@ -0,0 +1,71 @@
...
+import java.nio.charset.Charset;
+
+/**
+ * Converts response to a pojo.
+ *
+ * @param <T>
+ */
+public class POJOResponseHandler<T> implements ResponseHandler<T> {
+
+    public static final Charset DEFAULT_CHARSET = Charset.forName("UTF-8");
```
davidjgonzalez: Where is this used?

bundle/src/main/java/com/adobe/acs/commons/http/POJOResponseHandler.java
```diff
@@ -0,0 +1,71 @@
...
+import java.io.InputStream;
+import java.lang.reflect.Type;
+import java.nio.charset.Charset;
+
+/**
+ * Converts response to a pojo.
+ *
+ * @param <T>
+ */
+public class POJOResponseHandler<T> implements ResponseHandler<T> {
```
davidjgonzalez: Should this be called `JSONToPOJOResponseHandler` ? It only supports Responses that are in a JSON format, right?

bundle/src/main/java/com/adobe/acs/commons/http/impl/HttpClientFactoryImpl.java
```diff
@@ -0,0 +1,128 @@
...
+
+@Component(
+        label = "Http Components Fluent Executor Factory",
+        description = "Http Components Fluent Executor Factory", immediate =
+        false,
+        metatype = false, configurationFactory = true, policy =
+        ConfigurationPolicy.REQUIRE)
+@Service
+public class HttpClientFactoryImpl implements HttpClientFactory {
+    @Property(label = "gateway domain", description = "gateway domain", value
```
davidjgonzalez: Thought this a bit odd that its called "Domain" when the example is an "IP" .. maybe better named "host" ?

justinedelson:
Replaced with #730

PR #479, opened 2015-06-28
Enhancement to ajax component: Add option to pass query parameters to…

b-Johnson:
Enhancement to ajax component: Add option to pass query parameters to the ajax request. Particularly useful for components (search) in which query parameters are require for functionality.


davidjgonzalez:
Replaced via #484

PR #484, opened 2015-07-27
Ajax Base Component - Query Parameter pass-through support

davidjgonzalez:
JS-centric version of #479 

Setting  `[cq:Component]@ajaxPassQueryParameters = true` on the Ajax'ified component will persist the QPs from the page request into a data-* attribute, which the AJAX GET will pick up and apply to the XHR call to AJAX in the component.

@B-Johnson @ice1138 ... WDYT?  Can you double check the encoding in the JSP and the decoding the JS .. think it's right; encodings sometimes confuse me though.

Thanks!

2015-07-26 at 10 00 pm


bstopp:
You should use [this](http://api.jquery.com/jQuery.param/) to define parameters, rather than encoding them yourself. 

Also, use the [cache option](http://api.jquery.com/jquery.ajax) on the get() call to do cache busting.

PR #503, opened 2015-08-16
Converting Oak Index Manager, Bulk Workflow Manager, Workflow Remover, and Version Comparison to CUI2

justinedelson:
I think this is all of it.

@davidjgonzalez please test. I think I got everything, but I may have missed something here and there.


davidjgonzalez:
@justinedelson will try to test the branch out tonight

justinedelson:
BTW, I realized I forgot to respond to some of your comments earlier - regarding using a template vs. DOM maniuplation I'm agnostic. I find the DOM maniuplation easier to understand, but ultimately this is an implemenetation detail of the directive.

Regarding directive vs. LESS mixin - list could probably be done as a mixin, but TBH once I got on the directive train, I couldn't get off :)

davidjgonzalez:
@justinedelson btw - the LESS stuff doesnt work well as mixins dont seem to be shared across clientlibs, i think we'd need to directly reference the CUI2 Less files in our css.txt which isnt a great pattern...

PR #516, opened 2015-08-26
Feature/Synthesized resource rendering

dherges:
is this useful to you?

Example: render HTML mails with dynamoc content, e.g.

``/apps/myapp/components/mail/mail.html``:
```html
<html>
<head><title>${properties.title}</title></head>
<body><p>${properties.message}</title></body>
</html>
```

Usage:

```java
Map<String, Object> props = new HashMap<String, Object>();
props.put("title", "Hello You");
props.put("message", "This is a mesage to you...");

String htmlResult = SynthesizedUtil.renderSynthesized("/path/that/doesnt-really-exist",
   "myapp/components/mail", props, request, response);
```


davidjgonzalez:
@dherges This is intriguing; It's essentially a way of creating a logic-ful templates.

Your e-mail template is pretty powerful IMO, as typically our email templates are static and we simply inject the vars. Using your creation, templates could respond to different passed params (via the properties).

bundle/src/main/java/com/adobe/acs/commons/synthesizedsling/SynthesizedSlingHttpServletRequest.java
```diff
@@ -0,0 +1,200 @@
...
+
+/**
+ * Used to build internal sling requests to do nice things :)
+ *
+ * @see com.adobe.acs.commons.forms.helpers.impl.synthetics.SyntheticSlingHttpServletGetRequest
+ */
+public class SynthesizedSlingHttpServletRequest extends SlingHttpServletRequestWrapper {
+
+    /** GET */
+    public static final String METHOD_GET = "GET";
```
davidjgonzalez: Are these METHOD_* vars needed? It doesnt seem like theyre used in the context of Synthesized ?

bundle/src/main/java/com/adobe/acs/commons/synthesizedsling/SynthesizedUtil.java
```diff
@@ -0,0 +1,75 @@
...
+import javax.servlet.RequestDispatcher;
+import javax.servlet.ServletException;
+import java.io.IOException;
+import java.util.Map;
+
+public final class SynthesizedUtil {
+
+    private SynthesizedUtil() {}
+
+    public static Resource buildSynthesizedResource(ResourceResolver resourceResolver, String path,
```
davidjgonzalez: Is there value in passing in a "path";  is there any advantage to providing it vs hard-coding it to something like "/var/acs-commons/synthesized" in this impl? 

Isnt it the properties and resourceType that matter?

bundle/src/main/java/com/adobe/acs/commons/synthesizedsling/SynthesizedUtil.java
```diff
@@ -0,0 +1,75 @@
...
+                                                    String resourceType, Map<String, Object> properties) {
+
+        SynthesizedResource synthesizedResource = new SynthesizedResource(resourceResolver, path, resourceType);
+        synthesizedResource.setValueMap(properties);
+
+        return synthesizedResource;
+    }
+
+    public static String renderSynthesized(String path, String resourceType, Map<String, Object> properties,
```
davidjgonzalez: Same question around path as above.

dherges:
@davidjgonzalez 

> This is intriguing; It's essentially a way of creating a logic-ful templates.

I do not understand. In an application, we use it as a way to render html output "from code". ``POST /services/subscription.json`` --> take some input parameters from request, set properties on a synthie resource --> render the ``GET 'synthie-resource'.html`` so that one can re-use sling scripts (components that already exist...)

templates don't need to be logic-ful ... and shouldn't really be ...

davidjgonzalez:
@dherges sightly is a logic-ful templating system, so is handlebars? lots of people seem to like and use them? we can skip this lively debate though :)

I understand the use case a little more w the above example... (POST > GET)

I read the original comment as making a HTML Email template (maybe it was the resource names)

justinedelson:
I think this looks like an interesting feature. A few comments:
* Can you add the license headers?
* And maybe some JavaDoc for the Util class?
* I suspect only the Util class should be exported. It looks like everything else should be in a .impl package.
* I'd suggest com.adobe.acs.commons.synth as a package name, rather than synthetizedsling.

dherges:
Added license header
Added some javadocs for the util class
Moved some stuff to the *.impl package
Added a demo servlet

However, it doesn't work with all components/sling script, e.g.

``curl 'http://localhost:4502/services/synthesizer?resourceType=wcm/foundation/components/title&p.title=abc&p.type=h2'`` results in

```*ERROR* [0:0:0:0:0:0:0:1 [1441210922006] GET /services/synthesizer HTTP/1.1] org.apache.sling.servlets.resolver.internal.SlingServletResolver Original error class io.sightly.java.api.SightlyUseException
io.sightly.java.api.SightlyUseException: org.mozilla.javascript.EcmaError: TypeError: Cannot read property "properties" from undefined (/libs/wcm/foundation/components/title/title.html#37)
	at io.sightly.js.loop.EventLoop.run(EventLoop.java:71
Caused by: org.mozilla.javascript.EcmaError: TypeError: Cannot read property "properties" from undefined (/libs/wcm/foundation/components/title/title.html#37)
	at org.mozilla.javascript.ScriptRuntime.constructError(ScriptRuntime.java:3687)
	at org.mozilla.javascript.ScriptRuntime.constructError(ScriptRuntime.java:3665)
```

whereas this works:

``curl 'http://localhost:4502/services/synthesizer?resourceType=myapp/components/content/iframe&p.title=abc&p.src=www.adobe.com'``

```html
<iframe src="www.adobe.com" alt="abc"></iframe>
```


Fairly expetrimental all the stuff ... :rocket:

bundle/src/main/java/com/adobe/acs/commons/synth/SynthesizerServlet.java
```diff
@@ -0,0 +1,43 @@
+package com.adobe.acs.commons.synth;
```
davidjgonzalez: @dherges is this intended as a HTTP endpoint for Synthesizer? Or is this the "demo servlet"?

If its the demo, can we remove, if not can we move to the impl package?

bundle/src/main/java/com/adobe/acs/commons/synth/SynthesizerServlet.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.sling.api.servlets.SlingSafeMethodsServlet;
+
+import javax.servlet.ServletException;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@SlingServlet(
+        paths = "/services/synthesizer"
```
davidjgonzalez: can you change to `/bin/acs-commons/synthesizer` all our other path-based servlets are registered under that common path prefix

bundle/src/main/java/com/adobe/acs/commons/synth/impl/SynthesizedSlingHttpServletRequest.java
```diff
@@ -0,0 +1,220 @@
...
+import org.apache.sling.api.request.RequestPathInfo;
+import org.apache.sling.api.resource.Resource;
+import org.apache.sling.api.wrappers.SlingHttpServletRequestWrapper;
+
+/**
+ * Used to build internal sling requests to do nice things :)
+ *
+ * @see com.adobe.acs.commons.forms.helpers.impl.synthetics.SyntheticSlingHttpServletGetRequest
+ */
+public class SynthesizedSlingHttpServletRequest extends SlingHttpServletRequestWrapper {
```
davidjgonzalez: After this gets merged, would be cool to replace use of com.adobe.acs.commons.forms.helpers.impl.synthetics.SyntheticSlingHttpServletGetRequest w this; looks a bit nicer!

davidjgonzalez:
@dherges can you checkout comment on the Servlet to see if we need to keep in ACS Commons project? Few minors, but otherwise LGTM! Thanks!

davidjgonzalez:
@dherges can you provide feedback on if we should include the servlet in ACS Commons? I can take care of the other feedbacks at merge. Thanks

dherges:
note to myself... need to take a look at this early in the office tomorrow ...

davidjgonzalez:
FYI - looking to cut release Sunday or Monday; would love to get this in!!! (That's why I'm pestering you ;))

davidjgonzalez:
@dherges ran out of time last weekend so targeting a release this weekend. If you want this in, please address the Qs wrt the servlet!

bundle/src/main/java/com/adobe/acs/commons/synth/SynthesizerServlet.java
dherges: more of a demo. commit coming.

dherges:
@davidjgonzalez finally, found some time :sunrise:

PR #522, opened 2015-09-01
Feature/jcrchecksumcompare

andrewmkhoury:
This is a tool for generating checksums for JCR nodes.  You can pass it a path and it will generate aggregate checksums (off its own name, properties, child node names and their properties).  By default it generates checksums for cq:PageContent and dam:AssetContent nodes.

The checksum output can be used to identify which content differs between multiple AEM instances.

By default it generates the checksums with the following options:
* *path* – the tree to generate hashes for
* *nodeTypes* – the node types to calculate hashes for.  Default node types are _cq:PageContent_ and _dam:AssetContent_
* *nodeTypeExcludes* - the node types to exclude from the calculation (when traversing the tree, node with these types, and their children would be skipped).  Default nodetype excluded is _rep:ACL_.
* *propertyExcludes* — names of properties to exclude from the hash calculation.  By default these are excluded:
```
jcr:created
jcr:createdBy
jcr:uuid
jcr:lastModified
jcr:lastModifiedBy
cq:lastModified
cq:lastModifiedBy
cq:lastReplicated
cq:lastReplicatedBy
cq:lastReplicationAction
jcr:versionHistory
jcr:predecessors
jcr:baseVersion
jcr:lastModifiedBy
cq:lastModified 
```
* *sortValuesFor* – multi value property names to ignore value ordering on.  Default property to ignore value ordering is _cq:tags_.

These default settings allows comparison of AEM author to publish instances or publish to other publish instances.

To use the tool, do the following:
1. Go to /bin/hashes.txt?path=/content/foo (where /content/foo should be replaced with the path you want to generate hashes for). Also, you can use curl to do this, for example:
curl -u admin:admin http://localhost:4502/bin/hashes.txt?path=/content/foo > hashes_pub1.txt
2. Save the output to a file and sort the results using the sort command on a linux machine then diff them. For example:
cat hashes_pub1.txt | sort > hashes_pub1_sorted.txt
cat hashes_pub2.txt | sort > hashes_pub2_sorted.txt
diff hashes_pub1_sorted.txt hashes_pub2_sorted.txt

Once you know which paths are different then you can go to each server you retrieved a diff against and use the /bin/jsondump.json servlet to get a dump of the content. For example here is how you would request this for a particular node:
curl -u admin:admin http://localhost:4502/bin/jsondump.json?path=/content/foo/en/jcr:content

Then you can use a json diff tool like this one or the one to manually see the differences:
http://tlrobinson.net/projects/javascript-fun/jsondiff


davidjgonzalez:
@andrewmkhoury looking forward to checking this out!!

davidjgonzalez:
@andrewmkhoury i grabbed this and started building a WebUI on top of it here..

https://github.com/davidjgonzalez/acs-aem-commons/tree/feature/jcr-checksum-combined

davidjgonzalez:
2015-09-13 at 5 00 pm

PR #528, opened 2015-09-10
Add a null check on the adapted ValueMap

dlh3:
If resource is a SyntheticResource (which happens when the invoking component is included via a <cq:include path="<a non-existent path>" />), an attempt to adaptTo(ValueMap.class) will result in a null value.  In this case, it makes more sense to fail gracefully, returning an empty results object instead of causing the NPE to bubble up as an ELException.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/widgets/MultiFieldPanelFunctions.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/widgets/MultiFieldPanelFunctions.java
@@ -57,7 +57,7 @@ public final class MultiFieldPanelFunctions {
     public static List<Map<String, String>> getMultiFieldPanelValues(Resource resource, String name) {
         ValueMap map = resource.adaptTo(ValueMap.class);
         List<Map<String, String>> results = new ArrayList<Map<String, String>>();
-        if (map.containsKey(name)) {
+        if (map != null && map.containsKey(name)) {
             String[] values = map.get(name, new String[0]);
             for (String value : values) {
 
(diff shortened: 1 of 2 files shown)


dlh3:
I currently work around this with a type check:
```jsp
<%-- 
if the included component has been generated (path does not yet exist), 
it is a SyntheticResource and will fail on widgets:getMultiFieldPanelValues() 
--%>
<c:if test="${resource.class.name ne 'org.apache.sling.api.resource.SyntheticResource'}">
    <c:set var="breakpoints" value="${widgets:getMultiFieldPanelValues(resource, 'breakpointOptions')}"/>
</c:if>
```

Of course, this would be a lot nicer if handled by a null check in the EL function (this PR).

davidjgonzalez:
@dlh3 thank you sir!

dlh3:
Thank you!

bundle/src/main/java/com/adobe/acs/commons/widgets/MultiFieldPanelFunctions.java
```diff
@@ -57,7 +57,7 @@ private MultiFieldPanelFunctions() {
     public static List<Map<String, String>> getMultiFieldPanelValues(Resource resource, String name) {
         ValueMap map = resource.adaptTo(ValueMap.class);
```
justinedelson: just for future reference (no reason to change this now), since we're now requiring AEM 6.0, a slightly better solution would be to replace this line with

    ValueMap map = resource.getValueMap();

That will guarantee that map is a non-null object.

dlh3: Great point, thanks for the input.

PR #536, opened 2015-09-14
Add alternative forms PRG strategy using "flash cache" instead of serialising form in URL

cpilsworth:
ENHANCEMENT IDEA:
In the forms module [documentation](http://adobe-consulting-services.github.io/acs-aem-commons/features/forms.html) there are a couple of downsides listed for the available  `Forward-as-GET`  & `POST-Redirect-GET` strategies.  One possible way of mitigating the form-size/messy-url issues of the PRG pattern is to use a [flash cache](http://guides.rubyonrails.org/action_controller_overview.html#the-flash) approach as popularised by RoR & later in the [Spring FlashMap](http://docs.spring.io/spring-framework/docs/current/spring-framework-reference/html/mvc.html#mvc-flash-attributes).

The main difference from the current PRG implementation would be rather than transferring the JSON encoded form contents on the URL between the POST and the GET, instead a token will be written to the URL.  When the GET request arrives, the form contents can be recovered using that token, the form would be removed once read.  The role of the flash cache is just to temporarily hold data during the course of the client redirect.

This approach would remove the constraints on the encoded form size, prettify the urls somewhat whilst continuing to stop prompts for users to repost forms when using the back button.

I am happy to create a PR for this, I just wanted to see if this is something that had been considered already and whether there were any thoughts on the approach.


cpilsworth:
The PR is a WIP using cookies.  Right now the cookies are not cleaned up after they are read.  

The form is JSON serialised and [escaped](https://sling.apache.org/apidocs/sling5/org/apache/sling/commons/json/http/Cookie.html#escape\(java.lang.String\)) and saved as a cookie with a name of the `getLookupKey`

cpilsworth:
It seems as though this feature might require an API change.  The right place to clear the cookies would be when the form is read from the request:
```
FormHelper formHelper = sling.getService(PostRedirectGetWithCookiesHelper.class);
Form form = formHelper.getForm("demo", slingRequest);
```
At this point there is no access to the response needed to clear the cookies.  Of course, we could pass the response in like so: 
```
Form form = formHelper.getForm("demo", slingRequest, slingResponse);
```
...but this changes the API for existing users.  Perhaps it's possible to provide some backwards compatibility using overloading, but that could get messy.

Thoughts?

davidjgonzalez:
@cpilsworth - I wouldn't mind another method; for the existing w could just have it call the 2 param.

One thing I've always wanted to change w Forms is creating a Form interface rather than a Form class w the impls. Not sure if you have thoughts on how that could be easily achieved (haven't looked in a while) 

Also wish the Form handled multi values

cpilsworth:
Added clear up of cookies, needs some tests to prove the various paths/cases are working though.

cpilsworth:
@davidjgonzalez With regards to the Form interface thing...  do you mean to extract the Form interface and push the logic from it, `AbstractFormHelperImpl.getPostForm` & `PostRedirectGetFormHelperImpl.getGetForm` down into private implementation types?  Or were you thinking of using the interface to define the form, something like [retrofit](http://square.github.io/retrofit/)?

I'll take a look at the multi-value field thing once I have some tests for this.

cpilsworth:
@davidjgonzalez @justinedelson Is there any reason for the junit:junit:4.8.1 dependency?  It's nearly 6 years old.  

Mocking out chains of dependencies (e.g. resourceResolverFactory > resourceResolver > resource) and injecting SCR @References (e.g. XSSAPI) is proving to be a pain.

It would make things easier if we could add the [sling-mock](https://sling.apache.org/documentation/development/sling-mock.html) and update the junit version to one that supports @Rule.

justinedelson:
> It would make things easier if we could add the sling-mock and update the junit version to one that supports @Rule.

Go for it! Not sure there's a lot of value in updating old unit tests just for the sake of it, but for new stuff, we can definitely start using sling-mocks.

cpilsworth:
Perfect, I just didn't want to wade in if it was kept for some compatibility reason.

justinedelson:
> Perfect, I just didn't want to wade in if it was kept for some compatibility reason.

Yeah, I'm assuming that updating the dependency doesn't break anything.

davidjgonzalez:
@cpilsworth I was thinking of making Form and interface and The guts of current Form.java into FormImpl.java ..

Iirc all forms are created by the Helpers anyhow and they shouldn't be created via new Form() in client code. My fear would be there is custome rcode that creates new Form() which would then break if it was changed to an interface

cpilsworth:
JUnit 4.12 might have made some internal changes that Powermock were relying on it seems... https://code.google.com/p/powermock/issues/detail?id=531.  Trying 4.11 that's been around a lot longer.

cpilsworth:
It looks like sling-mock has always [depended upon `sling.models.impl` = 1.1.0](https://github.com/apache/sling/commit/cd49ae54d4769cbf913e9832269e42e1e5035854#diff-6f35b0f03ef530abebe813c16ab1a6d0R71).  The `aem-api-6.0.0.1` has an earlier version (even though models wasn't a part of AEM6 if I remember correctly).  Upgrading to sling.models.impl 1.1.1 causes problems as the the code goes looking for `Model.defaultInjectionStrategy()` which isn't there in the `aem-api-6.0.0.1` version of the code.

justinedelson:
Might work if you moved the dependency order so that sling mocks is first. Actually, it might be best if the uberjar was the last dependency.

bundle/src/main/java/com/adobe/acs/commons/forms/helpers/impl/PostRedirectGetWithCookiesFormHelperImpl.java
```diff
@@ -0,0 +1,165 @@
...
+        return StringUtils.isBlank(encoded) ? "" : org.apache.sling.commons.json.http.Cookie.unescape(encoded);
+    }
+
+    /**
+     * Find the cookie on the request
+     * @param request
+     * @param name
+     * @return form cookie or null
+     */
+    protected Cookie findCookie(SlingHttpServletRequest request, String name) {
```
davidjgonzalez: I dont think its worth changing at this point; but we have a handy CookieUtil  that does a lot of Cookie reading/writing

cpilsworth:
@justinedelson It seems there was a sweet-spot for aem-api, above `org.apache.sling.commons.testing` and `org.apache.sling.testing.sling-mock`.  I've specified the `org.apache.sling.api` as `2.7.0`, the version used in `AEM-6.0`.

cpilsworth:
I've noticed there are a few project files with CRLF line endings:
```
find . -not -type d -exec file "{}" ";" | grep CRLF
./bundle/src/main/java/com/adobe/acs/commons/email/EmailService.java: ASCII c program text, with CRLF line terminators
./bundle/src/main/java/com/adobe/acs/commons/email/EmailServiceConstants.java: ASCII c program text, with CRLF line terminators
./bundle/src/main/java/com/adobe/acs/commons/email/impl/EmailServiceImpl.java: ASCII c program text, with CRLF line terminators
./bundle/src/main/java/com/adobe/acs/commons/email/process/impl/SendTemplatedEmailConstants.java: ASCII c program text, with CRLF line terminators
./bundle/src/main/java/com/adobe/acs/commons/forms/components/FormComponent.java: ASCII c program text, with CRLF, LF line terminators
./bundle/src/main/java/com/adobe/acs/commons/forms/components/MultiStepFormComponent.java: ASCII c program text, with CRLF, LF line terminators
./bundle/src/main/java/com/adobe/acs/commons/forms/helpers/ForwardAsGetFormHelper.java: ASCII c program text, with CRLF, LF line terminators
./bundle/src/main/java/com/adobe/acs/commons/forms/helpers/impl/synthetics/SyntheticSlingHttpServletGetRequest.java: ASCII c program text, with CRLF, LF line terminators
./bundle/src/main/java/com/adobe/acs/commons/forms/helpers/PostRedirectGetFormHelper.java: ASCII c program text, with CRLF, LF line terminators
./bundle/src/main/java/com/adobe/acs/commons/wcm/tags/wcmmode/AbstractMode.java: ASCII c program text, with CRLF line terminators
./bundle/src/test/java/com/adobe/acs/commons/email/EmailServiceImplTest.java: ASCII c program text, with CRLF line terminators
./bundle/src/test/resources/emailTemplate.txt: ASCII text, with CRLF line terminators
./bundle-twitter/target/classes/META-INF/MANIFEST.MF: ASCII text, with CRLF line terminators
```

cpilsworth:
Here is the example form using sightly:
```
<form method="post" action="${form.action}" data-sly-use.form="test.js">
	${form.formInputs @ context='unsafe'}
	<fieldset>
		<legend>Form Demo</legend>
		<div data-sly-test.myFieldError="${form.errors['myField']}">${myFieldError}</div>
		<label data-sly-attribute.class="${myFieldError ? 'error' : ''}">My Field:</label>
		<input type="text" name="myField" value="${form.data['myField']}"/>

		<input type="submit" value="Submit"/>
	</fieldset>
</form>
```
[Full code, including the post handler is in this gist](https://gist.github.com/cpilsworth/828dc1d12404467a3f8f).

cpilsworth:
@davidjgonzalez I've updated the code to use the CookieUtil class.  I've noticed that the `dropCookies` method doesn't clear the value.  Given that the serialised form could be quite large, it might be a good idea to reset the value [here](https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/util/CookieUtil.java#L141) to "" before sending back to the client on the GET response.

davidjgonzalez:
@cpilsworth works for me wrt to Cookie Drop.. I think you meant here though [1] ?  If you want to add to the PR works for me.

https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/util/CookieUtil.java#L237

cpilsworth:
@davidjgonzalez You were right, wrong line. Committed the change to remove cookie value.  Had to clone cookies rather than pass reference to request cookie + mutate. Clearing the value meant that it could no longer be read from the request.

Unless there are any further review comments, I think this task is complete.

Issue #519, opened 2015-08-28
AEM 6.0 SP2: ACS Common Multifield: the ‘loadcontent’ event does not fire for the widgets even when placed in a multifield when authored for second time


rgummalla:
On AEM 6.0 SP2, customer reported the issue as mentioned below

1.Create a component with dialog with multifield
2.Add ACS Commons multifield in the dialog
3.In the multifield, add listener to loadcontent event for the widget in multifield.
4.Author the component once. The listener fires as expected.
5.Edit the component again.
6.The listener for Loadcontent does not get fired as expected.

rgummalla:
This is a blocker in our current release 1.15. Wherec ustomer was not able to move few complex components for testing, with this issue. Also, They are in their last sprint and if they don’t close on this at the earliest, it would impact our timeline

justinedelson:
Can you attach a sample component which can be used to reproduce this issue?

rgummalla:
A JIRA https://jira.corp.adobe.com/browse/CQ-52842 opened contains the sample component

justinedelson:
Please post the package here.

rgummalla:
Well i tried uploading here. i am getting "Unfortunately, we don’t support that file type." its not working. please download from the link below

https://app.box.com/s/4iy3iog2k5nxcwyvt010renahithb50l

justinedelson:
Sorry, I meant please post a public link to the package here. Looks like the box link will work.

rgummalla:
Hi Justin,

Thanks for looking into this. Do you have any lead on the root cause for this issue. I am having a weekly synch up with customer. Appreciate if you can provide any update.

justinedelson:
If there is a time sensitivity, I'd suggest having the customizer try to create a patch. I don't forsee having a chance to dig into this in the next week or so.

rgummalla:
Issue 1:
Issue: When there is a dropdown inside a multifield, the second time we open the dialog, the dropdown field gets shrink, even if data is present.
Summary: This is related to a component with multifield in its dialog. When there is a dropdown inside a multifield, the second time we open the dialog, the dropdown field gets shrink, even if data is present.
How to reproduce: 
Step 1: Drag and drop a component with dropdown inside a multifield.
Step 2: Click and edit the component and click on OK.
Step 3: Click on edit again and try to modify the component.
Step 4: The dropdown field size is reduced. (Requirement field)
Occurrence: It occurs always for a multifield
[...]

PR #540, opened 2015-09-17
Feature/519 sreek

schoudry:
Bug Fixes for McAfee

1) Drag and Drop Issue: drag and drop to pathfield (in multifield) doesn't work when a component dialog is reopened without refreshing the page

2) Shrink Issue:  widgets added in multifield in a tab other than the visible one, shrink...

3) LoadContent Issue: Loadcontent event doesn;t fire on widgets added in multifield

4) Allow Blank Issue:  validation issues for widgets added in multifield

check the following two code statements; when a multifield is added, the component dialog isn't cached anymore, so every double click creates new dialog window...

            //dialog caching is a real pain when there are multifield items; remove from cache
            delete editable.dialogs[CQ.wcm.EditBase.EDIT];
            delete CQ.WCM.getDialogs()["editdialog-" + this.path];


justinedelson:
So this fixes all of #519?

schoudry:
yes

PR #541, opened 2015-09-20
Added "scale" image transformer

MikePfaff:
Can be used to auto-generate smaller variants of images (e.g. non-retina at 50%, etc.)


davidjgonzalez:
replaced by #553

PR #542, opened 2015-09-22
excluding tokens from authorizable packages. fixes #538

### LINKED ISSUE #538: Filter rules being created by the Authorizable Packager should not include .tokens subnode
Since the .tokens subnode is only a temporary folder which tracks the currently active user sessions of a specific user it should never be included in any authorizable package. Please always exclude the ".tokens" subfolder automatically from those packages.

justinedelson:


davidjgonzalez:
Marked as undocumented to update the "Updated" status in docs

PR #543, opened 2015-09-23
add a Page Info Provider which filters the available workflow models …

justinedelson:
…for a page

To use this:

* Create a node (if it doesn't already exists) called `cq:infoProviders` under your base page component.
* Under this node, create a new node with a property named `className` set to `com.adobe.acs.commons.wcm.impl.WorkflowModelFilterPageInfoProvider`. (there's no need to copy the default providers from `/libs/foundation/components/page/cq:infoProviders`)
* In your workflow models, on their jcr:content node, specify a property named `allowedPaths` containing a regex matching the paths for which you want this model to appear. If there is no `allowedPaths` property, the model will always appear.

More information on the underlying functionality can be found here: https://docs.adobe.com/docs/en/aem/6-1/develop/components/pageinfo.html

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/WorkflowModelFilterPageInfoProvider.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/WorkflowModelFilterPageInfoProvider.java
@@ -52,7 +52,7 @@ public class WorkflowModelFilterPageInfoProvider implements PageInfoProvider {
 
     private static final String KEY_WORKFLOWS = "workflows";
 
-    private static final String PN_APPLIES_TO = "appliesTo";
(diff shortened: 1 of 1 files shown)


justinedelson:
BTW, I'm not in love with the property name `appliesTo` and open to other suggestions.

justinedelson:
Per discussion with @davidjgonzalez, changed the property name to `appliesToPath` so that if we add other filter options (template) in the future, new properties can be added easily.

davidjgonzalez:
@justinedelson LGTM .. i don't mind `appliesToPath` .. but `allowedPaths` might be more apt (keeping same vocab as templates)

justinedelson:
good call on `allowedPaths`. Changed and merging.

PR #546, opened 2015-09-28
TouchUI RTE Plugin for Dialogs

schoudry:
@justinedelson @davidjgonzalez : A simple extension to show Touch UI Dialogs in RTE plugin (fullscreen mode). A common use usecase is providing fields in RTE for entering ticker content

Let me know if you have any suggestions in general & specifically around cutting down the configuration steps 

1) RTE Plugin feature - <b>insertDialogContent</b> in group <b>acs-commons</b>

image

2) Child node <b>insertDialogContent</b> contains the path of dialog - <b>dialogPath</b> to be shown when plugin icon is clicked, any listeners (currently supports <b>onsubmit</b>)

image

3) Create the dialog to be shown eg. /apps/test-dialogs/cq:dialog

image

4) Dialog shown when the plugin icon is clicked

image

5) Configured <b>onsubmit</b> listener function should return passed in dialog data as html eg. &lt;h1&gt; ACS AEM Commons &lt;/h1&gt;

image

6) RTE html source

image

### DIFF
--- a/content/src/main/content/jcr_root/apps/acs-commons/rte-plugins/source/dialog-plugin.js
+++ b/content/src/main/content/jcr_root/apps/acs-commons/rte-plugins/source/dialog-plugin.js
@@ -78,7 +78,7 @@
         }
     });
 
-    //popover dialog thats hosts iframe
+    //popover dialog hosting iframe
     INSERT_DIALOG_CONTENT_PLUGIN_DIALOG = new Class({
         extend: CUI.rte.ui.cui.AbstractBaseDialog,
 
@@ -110,7 +110,7 @@
         }
     });
 
-    //extend the toolkit implementation for returning custom toolbar builder and dialog manager
+    //extend the toolkit implementation for custom toolbar builder and dialog manager
     ACS_TOOLKIT_IMPL = new Class({
         toString: "ACSToolkitImpl",
 
@@ -268,7 +268,13 @@
         },
 
         execute: function (execDef) {
-            execDef.editContext.doc.execCommand("insertHTML", false, execDef.value);
+            var acsPlugins = execDef.component.registeredPlugins[GROUP],
+                popoverConfig = acsPlugins.config[INSERT_DIALOG_CONTENT_FEATURE],
+                /*jshint -W061 */
+                onSubmitFn = eval("(" + popoverConfig.onsubmit + ")"),
+                html = onSubmitFn(execDef.value);
+
+            execDef.editContext.doc.execCommand("insertHTML", false, html);
         }
     });
 
--- a/content/src/main/content/jcr_root/apps/acs-commons/rte-plugins/source/dialog-plugin.js
+++ b/content/src/main/content/jcr_root/apps/acs-commons/rte-plugins/source/dialog-plugin.js
@@ -1,0 +1,361 @@
+/*
+ * #%L
+ * ACS AEM Commons Package
+ * %%
+ * Copyright (C) 2013 Adobe
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


justinedelson:
So, to perhaps oversimplify this, it is a generic extension point for adding Dialog-based RTE plugins?

schoudry:
yes... to show touch dialogs in RTE

content/src/main/content/jcr_root/apps/acs-commons/rte-plugins/source/dialog-plugin.js
```diff
@@ -0,0 +1,367 @@
...
+    "use strict";
+
+    var _ = window._,
+        Class = window.Class,
+        CUI = window.CUI,
+        REQUESTER = "requester",
+        GROUP = "acs-commons",
+        INSERT_DIALOG_CONTENT_FEATURE = "insertDialogContent",
+        INSERT_DIALOG_CONTENT_DIALOG = "insertDialogContentDialog",
+        ACS_CUI_TOOLBAR_BUILDER,
```
justinedelson: perhaps more a style issue than anything critical, but I'd rather we didn't use a constant styling for non-constant values. i.e. this just should be AcsCuiToolbarBuilder

content/src/main/content/jcr_root/apps/acs-commons/rte-plugins/source/dialog-plugin.js
```diff
@@ -0,0 +1,367 @@
...
+            var uiSettings = this.superClass._getUISettings(options),
+                toolbar = uiSettings.fullscreen.toolbar,
+                feature = getUISetting();
+
+            if (toolbar.indexOf(feature) === -1) {
+                toolbar.splice(3, 0, feature);
+            }
+
+            if (!this._getClassesForCommand(feature)) {
+                this.registerAdditionalClasses(feature, "coral-Icon coral-Icon--tableEdit");
```
justinedelson: Is this a default class which can be overridden? Or will the tableEdit class always be used?

justinedelson:
This looks very cool to me. There's no obvious way to me to make it simpler to configure; those seem like the natural steps. The only thing I might change is to flip the order. I would do step #3 before step #1. But that's more of a documentation issue than anything else.

The only functional concerns I have is the ability to set the className (maybe you can do this, it wasn't clear to me) and the rollover text.

I think if you can change the variable names which aren't constants to not look like constants, we're good to merge this.

schoudry:
yeah ill make the changes

1) remove the constant styling for non-constant values
2) config param for "open dialog icon" ( is coral-Icon--tableEdit good for default, any other icon suggestion?)
3) config param for rollover text / tooltip

schoudry:
@justinedelson i have to explore providing a config for

2) config param for "open dialog icon" ( is coral-Icon--tableEdit good for default, any other icon suggestion?)

added other changes...

justinedelson:
FWIW, I think coral-Icon--tableEdit is a fine default :)

davidjgonzalez:
@schoudry @justinedelson this good to merge?

schoudry:
@davidjgonzalez i think so - couldn't find an easy solution for "config param for - open dialog icon", but i believe we can fix it as a bug, later? @justinedelson

davidjgonzalez:
@schoudry - if there is no easy way, I'd log as an enhancement for later (IMO this isn't even a bug)

singhd12:
@justinedelson I am trying to test this plugin with AEM 6.3 but seems like its not working for me. I am not able to see plugin for both inline and fullscreen. 

Steps i am following
1) Install ACS 3.9.0  - https://github.com/Adobe-Consulting-Services/acs-aem-commons/releases/tag/acs-aem-commons-3.9.0
2) Overlay One of RTE components - /apps/mcm/campaign/components/personalization
3) Follow Instruction provided under step section - 

https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/content/src/main/content/jcr_root/apps/acs-commons/rte-plugins/source/dialog-plugin.js

Please let if this is compatible with AE 6.3

Devesh Singh

justinedelson:
@singhd12 please create a separate issue and provide a content package to reproduce the issue.

PR #548, opened 2015-09-30
Feature/jcr checksum combined

davidjgonzalez:
I've been working with @andrewmkhoury to build out JCR Checksum Compare (lionshare of the credit here goes to andrew).

We've made a Web UI that builds on top of his JCR checksum comparison APIs that allows JCR trees on multiple systems to be easily compared.

## Configure the JCR Check Comparison Parameters

2015-09-29 at 9 28 pm

2015-09-29 at 9 29 pm

2015-09-29 at 9 30 pm

## Get the Checksum Comparison

2015-09-29 at 9 31 pm

## Click on an entry to get the JSON diff

2015-09-29 at 9 32 pm

## Download the full JSON dump based on the Configuration tab settings

2015-09-29 at 9 32 pm

## JSON Dump file
2015-09-29 at 9 44 pm


bstopp:
This may seem like a dumb Q: but is this not more appropriate for ACS Tools instead of Commons?

davidjgonzalez:
@bstopp that was my initial thought, but the ~99% use case for this is troubleshooting Production and Tools is "intended for dev and qa" and generally cautioned against being installed on Production.

If Tools ever lifts that restraint, then i agree this would be better suited in Tools.

I actually made an issue in the Tools repo wrt to this feature and had a similar convo there.

bundle/src/main/java/com/adobe/acs/commons/analysis/jcrchecksum/impl/ServletConstants.java
```diff
@@ -0,0 +1,40 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+
+package com.adobe.acs.commons.analysis.jcrchecksum.impl;
+
+public final class ServletConstants {
```
bstopp: This is a personal opinion: but I usually use interfaces instead of final classes when creating a constants holder. Default state is no constructor, and all attributes are public static final.

justinedelson: You should read Effective Java :) Bloch makes some good arguments why constant interfaces are a bad idea.

bstopp:
Going through this: have you run a quick code analysis on this? without looking at all of the classes side-by-side, there appears to be duplication, but maybe i'm just seeing things.

justinedelson:
@davidjgonzalez can you take a look at why the build is failing for this PR? looks like a failing test.

davidjgonzalez:
@justinedelson yup - havent had the time, maybe this weekend

bundle/src/main/java/com/adobe/acs/commons/analysis/jcrchecksum/impl/ServletConstants.java
bstopp: I think that's actually on my reading list. Maybe i'll move it up in the queue, heh. Thanks!

venturewise: https://cebulko.com/Programming-Resources/Effective%20Java%20-%202nd%20Edition.pdf
:-)

bstopp: I get it, i get it!! ^_^ (i borrowed it off my boss' desk already)

PR #553, opened 2015-10-12
Feature/scale image transformer

davidjgonzalez:
Replaces PR #541 

Transformer Name: scale

Params
* scale = double
* round = up || down || round
 * if blank, uses round

----

Per @bstopp removed support for <String>% to align property API with other transforms that take % (as decimal numbers)


bundle/src/test/java/com/adobe/acs/commons/images/transformers/impl/ScaleImageTransformerImplTest.java
```diff
@@ -0,0 +1,147 @@
...
+    @Test
+    public void testTransform_emptyParams() throws Exception {
+        transformer.transform(layer, properties);
+
+        verifyZeroInteractions(layer);
+    }
+
+    @Test
+    public void testTransform_nullParams() throws Exception {
+
```
bstopp: you forgot to call the transformer here...

bundle/src/main/java/com/adobe/acs/commons/images/transformers/impl/ResizeImageTransformerImpl.java
```diff
@@ -53,14 +53,14 @@
...
+    static final String KEY_WIDTH_ALIAS = "w";
 
-    private static final String KEY_HEIGHT = "height";
-    private static final String KEY_HEIGHT_ALIAS = "h";
+    static final String KEY_HEIGHT = "height";
+    static final String KEY_HEIGHT_ALIAS = "h";
 
 
-    private static final int DEFAULT_MAX_DIMENSION = 50000;
+    public static final int DEFAULT_MAX_DIMENSION = 50000;
```
bstopp: Made this public but the others package default? Why the difference?

bundle/src/test/java/com/adobe/acs/commons/images/transformers/impl/ScaleImageTransformerImplTest.java
davidjgonzalez: no wonder that was passing w flying colors!!

bundle/src/main/java/com/adobe/acs/commons/images/transformers/impl/ResizeImageTransformerImpl.java
davidjgonzalez: no reason, moving back to private; others are package level since theyre used in ScaledImageTransformerImpl to populate the params map

bstopp:
Looks good to me.

Issue #499, opened 2015-08-13
maxItems not working for multifield


saulovenancio:
Hi Guys, 
I saw that there is a bug in this line of add_multifield_limit.js

            if (this.maxItems && (this.maxItems === this.getActualItemCount())) {

The comparison runs two different types so this never gets true.
A simple fix is to do 

if (this.maxItems && (this.maxItems == this.getActualItemCount())) {

that wont check the type.
The way it is implemented today is not working. 

Should i open a pull request and implement it?
Thanks.

justinedelson:
Aren't maxItems and getActualItemCount() both numbers?

justinedelson:
According to @amclin, #555 fixed this.

amclin:
To be clear, #555 would prevent the resulting form from being submitted, but I am unsure as to whether the check here that blocks an author from adding to many multifield entries would actually work or not.

amclin:
@saulovenancio is correct typeof(this.maxItems) is string whereas typeof(this.getActualItemCount()) is number. So the types never match the test never works.

PR #555, opened 2015-10-15
Ensure max limit is checked on form validation

amclin:
Currently if an author surpasses the max limit on a multipanel (eg. the limit was added after the content was first created) then the max limit is not respected and the author is able to submit a an invalid form. This change runs the max limit check during validation, just like the min limit already operates.


amclin:
This also would address #499

justinedelson:
Thanks!

PR #561, opened 2015-10-21
Allow a custom icon font prefix to be specified

justinedelson:
For cases where the default prefix of 'fa ' is creating conflicts, the property icon-font-prefix (for Granite UI) and iconFontPrefix (for ClassicUI) can be specified. The special value '-' can be used to suppress any prefix. The property value will be automatically appended with a space if necessary.


davidjgonzalez:
@justinedelson this is the solution for this right? I know there was some chatter on gitter - Looks fine to me - just want to make sure we merge the right one.

justinedelson:
Based on feedback, decided to go a different way in #571

PR #563, opened 2015-10-23
Letter/Pillar box support

bstopp:
This is a replacement for #364. Supports l/p-boxing when resize aspect ratio isn't right. Also allows configuration to specify background color and opacity if desired (default transparent).


bstopp:
FYI, verification may fail, there's some PMD issues on workflow code; I couldn't get past analysis step.

davidjgonzalez:
@bstopp its passing our PMD rules for me

PR #564, opened 2015-10-24
Feature/ResourceResolver.map(..) Transformer

davidjgonzalez:
Transformer that called `slingRequest.getResourceResolver().map(slingRequest, <path>)` on a configurable set of `<element>:<attributes>`

This is to allow mapping of attributes that may not end in .html or .htm and/or not "links" (OOTB Linkchecker has some limitations with its mapping)

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/ResourceResolverMapTransformerFactory.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/ResourceResolverMapTransformerFactory.java
@@ -1,0 +1,122 @@
+/*
+ * #%L
+ * ACS AEM Commons Bundle
+ * %%
+ * Copyright (C) 2015 Adobe
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


bstopp:
Add nameHints to this, so i don't have to update my other PR.

davidjgonzalez:
@bstopp done

bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/ResourceResolverMapTransformerFactory.java
```diff
@@ -0,0 +1,127 @@
...
+            value = "Pipeline Type: {pipeline.type}, for element:attributes [{attributes}]")
+})
+@Service
+public final class ResourceResolverMapTransformerFactory implements TransformerFactory {
+
+    private static final String[] DEFAULT_ATTRIBUTES = new String[]{"img:src"};
+    private Map<String, String[]> attributes;
+    @Property(label = "Rewrite Attributes",
+            description = "List of element/attribute pairs to rewrite",
+            value = {"img:src"})
```
bstopp: Can you not reference the constant here instead?

davidjgonzalez: It's just final not constant :T

bstopp: I find that strange; but fair enough.

bstopp:
Can you add a negative test case? verify no interactions when element doesn't have any matching transformation attributes?

PR #568, opened 2015-10-26
Sling/WCM Mode Utility

### LINKED ISSUE #54: Write an isAuthor method?
Does it help to write a utils method to get check the sling runmode and return an isAuthor boolean? I can do it, but, need to know from you guys if I should create a new SlingUtils class or add this to an existing class. 

Let me know what you thinks. 

Thanjs

bstopp:
Intended to close #54. 

Also @davidjgonzalez though this would be a better place for the mode logic from the component Helper.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/util/ModeUtil.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/util/ModeUtil.java
@@ -36,14 +36,8 @@ import com.day.cq.wcm.api.WCMMode;
 
 import aQute.bnd.annotation.ProviderType;
 
-//@formatter:off
 @ProviderType
-@Component(
-        immediate = true, 
-        enabled = true, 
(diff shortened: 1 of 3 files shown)


bundle/src/main/java/com/adobe/acs/commons/util/ModeUtil.java
```diff
@@ -0,0 +1,187 @@
...
+import com.day.cq.wcm.api.WCMMode;
+
+import aQute.bnd.annotation.ProviderType;
+
+//@formatter:off
+@ProviderType
+@Component(
+        immediate = true, 
+        enabled = true, 
+        label = "ACS AEM Commons - Mode Utilities",
```
davidjgonzalez: Oh yeh - the new convention is to only have labels/descriptions when metatype=true - also enabled=true is default, so dont need to explicitly set

bstopp:
Fixed.

PR #570, opened 2015-10-28
Find bugs fixes.

bstopp:
This fixes all items found in the find bugs with analysis profile. Please review carefully, I made some assumptions with a few classes with which i was unfamiliar.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/analysis/jcrchecksum/impl/options/RequestChecksumGeneratorOptions.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/analysis/jcrchecksum/impl/options/RequestChecksumGeneratorOptions.java
@@ -33,6 +33,7 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.nio.charset.Charset;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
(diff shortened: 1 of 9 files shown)


bundle/src/main/java/com/adobe/acs/commons/analysis/jcrchecksum/impl/options/RequestChecksumGeneratorOptions.java
```diff
@@ -93,13 +94,14 @@ public RequestChecksumGeneratorOptions(SlingHttpServletRequest request) throws I
...
 
-    private static Set<String> getPathsFromInputstream(InputStream is) throws IOException {
+    private static Set<String> getPathsFromInputstream(InputStream is, String encoding) throws IOException {
         if (is == null) {
             return Collections.EMPTY_SET;
         }
 
         Set<String> paths = new HashSet<String>();
-        BufferedReader br = new BufferedReader(new InputStreamReader(is));
+        encoding = (encoding != null) ?  encoding : Charset.defaultCharset().name();
```
jstomar: I would prefer to do this check before passing "ëncoding" to this method. Caller should pass either default or other value.

bundle/src/main/java/com/adobe/acs/commons/errorpagehandler/impl/ErrorPageHandlerImpl.java
```diff
@@ -514,18 +514,16 @@ private Resource findFirstRealParentOrSelf(SlingHttpServletRequest request, Reso
...
-        if (resource != null) {
-            final Resource parent = resource.getParent();
-            if (parent != null) {
-                log.debug("Found real aggregate resource via getParent() at [ {} ]", parent.getPath());
-                return parent;
-            }
+        final Resource parent = resource.getParent();
+        if (parent != null) {
+            log.debug("Found real aggregate resource via getParent() at [ {} ]", parent.getPath());
+            return parent;
```
jstomar: If Resource is not null, don't think parent of that resource could be null. isn't?

davidjgonzalez: Yes - it could be a non existing resource which is not bull but whose parent is null. This is a common occurrence in this use case (handling not found requests)

PR #571, opened 2015-10-28
removing prefix support in font icon picker and moving prefix for Fon…

justinedelson:
…t Awesome to list

As discussed on Gitter, while this prefix support worked well for Font Awesome, it makes using other icon fonts hard.


davidjgonzalez:
Added an alert at top of feature doc explaining any custom FA lists made under 2.0.0 must be updated.

PR #573, opened 2015-11-01
Created Email object PER recipient to resolve #462

### LINKED ISSUE #462: Issue with email service
1. If we have more than one recipients. The second recipient is getting one attachment which is the mail template. Third recipient gets two attachments etc. This is the behaviour in outlook.

2. When we are sending mail to gmail or other. Second recipient getting two mails. Third receipt getting three mails etc.

davidjgonzalez:
Resolves #462 

@maximilianvoss can you please review? Thanks

----
If we have more than one recipients. The second recipient is getting one attachment which is the mail template. Third recipient gets two attachments etc. This is the behaviour in outlook.

When we are sending mail to gmail or other. Second recipient getting two mails. Third receipt getting three mails etc.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/email/impl/EmailServiceImpl.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/email/impl/EmailServiceImpl.java
@@ -19,33 +19,34 @@
 */
 package com.adobe.acs.commons.email.impl;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-
-import javax.jcr.Session;
-import javax.mail.internet.AddressException;
-import javax.mail.internet.InternetAddress;
-
-import org.apache.commons.lang.StringUtils;
-import org.apache.commons.lang.text.StrLookup;
-import org.apache.commons.mail.Email;
-import org.apache.commons.mail.HtmlEmail;
-import org.apache.commons.mail.SimpleEmail;
-import org.apache.felix.scr.annotations.Component;
-import org.apache.felix.scr.annotations.Reference;
-import org.apache.felix.scr.annotations.Service;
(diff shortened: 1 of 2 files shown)


davidjgonzalez:
@dchang this may be of interest to you as well on current proj

dchang:
Thanks for the heads up @davidjgonzalez ! Looks like we used this email service in several places, will keep this fix in mind

maximilianvoss:
Actually I would have patched just following:

````
        Email email = getEmail(templatePath, emailParams);

        if (email == null) {
            throw new IllegalArgumentException("Error while creating template");
        }

        MessageGateway<Email> messageGateway = messageGatewayService.getGateway(email.getClass());

        for (InternetAddress address : recipients) {
            try {
                email.setTo(Collections.singleton(address));
                messageGateway.send(email);
            } catch (Exception e) {
                failureList.add(address);
                log.error("Exception sending email to " + address, e);
            }
         }

        return failureList;
````
to
````


        MessageGateway<Email> messageGateway = messageGatewayService.getGateway(email.getClass());

        for (InternetAddress address : recipients) {
            try {
               Email email = getEmail(templatePath, emailParams);
               if (email == null) {
                     throw new IllegalArgumentException("Error while creating template");
                }
                email.setTo(Collections.singleton(address));
                messageGateway.send(email);
            } catch (Exception e) {
                failureList.add(address);
                log.error("Exception sending email to " + address, e);
            }
         }

        return failureList;
````

It's the easier fix and not too much has to be touched. 
__WDYT?__

davidjgonzalez:
@maximilianvoss I made the more significant changes to minimize interactions w the data store. caches exist, but this guarentees 1 IO read for an unbounded recipient size.

maximilianvoss:
@davidjgonzalez indeed, your solution makes a lot more sense then :)

Issue #575, opened 2015-11-05
Fixed LAST_SUFFIX_PATTERN in NamedTransformImageServlet.java


m-kay:
Is there a reason for the fixed suffix pattern in the [NamedTransformImageServlet.java:125](https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/images/impl/NamedTransformImageServlet.java#L125)? For SEO reasons the name of the image should be the real file name and not the same name for every image.

My suggestion is to make this configurable via OSGi configuration so one could still use the current pattern but it would be possible to allow the real filename.

PR #576, opened 2015-11-05
Make last suffix (filename) pattern configurable

m-kay:
Fix: #575

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/images/impl/NamedTransformImageServlet.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/images/impl/NamedTransformImageServlet.java
@@ -87,6 +87,13 @@ import java.util.regex.Pattern;
                 propertyPrivate = false
         ),
         @Property(
+		label = "File Name Pattern",
+		description = "RegEx Pattern to filter allowed filenames",
+		name = NamedTransformImageServlet.NAMED_IMAGE_FILENAME_PATTERN,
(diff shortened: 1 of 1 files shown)


bstopp:
Maybe i'm not seeing the whole situation here. Wouldn't this require a significant number of configurations to support, since every image has a unique file name? 

That being true, then regex for the filename matcher should be a list of strings, instead of a single regex, since this class isn't a factory. But event then, now every time a new asset is uploaded you'd need to update the rules for this configuration to support that pattern.

Finally, the URL already has the file name in it, it's the resource name on which the *transform* extension is applied, does not duplication of tokens degrade the SEO performance? 

Just my thoughts.

m-kay:
To support all filenames I would use a regex like .*?\.(jpg|jpeg|png|gif) or even .*
I don't understand this suffix check at all and I think in my case I don't need it, but to still support it I recommended to make it configurable instead of remove it completely.

The search engines are not aware of this Sling suffix so they suppose this suffix as the real file name and not the one included in the path. Furthermore if someone saves one of these images the filename is always image.jpg instead of the real filename because the browser isn't aware of this suffix either.

davidjgonzalez:
Seems reasonable to me! Thanks!

m-kay:
Thank you for merging. 
When will you release the next version? We would like to include this fix as soon as possible.

davidjgonzalez:
@m-kay this is in 2.1.2; should be hitting adobe repo tomorrow (11/16)

PR #578, opened 2015-11-10
Allow callers to specify node types for WfPkgMgr.getPaths()

shsteimer:
I had a need to create a WF package with oak:Unstructured nodes.  From my testing, the creation part worked, but when using the getPaths API, it failed to return those nodes back to me.  While modifying the config could work, I probably don't want the oak:Unstructured returned in the general case, just this specific case, so modifying the API to allow this scenario seemed like the best option, especially since it's already supported by the underlying ResourceCollection API.

A note on testing: I looked at creating a test for this, but since the ResourceCollection is being mocked in the current tests, it didn't seem worth while.  Probably possible by mocking node type, extending the existing mock, etc.  If you want me to do that I will, just figured I'd start with this PR, and make those changes based on feedback.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/workflow/WorkflowPackageManager.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/workflow/WorkflowPackageManager.java
@@ -82,6 +82,23 @@ public interface WorkflowPackageManager {
     List<String> getPaths(ResourceResolver resourceResolver, String workflowPackagePath) throws RepositoryException;
 
     /**
+     * Gets the payload paths in the Workflow Package.
+     *
+     * This method will always return a List.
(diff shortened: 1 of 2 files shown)


davidjgonzalez:
@shsteimer Have you tried this code out? I swear i tried added cq:Tags to the default white list at first and bad things happened. It may have just been the Workflow Packager Web UI though too... 

Otherwise the PR looks good though!

shsteimer:
I did test it with oak:Unstructured and it seemed to work fine for my purposes.  I can run some tests with other node types and see what happens.

davidjgonzalez:
@shsteimer should be fine; really this functionality already "exists"; its just that its defined in the OSGi config, and not method params. Nothing would stop someone from putting unsupported types int he OSGi config today.

PR #580, opened 2015-11-11
Resolves 579; Only inject JS on author and when there is at least 1 a…

davidjgonzalez:
…ctive notification. Also aligns default style value to green

@andre-inside can you spot check this?


davidjgonzalez:
@justinedelson wdyt; should this require a config policy to start?

Issue #477, opened 2015-06-26
touchui-composite-multifield doesn't load values for several field types e.g. checkbox, dropdown selection


keseruk:
When using checkbox on a touch ui dialog built using touchui-composite-multifield, then values from those fields added to the composite multifield form will be saved to JCR properly, but several fields won't load/display the value when I open the dialog again. We observer this issue with checkbox and dropdown selection fields so far. To reproduce the issue, just add a checkbox to the dialog example in touchui-composite-multifield.js.

justinedelson:
Can you add a content package containing a test component?

keseruk:
Added simple component for demonstrating the issue. Rename to .zip. Thanks!
![checkbox](https://cloud.githubusercontent.com/assets/6104409/8377550/7f072a76-1c12-11e5-8a4e-56b3cc0f4465.png)

keseruk:
I found the same issue in the widget published on http://experience-aem.blogspot.hu/2015/06/aem-61-touch-ui-composite-multifield-store-values-as-child-nodes.html and the fix that Sreekanth provided worked. Probably works here, too?

justinedelson:
@kalyanar - can you help identify what needs to be done here?

kalyanar:
Sure.
I shall look at it over the weekend.

akuntsch:
I have the same problem with checkboxes.

Is there any update on the issue?

NielsInc:
Dropdown selection fields are still not being populated correctly in the latest release (2.1.2).
For me the fix that Sreekanth provides is an alternative, but has a different approach since it stores the values of the children in a child node instead of in JSON format - which means adapting code for an unofficial fix that might be implemented differently in the acs-aem-commons - so not very desirable...

schoudry:
I'll work on a fix

akuntsch:
Thanks @schoudry. Your fix works for checkboxes.

mt33:
Thanks for this PR, exactly what I was looking for!

NielsInc:
@schoudry Is there an event that is triggered when the data is loaded? 
The values for the multifield are set after the `foundation-contentloaded` event is triggered, so the content is not yet available then

schoudry:
@NielsInc with current code, no custom events are triggered after data load or data update

mt33:
I've run into the same use @NielsInc - submitted #591

PR #584, opened 2015-11-20
477 

schoudry:
touchui-composite-multifield doesn't load values for several field types e.g. checkbox, dropdown selection #477

@justinedelson @davidjgonzalez


davidjgonzalez:
@schoudry thanks!

NielsInc:
@schoudry Is there an event that is triggered when the data is loaded? 
The values for the multifield are set after the `foundation-contentloaded` event is triggered, so the content is not yet available then

sudipdgp:
Selection box is working but still checkbox values are not persisting in the dialog when used under multifield

PR #588, opened 2015-11-25
Feature/http cache

davidjgonzalez:
Minor adjustments to get it working on my local. It's *awesome* :)

One change I made was removing the `customAttributes` OSGi Property from HttpCacheConfigImpl. The reasoning is this list of strings may not be sufficient for all customizers. Instead I collapsed the GroupCacheKeyFactory into GroupCacheConfigExtension, and created a impl specific userGroups OSGi Property that both the Extension and CCFactory can leverage.

Technically these 2 classes could be decoupled since they are registered to their respective Service interfaces, and HttpCacheConfig's configs still let you specific the service filter for each individually, but i expect these 2 things (Extension and CCFactory) will be very tightly coupled in most impls.

Other minor changes included handling CacheKey exceptions on key-based invalidation (i have it invalidating the entire cache if that happens)

If you can take a look at the the MemPersistenceObject and make sure the Header logic looks right, would be handy. I was getting an exception when it tried to unwrap the Header values from a wrappedSet to a List.

Made some minor tweaks to the MBean to get it to come up; but think there may be a few issues w it still.

```
25.11.2015 00:07:40.629 *TRACE* [0:0:0:0:0:0:0:1 [1448428060356] GET /content/geometrixx/en.html HTTP/1.1] com.adobe.acs.commons.httpcache.filter.impl.HttpCacheFilter Delivered un-cached request [ /content/geometrixx/en.html ] in 265 ms
25.11.2015 00:07:54.323 *DEBUG* [0:0:0:0:0:0:0:1 [1448428074309] GET /content/geometrixx/en.html HTTP/1.1] com.adobe.acs.commons.httpcache.filter.impl.HttpCacheFilter Delivered cached request [ /content/geometrixx/en.html ] in 13 ms

25.11.2015 00:25:02.799 *TRACE* [0:0:0:0:0:0:0:1 [1448429099625] GET /content/geometrixx/en/events/techsummit.html HTTP/1.1] com.adobe.acs.commons.httpcache.filter.impl.HttpCacheFilter Delivered un-cached request [ /content/geometrixx/en/events/techsummit.html ] in 3171 ms
25.11.2015 00:27:24.362 *DEBUG* [0:0:0:0:0:0:0:1 [1448429244347] GET /content/geometrixx/en/events/techsummit.html HTTP/1.1] com.adobe.acs.commons.httpcache.filter.impl.HttpCacheFilter Delivered cached request [ /content/geometrixx/en/events/techsummit.html ] in 15 ms
```


justinedelson:
How is this different than the in memory HTTP response cache provided since AEM 6.0?

davidjgonzalez:
__Far___ more flexible. That only works w anon requests and has a 60 ls ttl

davidjgonzalez:
I accidentally PR'd here, and shouldve done got @Sivaramvt's - thus the closing.

@justinedelson btw - there is more/better info on this feature here: https://github.com/Sivaramvt/acs-aem-commons/pull/1

OOTB Cache
* Listens on a series of path prefixes
* Cache Invalidation is TTL-based on 60s timer
* Cache is only for anonymous requests

Our Cache
* Defines cache configs for each cache scheme, which includes
  * RequestURI patterns whitelist
  * RequestURI patterns blacklist
  * Authentication support (anon-only, auth-only, both)
  * Pluggable accept rules (roll your own)
  * Pluggable cache key creation (want to cache results by group's X and Y + profile attribute x? and do NOT cache requests by Group z ever?  no problem, just write it.)
  * Cache invalidation rules (TTL or CUD event on resource by path pattern)
* Pluggable persistence (mem right now; disk and jcr possible)
* MBean to inspect cache
* Other stuff im forgetting

PR #591, opened 2015-12-03
Trigger event after Touch UI composite multifields are processed

mt33:
This allows me to do extra things with fields within the multifield. Specifically, I wanted to disable a textfield if it has a certain value, but unable to do so without this hook.

Not sure if there's a convention for custom JS events – happy to revise.

### DIFF
--- a/content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/source/touchui-composite-multifield-nodestore.js
+++ b/content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/source/touchui-composite-multifield-nodestore.js
@@ -176,7 +176,7 @@
                     cmf.buildMultiField(data[mName], $multifield, mName);
                 });
 
-                $document.trigger('touchui-composite-multifield-nodestore-ready');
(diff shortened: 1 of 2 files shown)


content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/source/touchui-composite-multifield-nodestore.js
```diff
@@ -175,6 +175,8 @@
                 _.each(mNames, function($multifield, mName){
                     cmf.buildMultiField(data[mName], $multifield, mName);
                 });
+
+                $document.trigger('touchui-composite-multifield-nodestore-ready');
```
schoudry: passing which multifield's data was loaded might help with listening, something like 

```.trigger( "touchui-composite-multifield-nodestore-ready",  $multifield);```

mt33: Good idea

PR #595, opened 2015-12-09
Bug fix in touchui-composite-multifield

mt33:
Oops - $multified is out of scope, passing mNames instead.


mt33:
@schoudry what do you think of passing `mNames`?

Issue #596, opened 2015-12-10
Deprecate EnsurePropertyIndex


joerghoh:
With the EnsureIndex functionality in place I don't see a need for EnsurePropertyIndex any more. We should deprecate it asap (including a warn message when it's used) and then remove it from the codebase.

davidjgonzalez:
agree

justinedelson:
+1

davidjgonzalez:
Closed via #598

PR #598, opened 2015-12-11
Enhancement/ensure oak property scheduler

davidjgonzalez:
Deprecates EnsurePropertyIndex #596 

Formatting and minor cleanup of #593 - most notable change is relying on `resourceResolver.hasChanges()` over tracking `doSave` state for commit trigger.


davidjgonzalez:
@joerghoh this is a bit of a pain to see the diff since the formatter hit the indentation pretty hard. The only substantial change is the rr.hasChanges() check over the doSave check.

joerghoh:
ok :-)

PR #599, opened 2015-12-11
adding string substitution of system properties into html in AEM Envi…

### LINKED ISSUE #597: AEM environment indicator should be able to embedd system properties into HTML
I have a case where I have a large number of environments which makes configuration via OSGI config quite inconvenient. Thus I would like to use variables which are replaced by either system properties or environment variables.
This makes it very convenient to embedd for example the hostname into the indicator.

justinedelson:
…ronment Indicator. Fixes #597

@davidjgonzalez @joerghoh something like this would work (and is very simple)


davidjgonzalez:
ah - simple. i didnt take @joerghoh's request of system properties literally -- perfect!

justinedelson:
I always take @joerghoh literally :)

davidjgonzalez:
@joerghoh look good? I'd like to merge and cut a release shortly.

joerghoh:
looks good

PR #601, opened 2015-12-13
Dispatcher TTL - minor alignments

davidjgonzalez:
Moved Dispatcher TTL to dispatcher.ttl package, removed author attribution, make felix nameHints private

@bstopp WDYT?


davidjgonzalez:
Declining for http.headers packaging

PR #616, opened 2016-01-12
Children As Property Resource Wrapper

davidjgonzalez:
Serializes super-flat children into a single property as a JSON object. This tends to be much more performant.

```
Add Operation of 10,000
nt:unstructured: 8773ms ~> 8s
oak:Unstructured: 6374ms ~> 6s
unsorted property based: 632ms ~> .6s
sorted property based: 1,167ms ~> 1s

Add Operation of 100,000
nt:unstructured: 160,227ms ~> 160s
oak:Unstructured: 87,324ms ~> 87s  - Package built in 101377ms.
property based: 5,240ms ~> 5s - Package built in 4528ms.
sorted property based: 6,646ms ~> 7s
```

Does not support multi-value properties

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/synth/children/ChildrenAsPropertyResourceWrapper.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/synth/children/ChildrenAsPropertyResourceWrapper.java
@@ -1,6 +1,5 @@
 package com.adobe.acs.commons.synth.children;
 
-import com.adobe.acs.commons.util.TypeUtil;
 import org.apache.commons.lang.StringUtils;
 import org.apache.sling.api.resource.ModifiableValueMap;
 import org.apache.sling.api.resource.PersistenceException;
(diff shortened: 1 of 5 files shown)


justinedelson:
What's the use case? I think I get it, but I'm not 100% sure.

davidjgonzalez:
@justinedelson its used to model flat trees or high-volume data in a package-able format. Allows for the simplicity of a flat-structure (since you could shard it to make it less flat) and packages very quickly, while maintaining a custom (via comparator) order.

For example, you might have 250 products, each with 5 variations. Each variation has daily data extending back 10 years; that's  250 * 5 * 10 * 365 = 4,562,500 nodes. Serializing to a property all there to be 250 * 5 = 1250 nodes. Trying to package up 4562500 nodes for replication (or even a subset) tends to be fairly costly. This is also nice in that it does inundate Oak Indexes w 10ks ~> millions of new nodes to deal with, flip side is this data isnt very Oak searchable.

Admittedly this is a niche use-case.

davidjgonzalez:
Im going to change the API to align more w the "usual" Resource apis. 

Any suggestions are welcome ass this will require a decent amount of adjustments.

PR #619, opened 2016-01-13
Add ability for the sitemap to be configured to include asset files from the DAM, such as PDFs.

HitmanInWis:
Add ability for the sitemap to be configured to include asset files from the DAM, such as PDFs.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
@@ -20,15 +20,15 @@
 package com.adobe.acs.commons.wcm.impl;
 
 import java.io.IOException;
-import java.util.Calendar;
-import java.util.Iterator;
-import java.util.Map;
+import java.util.*;
 
 import javax.servlet.ServletException;
 import javax.xml.stream.XMLOutputFactory;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamWriter;
 
+import com.day.cq.dam.api.Asset;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.time.FastDateFormat;
 import org.apache.felix.scr.annotations.Activate;
 import org.apache.felix.scr.annotations.Component;
@@ -39,6 +39,7 @@ import org.apache.felix.scr.annotations.Reference;
 import org.apache.felix.scr.annotations.Service;
 import org.apache.sling.api.SlingHttpServletRequest;
 import org.apache.sling.api.SlingHttpServletResponse;
+import org.apache.sling.api.resource.Resource;
 import org.apache.sling.api.resource.ResourceResolver;
 import org.apache.sling.api.resource.ValueMap;
 import org.apache.sling.api.servlets.SlingSafeMethodsServlet;
@@ -78,13 +79,21 @@ public final class SiteMapServlet extends SlingSafeMethodsServlet {
             description = "If true, the last modified value will be included in the sitemap.")
     private static final String PROP_INCLUDE_LAST_MODIFIED = "include.lastmod";
 
-    @Property(label = "Change Frequency Properties",
+    @Property(label = "Change Frequency Properties", unbounded = PropertyUnbounded.ARRAY,
             description = "The set of JCR property names which will contain the change frequency value.")
     private static final String PROP_CHANGE_FREQUENCY_PROPERTIES = "changefreq.properties";
 
-    @Property(label = "Priority Properties",
+    @Property(label = "Priority Properties", unbounded = PropertyUnbounded.ARRAY,
             description = "The set of JCR property names which will contain the priority value.")
     private static final String PROP_PRIORITY_PROPERTIES = "priority.properties";
+
+    @Property(label = "DAM Folder Property",
+            description = "The JCR property name which will contain DAM folders to include in the sitemap.")
+    private static final String PROP_DAM_ASSETS_PROPERTY = "damassets.property";
+
+    @Property(label = "DAM Asset MIME Types", unbounded = PropertyUnbounded.ARRAY,
+            description = "MIME types allowed for DAM assets.")
+    private static final String PROP_DAM_ASSETS_TYPES = "damassets.types";
 
     private static final String NS = "http://www.sitemaps.org/schemas/sitemap/0.9";
 
@@ -99,6 +108,10 @@ public final class SiteMapServlet extends SlingSafeMethodsServlet {
 
     private String[] priorityProperties;
 
+    private String damAssetProperty;
+
+    private List<String> damAssetTypes;
+
     @Activate
     protected void activate(Map<String, Object> properties) {
         this.externalizerDomain = PropertiesUtil.toString(properties.get(PROP_EXTERNALIZER_DOMAIN),
@@ -106,6 +119,8 @@ public final class SiteMapServlet extends SlingSafeMethodsServlet {
         this.includeLastModified = PropertiesUtil.toBoolean(properties.get(PROP_INCLUDE_LAST_MODIFIED), DEFAULT_INCLUDE_LAST_MODIFIED);
         this.changefreqProperties = PropertiesUtil.toStringArray(properties.get(PROP_CHANGE_FREQUENCY_PROPERTIES), new String[0]);
         this.priorityProperties = PropertiesUtil.toStringArray(properties.get(PROP_PRIORITY_PROPERTIES), new String[0]);
+        this.damAssetProperty = PropertiesUtil.toString(properties.get(PROP_DAM_ASSETS_PROPERTY), "");
+        this.damAssetTypes = Arrays.asList(PropertiesUtil.toStringArray(properties.get(PROP_DAM_ASSETS_TYPES), new String[0]));
     }
 
     @Override
@@ -131,12 +146,51 @@ public final class SiteMapServlet extends SlingSafeMethodsServlet {
                 write(children.next(), stream, resourceResolver);
             }
 
+            if (damAssetTypes.size() > 0 && damAssetProperty.length() > 0) {
+                for (Resource assetFolder : getAssetFolders(page, resourceResolver)) {
+                    writeAssets(stream, assetFolder, resourceResolver);
+                }
+            }
+
             stream.writeEndElement();
 
             stream.writeEndDocument();
         } catch (XMLStreamException e) {
             throw new IOException(e);
         }
+    }
+
+    private Collection<Resource> getAssetFolders(Page page, ResourceResolver resolver) {
+        Map<String, Resource> allAssetFolders = new TreeMap<String, Resource>();
+        ValueMap properties = page.getProperties();
+        String[] configuredAssetFolderPaths = properties.get(damAssetProperty, String[].class);
+        if (configuredAssetFolderPaths != null) {
+            for (String configuredAssetFolderPath : configuredAssetFolderPaths) {
+                if (StringUtils.isNotBlank(configuredAssetFolderPath)) {
+                    Resource configuredAssetFolder = resolver.getResource(configuredAssetFolderPath);
+                    if (configuredAssetFolder != null && configuredAssetFolder.isResourceType("sling:OrderedFolder")) {
+                        for (Resource assetFolder : getAssetFolders(configuredAssetFolder)) {
+                            if (!allAssetFolders.containsKey(assetFolder.getPath())) {
+                                allAssetFolders.put(assetFolder.getPath(), assetFolder);
+                            }
+                        }
+                    }
+                }
+            }
+        }
+        return allAssetFolders.values();
+    }
+
+    private List<Resource> getAssetFolders(Resource assetFolder) {
+        List<Resource> damFolders = new ArrayList<Resource>();
+        damFolders.add(assetFolder);
+        for (Iterator<Resource> children = assetFolder.listChildren(); children.hasNext();) {
+            Resource assetFolderChild = children.next();
+            if (assetFolderChild.isResourceType("sling:OrderedFolder")) {
+                damFolders.addAll(getAssetFolders(assetFolderChild));
+            }
+        }
+        return damFolders;
     }
 
     private void write(Page page, XMLStreamWriter stream, ResourceResolver resolver) throws XMLStreamException {
@@ -159,6 +213,41 @@ public final class SiteMapServlet extends SlingSafeMethodsServlet {
         writeFirstPropertyValue(stream, "priority", priorityProperties, properties);
 
         stream.writeEndElement();
+    }
+
+    private void writeAsset(Asset asset, XMLStreamWriter stream, ResourceResolver resolver) throws XMLStreamException {
+        stream.writeStartElement(NS, "url");
+
+
+        String loc = externalizer.externalLink(resolver, externalizerDomain, asset.getPath());
+        writeElement(stream, "loc", loc);
+
+        if (includeLastModified) {
+            long lastModified = asset.getLastModified();
+            if (lastModified > 0) {
+                writeElement(stream, "lastmod", DATE_FORMAT.format(lastModified));
+            }
+        }
+
+        final ValueMap properties = asset.adaptTo(Resource.class).getChild("jcr:content").getValueMap();
+        writeFirstPropertyValue(stream, "changefreq", changefreqProperties, properties);
+        writeFirstPropertyValue(stream, "priority", priorityProperties, properties);
+
+        stream.writeEndElement();
+    }
+
+    private void writeAssets(final XMLStreamWriter stream, final Resource assetFolder, final ResourceResolver resolver)
+            throws XMLStreamException {
+        for (Iterator<Resource> children = assetFolder.listChildren(); children.hasNext();) {
+            Resource assetFolderChild = children.next();
+            if (assetFolderChild.isResourceType("dam:Asset")) {
+                Asset asset = assetFolderChild.adaptTo(Asset.class);
+
+                if (damAssetTypes.contains(asset.getMimeType())) {
+                    writeAsset(asset, stream, resolver);
+                }
+            }
+        }
     }
 
     private void writeFirstPropertyValue(final XMLStreamWriter stream, final String elementName, final String[] propertyNames,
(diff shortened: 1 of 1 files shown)


HitmanInWis:
Oops...did not mean to submit this yet...sorry.  Wish I could delete, but I'll just close for now.

venturewise:
I am looking forward to this :-)

HitmanInWis:
Awesome, we'll have it up for PR shortly - just doing internal review first.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -78,14 +79,22 @@
             description = "If true, the last modified value will be included in the sitemap.")
     private static final String PROP_INCLUDE_LAST_MODIFIED = "include.lastmod";
 
-    @Property(label = "Change Frequency Properties",
```
HitmanInWis: Change Frequency Properties and Priority Properties were existing in the current impl, and were being treated in code as multi-value, but the configs themselves were only simple strings. As such, updated to make them configurable as multiple values.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -78,14 +79,22 @@
...
             description = "The set of JCR property names which will contain the priority value.")
     private static final String PROP_PRIORITY_PROPERTIES = "priority.properties";
 
+    @Property(label = "DAM Folder Property",
+            description = "The JCR property name which will contain DAM folders to include in the sitemap.")
+    private static final String PROP_DAM_ASSETS_PROPERTY = "damassets.property";
+
+    @Property(label = "DAM Asset MIME Types", unbounded = PropertyUnbounded.ARRAY,
+            description = "MIME types allowed for DAM assets.")
+    private static final String PROP_DAM_ASSETS_TYPES = "damassets.types";
```
HitmanInWis: This allows the servlet to be configured to only allow certain file types, such as "application/pdf" to include just PDF files.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -139,6 +160,39 @@ protected void doGet(SlingHttpServletRequest request, SlingHttpServletResponse r
         }
     }
 
+    private Collection<Resource> getAssetFolders(Page page, ResourceResolver resolver) {
+        Map<String, Resource> allAssetFolders = new TreeMap<String, Resource>();
+        ValueMap properties = page.getProperties();
+        String[] configuredAssetFolderPaths = properties.get(damAssetProperty, String[].class);
+        if (configuredAssetFolderPaths != null) {
+            for (String configuredAssetFolderPath : configuredAssetFolderPaths) {
```
HitmanInWis: For each configured DAM Asset folder, find all subfolders and add to the list as well. Code filters out duplicates so that if a user specifies two folders where one is a child of the other the assets are not added to the sitemap twice.

HitmanInWis:
Do I need to create a corresponding PR on the https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io for documentation of the new configs?

davidjgonzalez:
@HitmanInWis exactly!

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -139,6 +160,39 @@ protected void doGet(SlingHttpServletRequest request, SlingHttpServletResponse r
         }
     }
 
+    private Collection<Resource> getAssetFolders(Page page, ResourceResolver resolver) {
```
davidjgonzalez: Just to make sure i understand .. this is getting the list of DAM Asset "folder" paths, and then traversing their sub-trees, collecting all the folders under them, which (the folder list) is used later to collect all the assets?

WDYT about forgoing this collection and just walk the resource tree from this list of root folders?

https://github.com/Adobe-Consulting-Services/acs-aem-samples/blob/master/bundle/src/main/java/com/adobe/acs/samples/resources/SampleResourceVisitor.java

https://docs.adobe.com/content/docs/en/cq/5-6-1/javadoc/index.html?org/apache/sling/api/resource/AbstractResourceVisitor.html

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -139,6 +160,39 @@ protected void doGet(SlingHttpServletRequest request, SlingHttpServletResponse r
...
+        Map<String, Resource> allAssetFolders = new TreeMap<String, Resource>();
+        ValueMap properties = page.getProperties();
+        String[] configuredAssetFolderPaths = properties.get(damAssetProperty, String[].class);
+        if (configuredAssetFolderPaths != null) {
+            for (String configuredAssetFolderPath : configuredAssetFolderPaths) {
+                if (StringUtils.isNotBlank(configuredAssetFolderPath)) {
+                    Resource configuredAssetFolder = resolver.getResource(configuredAssetFolderPath);
+                    if (configuredAssetFolder != null && configuredAssetFolder.isResourceType("sling:OrderedFolder")) {
```
justinedelson: why is this only sling:OrderedFolder nodes?

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -139,6 +160,39 @@ protected void doGet(SlingHttpServletRequest request, SlingHttpServletResponse r
...
+        }
+        return allAssetFolders.values();
+    }
+
+    private List<Resource> getAssetFolders(Resource assetFolder) {
+        List<Resource> damFolders = new ArrayList<Resource>();
+        damFolders.add(assetFolder);
+        for (Iterator<Resource> children = assetFolder.listChildren(); children.hasNext();) {
+            Resource assetFolderChild = children.next();
+            if (assetFolderChild.isResourceType("sling:OrderedFolder")) {
```
justinedelson: why is this only sling:OrderedFolder nodes?

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -161,6 +215,44 @@ private void write(Page page, XMLStreamWriter stream, ResourceResolver resolver)
...
+        writeElement(stream, "loc", loc);
+
+        if (includeLastModified) {
+            long lastModified = asset.getLastModified();
+            if (lastModified > 0) {
+                writeElement(stream, "lastmod", DATE_FORMAT.format(lastModified));
+            }
+        }
+
+        Resource contentResource = asset.adaptTo(Resource.class).getChild("jcr:content");
```
justinedelson: I think (but worth doublechecking) that asset.adaptTo(ValueMap.class) would work here and be a bit cleaner.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
justinedelson: good catch

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -20,15 +20,15 @@
 package com.adobe.acs.commons.wcm.impl;
 
 import java.io.IOException;
-import java.util.Calendar;
-import java.util.Iterator;
-import java.util.Map;
+import java.util.*;
```
justinedelson: IIRC, our checkstyle rules technically prohibit the use of star imports...

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -78,14 +79,22 @@
...
     private static final String PROP_CHANGE_FREQUENCY_PROPERTIES = "changefreq.properties";
 
-    @Property(label = "Priority Properties",
+    @Property(label = "Priority Properties", unbounded = PropertyUnbounded.ARRAY,
             description = "The set of JCR property names which will contain the priority value.")
     private static final String PROP_PRIORITY_PROPERTIES = "priority.properties";
 
+    @Property(label = "DAM Folder Property",
+            description = "The JCR property name which will contain DAM folders to include in the sitemap.")
+    private static final String PROP_DAM_ASSETS_PROPERTY = "damassets.property";
```
justinedelson: Just so I'm clear, the way this works is that on the root page (or whatever page is being requested for the site map), this property get set to point to one or more asset folders. Makes sense, although I think it would be better to make this an array of property names, similar to the priority and change frequency ones.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
justinedelson: Just thinking out loud... does this make sense to configure at the servlet level or the page level (next to the folders). I can't think of a reason why it would matter one way or the other, just thought I'd throw it out there.

justinedelson: Oooh, but I see you have to set this. Could we treat * (or something like that) as a special value to allow all?

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -161,6 +215,44 @@ private void write(Page page, XMLStreamWriter stream, ResourceResolver resolver)
...
+    }
+
+    private void writeAssets(final XMLStreamWriter stream, final Resource assetFolder, final ResourceResolver resolver)
+            throws XMLStreamException {
+        for (Iterator<Resource> children = assetFolder.listChildren(); children.hasNext();) {
+            Resource assetFolderChild = children.next();
+            if (assetFolderChild.isResourceType("dam:Asset")) {
+                Asset asset = assetFolderChild.adaptTo(Asset.class);
+
+                if (damAssetTypes.contains(asset.getMimeType())) {
```
justinedelson: WDYT about making this a bit more flexible so I could specify something like image/* (for all image mimetypes) ? I guess we could always add that later as it would be totally backwards compatible.

justinedelson:
This generally looks good. Of my comments, only the one about the node type restriction is critical. The rest (especially the mime-type handling) could be handled later.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
davidjgonzalez: I assume because  I ootb DAM folders are ordered- but agree. IMO the Folders aren't what's interesting; it should be walking the tree for dam:Assets..  The collection method is double duty: first collecting all sub folders and then iterate over this folder list for assets. .. This is what using the resource visitor comment (above) is about.

justinedelson: Yes, but they can be made sling:Folder nodes by way of a checkbox. If we need to be doing any node type check here (which to your point may be avoidable), we should see if the node type extends nt:folder. Which unfortunately can only be done with the JCR API (I think).

davidjgonzalez: right - forgot that was added. either way - agree. 

I cant think of a good reason not to traverse and key off dam:Asset making the type the containers nodes irrelevant (its either a dam:Asset or it isn't)

davidjgonzalez: made a PR to the orig here: https://github.com/HS2-SOLUTIONS/acs-aem-commons/pull/2

that uses a resource visitor and removes the need to check for the containing node type.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
HitmanInWis: It's the only folder type I saw in the DAM.  I tried "nt:folder" and "sling:Folder" but apparently neither of those are supertypes of "sling:OrderedFolder", as neither worked.

Alternatively, I didn't want to arbitrarily just check a node for children and assume it is a folder based on that, to avoid things like subassets or whatever other nodes might be in the DAM.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
HitmanInWis: I went super-restrictive out of paranoia :)
Considering it is a configurable OSGI config, a new MIME can always on the fly by admins as needed.

Do you think it's desirable enough to have a wildcard that we should get it in the first cut?  Otherwise I agree that someone could always add the option later.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
HitmanInWis: I see the comments above now too, regarding simply checking all children if they are DAM Assets.  I suppose that would work...just didn't know what constructs other than folders would contain Assets AND authors would expect those items to then be included in the sitemap.

One thing worth noting ... I collected asset folders first so that I could sift out duplicates.  I could do it all in one shot (filtering out duplicate assets as we go) this just felt a little more clean in that here we gather all the folders and handle dupes, then the other code simply handles just writing the assets (similar to how code is "just writing" the pages) to the sitemap.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
justinedelson: > Do you think it's desirable enough to have a wildcard that we should get it in the first cut? 
Dunno. Let's skip it for now and see what the feedback is.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
HitmanInWis: Answered the wildcard comment below.  Let's continue the discussion down there.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
HitmanInWis: I did consider that actually, especially since priority/frequency have multiple values.  I decided against it because it felt like a property that an engineer is going to have to add a specific field to pages for, and there would never be a "real" need for mutliple different names for that property.  That said, for the sake of consistency I'd be happy to make the change if you think it's better that way.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
HitmanInWis: Good question on whether accepted MIME types should be configured at the page level or the servlet level.  I think for a first cut configuration at the servlet level is a good start for simplicity sake.  If we allow filtering MIME types at the folder level, then we'd have to handle things like a parent folder configured one way with a child folder configured a different way.  If someone needs that level of functionality we could update this later in a backward-compatible way by making the OSGI configuration the default and allowing overrides at the page level.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
HitmanInWis: Responded to the ResourceVisitor refactor on https://github.com/HS2-SOLUTIONS/acs-aem-commons/pull/2.  Let's continue the conversation there.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
davidjgonzalez: Duplicates? is this if there are multiple Dam Paths configures with one being under another .. ex.

For instance, do you mean if the sitemap was configured with 2 DAM Paths:

* /content/dam/demo/documents/brochures
* /content/dam/demo/documents

I expect it would be quicker to do the check on the list of paths to traverse first. In this case, you wouldnt traverse `/content/dam/demo/documents/brochures` since you have a ancestor rule to traverse `/content/dam/demo/documents`

I just worry that these traversals could get expensive on large DAMs, so would be nice to minimize as much as possible.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
HitmanInWis: Gave it a shot, but asset.adaptTo(ValueMap.class) is returning null.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
HitmanInWis: Yes, that's the exact scenario I'm preventing.  But you're absolutely right in that I should be able to simply look at the paths and filter out the dupes.  Seems like a good idea to prevent some unnecessary traversal.  I'll look into it.

davidjgonzalez: Good call on that corner case though; i hadnt thought of it but could totally see it happening.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
HitmanInWis: This is now resolved.  Configured IntelliJ to behave better :)

HitmanInWis:
Regarding the check of `sling:OrderedFolder` type...are we thinking that I should omit that check all together, and simply traverse children of any nodes as if they are folders (regardless of their actual node type)?

justinedelson:
I think we have to remove the node type check. It's possible to create `sling:Folder` nodes by unchecking the "Re-orderable" checkbox in the UI and it is legal to have `nt:folder` nodes. I think doing a general traversal, stopping at `dam:Asset` nodes (so that you don't get subassets) makes sense.

HitmanInWis:
Latest commit removes the check for `sling:OrderedFolder` type.  Note that this does allow a user to specify the path of a DAM asset as a "folder", which would then allow the subassets of that particular asset to be picked up.  But that's really no different than before where a user could specify a subassets folder directly.

HitmanInWis:
Added PR for documentation as well: https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io/pull/39

justinedelson:
Merged. Thanks!

PR #627, opened 2016-01-18
Nicer solution to support both scr pre 2.0 and scr 2.0 for disabling

### LINKED ISSUE #622: With newest Felix SCR, ACS AEM Commons doesn't resolve
The latest (2.0.0) version of Felix SCR removes the `org.apache.felix.scr` package in favor of the standardized `org.osgi.service.component.runtime` package. As a result, ACS AEM Commons no longer resolves (since we import this bundle). We should find a way to support this while being cross-compatible.

kwin:
components.
This is an alternative to the solution relying on ECMAScript.

This closes #622.


justinedelson:
This PR is not backwards compatible and introduces new API, neither of which should be necessary.

PR #628, opened 2016-01-18
provide non-ECMAScript mechanism for choosing the correct implementat…

justinedelson:
…ion for the ComponentDisabler


justinedelson:
fair enough. check out the changes in cc8d5b1

PR #630, opened 2016-01-18
Feature/607 oak ensure index

joerghoh:
Added more features:
* the automatic application of an EnsureOakIndex definition can be disabled (default: enabled)
* there is the EnsureOakIndexManager which control the EnsureOakIndex instances

These features allow you to have a much better control of the EnsureOakIndex process; for example you can apply them based on a scheduling.

* MBean support


joerghoh:
this pull request contains way too much individual commits, obviously the branch has not been setup correctly. I will close it and create a new one with only the relevant pieces.

PR #631, opened 2016-01-19
More features for EnsureOakIndex

joerghoh:
Compared with an earlier pull request I fixed my local history (obviously I applied some changes to my local master which I was no longer aware of).

Nevertheless, here some new features for the EnsureOakIndex functionality:
* An EnsureOakIndex instance can be configured not apply the Oak Index immediately
* The individual EnsureOakIndex service instances can be controlled through the EnsureOakIndexManager to apply the instances now.

This gives you much more control, when you need to split the deployment of the EnsureOakIndex definition from the actual application of this change.

* MBean support. YAY!

Many thanks to Davide for review and refactoring!


joerghoh:
I rebased this branch, merging should work

davidjgonzalez:
LGTM! well let Travis give it the :+1: and merrrrge!

PR #638, opened 2016-01-22
Add a Sling Models 1.2 compliant injector-specific annotation.

### LINKED ISSUE #470: Provide an annotation for the AEM Objects Sling Models Injector
That way errors with the deployment become more obvious, as the annotation will then not be resolvable. Also there is no fallback to other injectors.

kwin:
All dependencies are marked optional to make sure the bundle still runs with AEM 6.0.

This fixes #470

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/AemObjectInjector.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/AemObjectInjector.java
@@ -76,7 +76,7 @@ import java.lang.reflect.Type;
  * pre-configured from the current request.
  */
 @Property(name = Constants.SERVICE_RANKING, intValue = 4500)
-public final class AemObjectsInjector implements Injector {
+public final class AemObjectInjector implements Injector {
 
     private static final String COM_DAY_CQ_WCM_TAGS_DEFINE_OBJECTS_TAG = "com.day.cq.wcm.tags.DefineObjectsTag";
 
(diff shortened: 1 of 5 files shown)


bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/AemObjectInjector.java
```diff
@@ -76,7 +76,7 @@
  * pre-configured from the current request.
  */
 @Property(name = Constants.SERVICE_RANKING, intValue = 4500)
-public final class AemObjectsInjector implements Injector {
+public final class AemObjectInjector implements Injector {
```
justinedelson: Is there a reason for this name change?

kwin:
Some comments:
I renamed the AemObjectsInjector to AemObjectInjector to be more inline with the original SlingInjector (and its annotation, injector).
This injector is only compatible with Sling Models 1.2 which is not shipped until AEM 6.2. If I would provide an injector which is compatible with Sling Models 1.1 (shipped with AEM 6.1) that one would not support default injection strategies properly (https://issues.apache.org/jira/browse/SLING-4155).
@justinedelson WDYT?

justinedelson:
So with this change, what happens when you install ACS AEM Commons on an AEM instance with Sling Models 1.0 or 1.1 installed?

kwin:
With Sling Models 1.0 or 1.1 the annotation processor factory would just not be available and therefore the annotation would be ignored. I had to mark the according dependencies as optional in the manifest.

davidjgonzalez:
@justinedelson any outstanding concerns with this or can we merge?

justinedelson:
This cannot be merged into master as-is. Doing so makes ACS AEM Commons uninstallable on an otherwise unmodified 6.1 instance due to

    org.apache.sling.models.spi,version=[1.1,2) -- Cannot be resolved

justinedelson:
Merged into the 6.2-compatibility branch. Not sure how this can be made compatible with 6.1, but if it can, we can address on master at a later point in time. Closing this.

PR #639, opened 2016-01-22
AEM 61 - Touch UI Limit Parsys

schoudry:
@davidjgonzalez @justinedelson 

Demo - https://drive.google.com/file/d/0B4d6KmbLkAumREJndHhZZURLXzA/view?usp=sharing

1) Set the component limit "acsComponentsLimit" (manually in CRX) on parsys design node

image

2) Error notification when user tries drag/drop, copy/paste or insert

image

i thought showing error notification is a better UX compared to disabling drop areas; let me know what you think....

### DIFF
--- a/content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/js.txt
+++ b/content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/js.txt
@@ -4,3 +4,4 @@ touchui-widgets-init.js
 touchui-composite-multifield.js
 touchui-composite-multifield-nodestore.js
 icon-picker.js
+touchui-limit-parsys.js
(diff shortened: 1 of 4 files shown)


justinedelson:
Is there a way to get this functionality also working in ClassicUI using the same design property?

justinedelson:
(with a reasonable level of effort)

schoudry:
@justinedelson we can, but avoided assuming there is no future for classic ui, should i add the classic one as well?

justinedelson:
well, the only reason I was thinking about it is that without supporting the ClassicUI, this restriction is really easy to workaround :)

schoudry:
@justinedelson  added the classic checks

image

imrul001:
Is there anyway to disable the drop zone instead of showing error message?

kiranbabuveeranki:
In AEM6.2 touch ui it is working fine.Can you explain how to achieve this with AEM 6.3 touch UI?

Issue #605, opened 2015-12-21
"mvn clean install -Panalysis" does not work with Java 1.7


joerghoh:
I still run Java 1.7 as default on my system, especially for maven. When I want to run the analysis profile, the maven build breaks:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:2.9:javadoc (default) on project acs-aem-commons-bundle: An error has occurred in JavaDocs report generation:
[ERROR] Exit code: 1 - javadoc: error - invalid flag: -Xdoclint:none
[ERROR]
[ERROR] Command line was: /Library/Java/JavaVirtualMachines/jdk1.7.0_75.jdk/Contents/Home/bin/javadoc @options @packages
[ERROR]
[ERROR] Refer to the generated Javadoc files in '/Users/jhoh/repos/acs-aem-commons/bundle/target/site/apidocs' dir.
[ERROR] -> [Help 1]
```

You can easily reproduce it like this (requires a OS X and Java 1.7 being installed):

    export JAVA_HOME=`/usr/libexec/java_home -v 1.7` && mvn clean install -Panalysis

We should document, that this only works with Java 1.8

PR #640, opened 2016-01-25
remove java8 specific javadoc-parameters from analysis profile

joerghoh:
Fixing #605 
Now the analysis profile should work with java7 again.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -161,8 +161,7 @@
                     <configuration>
                         <excludePackageNames>
                             *.impl
-                        </excludePackageNames>
-                        <additionalparam>-Xdoclint:none</additionalparam>
+                        </excludePackageNames>          
                     </configuration>
                 </plugin>
                 <plugin>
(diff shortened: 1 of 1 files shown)


bstopp:
Great name for the profile.

justinedelson:
Hmmm. JDK 7 has been EOL for 8 months. can't we just switch the minimum JDK version to 1.8?

bstopp:
There are still a large number of companies which haven't shifted over to JDK 8, even with the EOL out there, for whatever reason.

justinedelson:
in d77cd77, I implemented a different solution which solves the same problem but without requiring the inclusion of a different profile on the command line.

joerghoh:
@justinedelson I don't really see the huge difference here ... My patch did not require to provide a profile on the command line, but rather activated the profile when you're running on java 8. Pretty much the same as your patch does.

justinedelson:
sorry - my bad. I misread your PR. I guess the change is identical, except for the formatting :)

Issue #677, opened 2016-03-10
Touch ui limit parsys issue: Granite.author is undefined


ahmed-musallam:
tested on AEM 6.1 with SP1 installed.
There is an issue with the js file at: `/apps/acs-commons/touchui-widgets/source/touchui-limit-parsys.js`
error in browser console where Granite.author is undefined. The issue is because the file loads before the initialization of Granite.author

this issue can be easily fixed by adding: `window.Granite.author = window.Granite.author || {};` to the top of the file `touchui-limit-parsys.js`
Note that this line is used in many places  (OOTB) to insure we don't encounter undefined errors
eg: `/libs/cq/gui/components/authoring/clientlibs/editor`

paul-bjorkstrand:
I think the underlying issue is that the touchui-widgets clientlib does not explicitly depend on the AEM OOTB clientlib that provides Granite.author. In this specific case, it is causing the touchui-widgets library to load too early. This in turn, causes `Granite.author` to be undefined when passed into the IIFE in touchui-limit-parsys. In the end, the trying to read the property `gAuthor.ui` causes an error, because `gAuthor` is undefined.

paul-bjorkstrand:
I made an alternate pull request to address this in a slightly better way. Rather than just short-circuiting when `gAuthor` or its properties are undefined/null, set it to the actual reference of `Granite.author`, creating an object if necessary.

PR #678, opened 2016-03-10
fix issue #677

ahmed-musallam:
fix for limit_parysys which breaks in AEM 6.1 SP1 because Granite.Author is undefined

### DIFF
--- a/content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/source/touchui-limit-parsys.js
+++ b/content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/source/touchui-limit-parsys.js
@@ -23,6 +23,7 @@
  * eg. to limit the components to 4 on rightpar of /content/geometrixx/en.html
  * set acsComponentsLimit=4 on /etc/designs/geometrixx/jcr:content/homepage/rightpar
  */
+ window.Granite.author = window.Granite.author || {};
 (function ($, $document, gAuthor) {
     "use strict";
 


content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/source/touchui-limit-parsys.js
```diff
@@ -23,6 +23,7 @@
  * eg. to limit the components to 4 on rightpar of /content/geometrixx/en.html
  * set acsComponentsLimit=4 on /etc/designs/geometrixx/jcr:content/homepage/rightpar
  */
+ window.Granite.author = window.Granite.author || {};
```
schoudry: @ahmed-musallam i think it still breaks in extendComponentDrop() with gAuthor.edit.actions undefined

this extension should be loaded after Granite.author is available, otherwise its unusable

ahmed-musallam: @schoudry there are no guarantees that Granite.author will be available before the extension anyway, since it is not a dependency of the clientlib. This fix will at least pass the reference to Granite.authoring. By the time the functions in the extension are called, Granite.author would have been loaded. again, no guarantees there, but I'd assume the extension wont be usable untill the page has loaded.

so, there are tow solutions here,
1. wait for document ready, then have the extension available. or
2. move this file into it's own clientlib and make sure that it's loaded after Granite.author, through some dependency

davidjgonzalez:
Replaced by #680

PR #680, opened 2016-03-14
#677 fix2

ahmed-musallam:
this is a better fix for #677  where we check if Granite.author and some of it's properties are truthy or falsy and return accordingly.

The best solution would be to move  the touchui-limit-parsys to it's own clientlib and add the correct dependencies to it to ensure proper loading order. I have tried doing that but wasn't very successful in finding the exact dependencies needed.


davidjgonzalez:
@ahmed-musallam what are the uses cases where these dependencies arent loaded? If they arent, then this functionality just wont work (since the checks will avoid executing the functionality).

Its curious the normal touchui authoring dependencies arent sufficient...?

paul-bjorkstrand:
@davidjgonzalez, see #688 for an alternate solution.

davidjgonzalez:
Replaced by #688

PR #687, opened 2016-04-12
fix compatibility of touchui widgets throwing errors in 6.0

justinedelson:
@schoudry this was reported in gitter. I think this change just disabled these features in 6.0 when the JS objects aren't defined.


schoudry:
@justinedelson LGTM

PR #691, opened 2016-04-15
F A C E   M E L T E R !

badvision:
This is a major feature for running deferred actions in a managed thread pool.  The actions can also be throttled to limit max cpu/memory consumption as well as auto-kill any long-running threads.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/functions/Function.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/functions/Function.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2016 Adobe.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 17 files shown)


bundle/src/main/java/com/adobe/acs/commons/functions/Function.java
```diff
@@ -0,0 +1,115 @@
+/*
+ * Copyright (c) 2010, 2013, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
```
justinedelson: pretty sure we can't include this. Isn't there a Function class in Guava which can be reused...

badvision: Those were corrected.  Good catch.

badvision:
I built an asset ingestion routine on top of FaceMelter and found some easy avenues to improve things.  The Retry action is a good action wrapper to provide auto-retry support in case of Oak Merge conflicts, etc.

PR #693, opened 2016-04-19
allow for clientlibrary versioning to be disabled through OSGi config

justinedelson:


justinedelson:
@cpilsworth please take a look at this change. I think it is what you were describing offline.

cpilsworth:
LGTM

Issue #692, opened 2016-04-17
Path Based Render Condition


kaushalmall:
Out of the box, there's really only two render conditions available. Privilege and simple. Although simple is pretty powerful as you can provide an EL expression, but it doesn't something as simple as "show this menu item under path /content/geometrixx/*". For this we can use a path based render condition. 

Is this something that'll be useful? WDYGT? @justinedelson @davidjgonzalez

davidjgonzalez:
So this would be a convenience method for using el on requestPathinfo.resourcePath matches XXX?

I can see this being a very common use case. Are you thinking a startsWith check or a refer matches? Also could multiple path patterns be specified or would an OR rendition condition be used?

kaushalmall:
>So this would be a convenience method for using el on requestPathinfo.resourcePath matches XXX?

Yes. 

>Are you thinking a startsWith check or a refer matches? 

I was thinking something like below. 

`boolean vote = false;
String suffix = slingRequest.getRequestPathInfo().getSuffix();
    if (suffix != null) {
        vote = suffix.matches(pathRegex);
    }`

>Also could multiple path patterns be specified or would an OR rendition condition be used?

You can use OR.

kaushalmall:
Where should I put the render condition? Under /apps/acs-commons/authoring?

davidjgonzalez:
Per our offline discussion i think your suggestion of: `acs-commons/granite/ui/components/renderconditions/path` is great. aligns well w the product pathing.

justinedelson:
+1

PR #695, opened 2016-04-20
adding path regular expression render condition. Issue #692

kaushalmall:
Sample rendercondition node would look something like 

`<rendercondition
                                jcr:primaryType="nt:unstructured"
                                sling:resourceType="acs-commons/granite/ui/components/renderconditions/path"
                                path="/content/geometrixx/.*"/>`

### DIFF
--- a/content/src/main/content/jcr_root/apps/acs-commons/granite/ui/components/renderconditions/path/path.jsp
+++ b/content/src/main/content/jcr_root/apps/acs-commons/granite/ui/components/renderconditions/path/path.jsp
@@ -1,0 +1,50 @@
+<%--
+  #%L
+  ACS AEM Commons Package
+  %%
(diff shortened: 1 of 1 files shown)


kaushalmall:
@justinedelson @davidjgonzalez it says the CI build failed, but I am not sure what that has to do with my PR. Is this something I can ignore?

justinedelson:
I've seen this a few times. @davidjgonzalez any ideas?

>   MonthlyExpiresHeaderFilterTest.testDoActivateSuccess:220 Values should be different. Actual: 2
>   WeeklyExpiresHeaderFilterTest.testDoActivateSuccess:175 Values should be different. Actual: 2

bstopp:
I'll take a look, I think there may be a timing issue.

Issue #699, opened 2016-04-22
EnvironmentIndicator should not be displayed with wcmmode=disabled


joerghoh:
To have a consistent preview on author, the environment indicator should not be displayed when the wcmmode=disabled is set. With the current version it is always injected into the output.

This would allow to use this feature also on publish instances, where it's useful for CRXDE Lite and Webconsole.

davidjgonzalez:
@joerghoh dont a number of authoring console have wcmmode set to disabled?

Are you suggesting this on requests for /editor.html and /cf only? 

Would `preview` fall into this blacklist bucket as well?

joerghoh:
@davidjgonzalez not sure about the authoring consoles. I got the feedback especially for /cf and wcmmode=disabled, but for preview it should work as well.
But thanks for the hint, wasn't thinking of that yet.

davidjgonzalez:
Added in 2.5.0/3.1.0 - thanks @joerghoh

PR #706, opened 2016-04-28
Indicators might not appear in certain WCMModes

joerghoh:
WCMModes can be configured where the EnvironmentIndicators don't appear. Is mostly useful for PREVIEW and DISABLED modes (see #699)


joerghoh:
@davidjgonzalez thanks for the review, I pushed some updates

PR #707, opened 2016-05-01
Feature/queryautocompletelist

kaushalmall:
Query autocomplete datasoruce.

  A datasource that takes a list of predicates and a display property to build out the autocomplete list.
  The path of the resource is saved in the JCR and the display property is used for better UX.

```
  /**
   * The predicate map.
   */
  - predicates (String[]) = type=cq:Page,path=/content/geometrixx,property=jcr:content/sling:resourceType,property.value=geometrixx/components/homepage

   /**
   * The display property name.
   */
   -displayProperty (String) = jcr:content/jcr:title.
```

PR #711, opened 2016-05-03
Search based PathBrowser

schoudry:
@justinedelson @davidjgonzalez 

Pathbrowser widget for showing search (query builder) results

1) Set the _queryParameters_  property and _sling:resourceType_ - _/apps/acs-commons/components/authoring/search-pathbrowser_

image

2) Results in widget

image


justinedelson:
As discussed offline, I think we should make this consistent with #707 and configure it with a generic predicates String[] property.

schoudry:
changed to use "predicates"

PR #718, opened 2016-05-19
invalidate references in httpcache

YegorKozlov:
httpcache should support invalidation of references. 
Suppose Page A prints title of Page B, e.g. via the List component. User loads Page A and it gets into the httpcache.  At some point of time the title of Page B changes and then user loads Page A  again.  The problem with the current implementation is that the user will get stale data with the old title because cache invalidation is not cascade.  

the logic should be as follows:
1. invalidate the resource
2. find all references of the given resource and invalidate them. 
For #2 we can use com.day.cq.wcm.commons.ReferenceSearch to find all pages that reference the given path.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/httpcache/invalidator/HttpCacheInvalidationJobConsumer.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/httpcache/invalidator/HttpCacheInvalidationJobConsumer.java
@@ -26,6 +26,7 @@ import org.apache.commons.lang.StringUtils;
 import org.apache.felix.scr.annotations.*;
 import org.apache.sling.api.resource.ResourceResolver;
 import org.apache.sling.api.resource.ResourceResolverFactory;
+import org.apache.sling.commons.osgi.PropertiesUtil;
 import org.apache.sling.event.jobs.Job;
 import org.apache.sling.event.jobs.consumer.JobConsumer;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


davidjgonzalez:
Like it - wdyt about enabling reference invalidation via osgi? My concern is that reference search query could be expensive under certain circumstances; would be good to have a way to avoid the load if needed.

@sivaramt wdyt?

Sivaramvt:
Good find. This issue would occur when invalidation job is created as part of application code instead of the provided JCR node change event handler.
Agree @davidjgonzalez. Reference searches are known to be expensive. A switch through OSGi config gives us flexibility.

YegorKozlov:
I made it configurable in OSGi. In my world this flag is always true but I understand there can be systems that would want to avoid the overhead on searching references.  The default is false (invalidation of references is diabled). 
To enable this feature user need to add a configuration file 
com.adobe.acs.commons.httpcache.invalidator.HttpCacheInvalidationJobConsumer.xml with the following contents:

`
<jcr:root xmlns:sling="http://sling.apache.org/jcr/sling/1.0" xmlns:cq="http://www.day.com/jcr/cq/1.0"
          xmlns:jcr="http://www.jcp.org/jcr/1.0" xmlns:nt="http://www.jcp.org/jcr/nt/1.0"
          httpcache.config.invalidation.references="{Boolean}true"
/>`

YegorKozlov:
@davidjgonzalez  Hey David. This is Yegor, we worked on the CEB WebCQ project for a while back in 2013.

davidjgonzalez:
Not to clutter this PR thread ...but im gonna anyhow :)

@YegorKozlov when i saw your name come in i thought it was you, and then i saw it was as PR for a more advanced feature and thought "its totally yegor"  and then i looked at your profile picture and thought "i dont remember him having that hair! maybe its not him!"

glad to see your still in the space -- you will working w Vadim? Hit me on gchat or gitter.im (https://gitter.im/Adobe-Consulting-Services/acs-aem-commons) to catch up!

YegorKozlov:
that's me :) I still work with Vadim for CEB.  Evaluating the aem-commons stuff for their needs . More patches are coming.

Issue #728, opened 2016-06-02
Dates and times should follow ISO 8601


ovelindstrom:
Dates and time, like in com.adobe.acs.commons.email.process.impl.SendTemplatedEmailUtilsTest.mockJcrProperties(), should use the ISO 8601 as format.

(Alternatively, the Locale corresponding to the used format should be used when creating a new Calendar but that is not the right way to go for an international product.)

PR #729, opened 2016-06-02
Using ISO 8601 format for dates. Solves #728

ovelindstrom:
Just using a SDF and Calendar item that is not dependent on a US/UK Locale being set.


justinedelson:
Thanks!

PR #730, opened 2016-06-07
implementation of a factory for HttpClient Executor/Request objects

justinedelson:
This pull request replaces #478 with the following changes:

* Removed StringResponseHandler as that functionality already exists in HttpClient.
* Refactored JsonObjectResponseHandler (which was renamed to reflect that it handles JSON objects, not arbitrary JSON) to use the BasicResponseHandler to do the String extraction from the response.
* Added factory.name configuration option to actually find the configured factories from other OSGi components.
* Remove Jackson-based response handler as that seemed to be incompatible with AEM 6.2 (it can always be added later)
* Implemented relaxed SSL cert checking (it was configurable but not actually implemented)
* Misc. code cleanup and formatting.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/http/JsonObjectResponseHandler.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/http/JsonObjectResponseHandler.java
@@ -1,0 +1,60 @@
+/*
+ * #%L
+ * ACS AEM Commons Bundle
+ * %%
+ * Copyright (C) 2016 Adobe
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+package com.adobe.acs.commons.http;
+
+import org.apache.commons.io.IOUtils;
+import org.apache.http.HttpResponse;
+import org.apache.http.client.ClientProtocolException;
+import org.apache.http.client.ResponseHandler;
+import org.apache.http.impl.client.BasicResponseHandler;
+import org.apache.sling.commons.json.JSONException;
+import org.apache.sling.commons.json.JSONObject;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.Charset;
+
+/**
+ * Converts response to a JSON Object.
+ */
+public class JsonObjectResponseHandler implements ResponseHandler<JSONObject> {
+
+    private BasicResponseHandler handler = new BasicResponseHandler();
+
+    @Override
+    public JSONObject handleResponse(HttpResponse httpResponse) throws
+            ClientProtocolException, IOException {
+        String json = handler.handleResponse(httpResponse);
+        if (json == null) {
+            return null;
+        }
+        try {
+            return new JSONObject(json);
+        } catch (JSONException e) {
+            JSONObject jsonObject = new JSONObject();
+            try {
+                jsonObject.put("error", e.getMessage());
+            } catch (JSONException e1) {
+            }
+            return jsonObject;
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


rohanraj14:
Hi Justin/David,

I am facing issue building my project jar in 6.2. And can see that while you also faced similar build error:
https://travis-ci.org/Adobe-Consulting-Services/acs-aem-commons/builds/133937577 

Can you please let me know the fix for it:

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:2.3.2:compile (default-compile) on project sandiskweb.core: Compilation failure
[ERROR] error: error reading M:\mvn-repo\com\adobe\aem\uber-jar\6.2.0\uber-jar-6.2.0-apis.jar; zip file is empty

Rohan

justinedelson:
@davidjgonzalez @schoudry any comments on this?

bundle/src/main/java/com/adobe/acs/commons/http/JsonObjectResponseHandler.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+public class JsonObjectResponseHandler implements ResponseHandler<JSONObject> {
+
+    private BasicResponseHandler handler = new BasicResponseHandler();
+
+    @Override
+    public JSONObject handleResponse(HttpResponse httpResponse) throws
+            ClientProtocolException, IOException {
+        String json = handler.handleResponse(httpResponse);
+        if (json == null) {
```
schoudry: Use StringUtils.isEmpty(json) ? http responses are more likely to be empty than null?

bundle/src/main/java/com/adobe/acs/commons/http/impl/HttpClientFactoryImpl.java
```diff
@@ -0,0 +1,181 @@
...
+@Component(
+        label = "ACS AEM Commons - Http Components Fluent Executor Factory",
+        description = "ACS AEM Commons - Http Components Fluent Executor Factory",
+        configurationFactory = true, policy = ConfigurationPolicy.REQUIRE)
+@Service
+@Property(label = "Factory Name", description = "Name of this factory", name = "factory.name")
+public class HttpClientFactoryImpl implements HttpClientFactory {
+    public static final boolean DEFAULT_USE_SSL = false;
+
+    public static final boolean DEFAULT_DISABLE_CERT_CHECK = false;
```
schoudry: just a thought, is it better to have final vars in HttpClientFactory? to call HttpClientFactory.DEFAULT_DISABLE_CERT_CHECK rather from implementation HttpClientFactoryImpl.DEFAULT_DISABLE_CERT_CHECK

bundle/src/main/java/com/adobe/acs/commons/http/impl/HttpClientFactoryImpl.java
```diff
@@ -0,0 +1,181 @@
...
+                public boolean isTrusted(X509Certificate[] arg0, String arg1) throws CertificateException {
+                    return true;
+                }
+            }).build();
+            builder.setHostnameVerifier(new AllowAllHostnameVerifier()).setSslcontext(sslContext);
+        }
+        httpClient = builder.build();
+        executor = Executor.newInstance(httpClient);
+
+        String username = PropertiesUtil.toString(config.get(PROP_USERNAME), null);
```
schoudry: use StringUtils.isNotEmpty(username) && StringUtils.isNotEmpty(password) ?

bundle/src/main/java/com/adobe/acs/commons/http/JsonObjectResponseHandler.java
davidjgonzalez: Good Q .. if the response body is blank, then handler.handleResponse(httpResponse) returns null.... if the response is blank, do we want an empty JSON obj or do we want to return null? guess thats up the contract of this API; i could go either way as long its documented.

bundle/src/main/java/com/adobe/acs/commons/http/JsonObjectResponseHandler.java
```diff
@@ -0,0 +1,60 @@
...
+            return null;
+        }
+        try {
+            return new JSONObject(json);
+        } catch (JSONException e) {
+            JSONObject jsonObject = new JSONObject();
+            try {
+                jsonObject.put("error", e.getMessage());
+            } catch (JSONException e1) {
+            }
```
davidjgonzalez: Should at least  log.error(..)  --- Should we be setting a 500 response in this case?

bundle/src/main/java/com/adobe/acs/commons/http/impl/HttpClientFactoryImpl.java
davidjgonzalez: are these really useful outside the context of this impl? If we do expose these on the interface level, then we cant as easily change them (due to the implication that they are consumer usable defaults)

davidjgonzalez:
other than the minor comments, LGTM!

PR #741, opened 2016-06-25
Content traversing workflows - Synthetic WF and Replication w Options

davidjgonzalez:
This PR contains mechanics to support Content Traversing workflows. This is useful in AEM 6.2 as WF Folders can now be selected for applying Workflow to.

* Content Visitor - Walks Page or Asset trees looking for cq:Pages and dam:Assets to process.

* SyntheticWorkflowWrapper - WF Process that executes a param WF Model as Synthetic WF. Allows for easy execution of Synth WF via normal WF UIs though does require invocation of a WF. Can be used to traverse content trees.

* Replication w Options - performs a Content Traversing replication w replication options passed in via PROCESS_ARGS

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/ReplicateWithOptionsWorkflowProcess.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/ReplicateWithOptionsWorkflowProcess.java
@@ -71,6 +71,7 @@ public class ReplicateWithOptionsWorkflowProcess implements WorkflowProcess {
     private static final String ARG_REPLICATION_ACTION_TYPE = "replicationActionType";
     private static final String ARG_REPLICATION_SYNCHRONOUS = "synchronous";
     private static final String ARG_REPLICATION_SUPPRESS_VERSIONS = "suppressVersions";
+    private static final String ARG_THROTTLE = "throttle";
 
     @Reference
     private WorkflowPackageManager workflowPackageManager;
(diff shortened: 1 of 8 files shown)


bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/ReplicateWithOptionsWorkflowProcess.java
```diff
@@ -0,0 +1,164 @@
...
+    public void execute(WorkItem workItem, WorkflowSession workflowSession, MetaDataMap metaDataMap) throws WorkflowException {
+        ResourceResolver resourceResolver = null;
+
+        try {
+            resourceResolver = workflowHelper.getResourceResolver(workflowSession);
+            final String originalPayload = (String) workItem.getWorkflowData().getPayload();
+            final List<String> payloads = workflowPackageManager.getPaths(resourceResolver, originalPayload);
+            final ProcessArgs processArgs = new ProcessArgs(metaDataMap);
+
+            final AtomicInteger count = new AtomicInteger(0);
```
bstopp: This is created/instantiated, but never consumed.

bstopp:
The _SyntheticWrapperWorkflowProcess_ and _ReplicateWithOptionsWorkflowProcess_ have a lot of shared/duplicated logic; is there no way to remove some of the C/P/D risk?

davidjgonzalez:
@bstopp agree on the code dupe; started pulling it out but realized 1) it really seems useful to only these 2 use cases  2) its the same but every other line is slightly different and i'd have to make a bunch of partial abstractions, which creates a bunch of supporting files for 2 otherwise simple WFs

bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/ReplicateWithOptionsWorkflowProcess.java
davidjgonzalez: Fixed

PR #747, opened 2016-07-01
Updates to FAM actions utility functions

badvision:
- Complete rewrite of how ActionManager manages resource resolvers
- Identify assets with missing or outdated renditions
- Activate content to a list of publishers in round-robin manner (parallelizes publishing where available)
- Translation function to apply actions against renditions (useful for activating or deleting specific renditions)

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/fam/impl/ActionManagerImpl.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/fam/impl/ActionManagerImpl.java
@@ -20,10 +20,10 @@ import com.adobe.acs.commons.fam.Failure;
 import com.adobe.acs.commons.fam.ThrottledTaskRunner;
 import com.adobe.acs.commons.functions.*;
 import java.util.ArrayList;
-import java.util.Deque;
+import java.util.Collections;
 import java.util.List;
-import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 import javax.jcr.NodeIterator;
 import javax.jcr.RepositoryException;
 import javax.jcr.Session;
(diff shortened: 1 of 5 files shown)


bundle/src/main/java/com/adobe/acs/commons/fam/impl/ActionManagerImpl.java
```diff
@@ -131,25 +147,26 @@ public void accept(ResourceResolver r) throws Exception {
...
-                        t.commit();
-                        t.close();
-                    } catch (PersistenceException ex) {
-                        LOG.error("Persistence exception closing session pool", ex);
+        Runnable r = new Runnable() {
+            @Override
+            public void run() {
+                while (!isComplete()) {
+                    try {                    
+                        Thread.sleep(100);
```
davidjgonzalez: Curious, whats the significance of this magic number?

bundle/src/main/java/com/adobe/acs/commons/fam/impl/ActionManagerImpl.java
```diff
@@ -298,10 +287,13 @@ public CompositeData getStatistics() throws OpenDataException {
     public void closeAllResolvers() {
         if (!resolvers.isEmpty()) {
             for (ReusableResolver resolver : resolvers) {
-                resolver.getResolver().close();
+                if (resolver.getResolver().isLive()) {
```
davidjgonzalez: This should be closed regardless of `.isLive()` check .. .javadocs are misleading. see @cziegeler note on the sling lists: 

http://apache-sling.73963.n3.nabble.com/Using-the-resource-resolver-td4061854.html

davidjgonzalez: Will fix this when i use it.

badvision:
Seriously?   Wow.  I thought that isLive() indicates if the session is open and false = closed.  Yeah, please change it for sure if you think it will be ok.

bundle/src/main/java/com/adobe/acs/commons/fam/impl/ActionManagerImpl.java
badvision: 100ms?  Nothing terribly significant, I just want to make sure that we're not in a tight wait-loop while other threads complete work and waiting several seconds for that to happen is almost certain.

Issue #742, opened 2016-06-27
EmailService to send the attachments


vivekdhiman:
Do we have any way out to add the attachment with email send via EmailService?

kaushalmall:
@justinedelson @davidjgonzalez we wrote an EmailServiceWithAttachment service based on the ACS Commons EmailService. If this is needed, we can submit a PR.

justinedelson:
@kaushalmall this is the first I've heard of this requirement, but I'm sure it would be a nice addition. Is a new service necessary or could this be a new method on the existing service?

kaushalmall:
we could add a new method, we have the code, so most of the work is done already.

shamalroy:
I worked with @kaushalmall on this and will submit a PR soon.

PR #748, opened 2016-07-06
Feature/email service to send attachment #742

shamalroy:


bundle/src/main/java/com/adobe/acs/commons/email/impl/EmailServiceImpl.java
```diff
@@ -123,6 +136,72 @@
         return failureList;
     }
 
+    @Override
+    public List<InternetAddress> sendEmail(String templatePath, Map<String, String> emailParams, Map<String, DataSource> attachments, InternetAddress... recipients) {
```
bstopp: There any reason the _sendEmail_ method w/o attachments can't call this with an empty map? The method appears to already handle that case, and it would consolidate all of the logic to a single(-ish) point.

Thoughts?

davidjgonzalez: +1 lots of copy/pasta otherwise. Ill merge this and make an issue to consolidate.

PR #750, opened 2016-07-06
HttpCache - ResourceType Include based caching

davidjgonzalez:
Augments HttpCache to support Include-based caching. Major changes

1) Added a OSGi Property to HttpCacheConfigImpl that selects the Filter Scope (REQUEST, INCLUDE), defaults to REQUEST as that it the current state.
2) Added a ResourceTypeHttpCacheConfigExtension that caches based on the resource path; currently does not support selector/extensions/suffixes.
* Restrictable by content path regex, example: /content/.*/jcr:content/footer/.* and by ResourceType (via resource.getResourceType())
3) Minors enhancements to the Mbeans to give more visibility.

REQUEST and INCLUDES can work together, and the INCLUDES work in a russian-doll stacking manner.

The Filters (Request, Include) are only enabled if at least 1 of CacheConfig exists for that scope.

/cc @Sivaramvt, @YegorKozlov, @steffenrosi

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/httpcache/config/HttpCacheConfig.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/httpcache/config/HttpCacheConfig.java
@@ -37,7 +37,7 @@ import java.util.regex.Pattern;
 @ProviderType
 public interface HttpCacheConfig {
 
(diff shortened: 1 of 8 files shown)


davidjgonzalez:
Declining for merge and squash

PR #776, opened 2016-08-16
fixing issue where composite multifield not displayed correctly in pa…

### LINKED ISSUE #715: TouchUI Composite MultiField not working when accessed through action bar properties
When accessing page properties via the page (Open > Open properties), 
image
image

a nested multifield is working: the saved values are loaded correctly and the nested multifield is displayed.
multifield-on-page

When accessing the same page properties via the site console (View properties), 
image

the nested multifield is not working anymore: the values are not loaded and any nesting is not shown.
multifield-on-site

We found out that the `dialog-ready` event is not fired in the site view dialog and even when it would be fired, the form submit listener would never be added to the  form, since the form does not have a `cq-dialog` class, which seems required.

**/apps/acs-commons/touchui-widgets/source/touchui-composite-multifield.js**
```javascript
  $document.ready(function () {
       addDataInFields();

       $document.on("dialog-ready", function() {
           var dialog = document.querySelector('form.cq-dialog');

           if(dialog){
               dialog.addEventListener('submit', collectDataFromFields, true);
           }
       });
   });
```

Could it be that some JavaScript libraries/events are not available in the site view, which are present in the page view?
If so, can this be fixed, or is this a limitation?

justinedelson:
…ge properties. fixes #715

@schoudry please review

### DIFF
--- a/content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/source/touchui-composite-multifield-nodestore.js
+++ b/content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/source/touchui-composite-multifield-nodestore.js
@@ -96,10 +96,6 @@
 
     var _ = window._, Class = window.Class;
 
-    function onPropertiesPage() {
-        return $("form#cq-sites-properties-form").length === 1;
-    }
-
     ACS.TouchUI.NodeCompositeMultiField = new Class({
         toString: 'ACS TouchUI Composite Multifield Store as Nodes',
         extend: ACS.TouchUI.Widget,
@@ -252,7 +248,7 @@
     $document.ready(function () {
         var compositeMultiField = new ACS.TouchUI.NodeCompositeMultiField();
 
-        if (onPropertiesPage()) {
+        if (ACS.TouchUI.Util.onPropertiesPage()) {
             compositeMultiField.addDataInFields();
 
             $document.on("click", "[form=cq-sites-properties-form]", function(){
--- a/content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/source/touchui-composite-multifield-nodestore.js
+++ b/content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/source/touchui-composite-multifield-nodestore.js
@@ -96,6 +96,10 @@
 
     var _ = window._, Class = window.Class;
 
+    function onPropertiesPage() {
+        return $("form#cq-sites-properties-form").length === 1;
+    }
+
     ACS.TouchUI.NodeCompositeMultiField = new Class({
         toString: 'ACS TouchUI Composite Multifield Store as Nodes',
         extend: ACS.TouchUI.Widget,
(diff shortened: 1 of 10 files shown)


content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/source/touchui-composite-multifield-nodestore.js
```diff
@@ -248,12 +252,20 @@
     $document.ready(function () {
         var compositeMultiField = new ACS.TouchUI.NodeCompositeMultiField();
 
-        $document.on("dialog-ready", function(){
+        if (onPropertiesPage()) {
```
schoudry: i think we can move this common function to touchui-widgets-init.js and call it from both multifield files (json and node store) 

`compositeMultiField.isOnPropertiesPage()`

schoudry:
@justinedelson one comment regd. moving the common function (onProperitesPage) to init file, rest all looks good

@kaushalmall the fix for your amex issue

justinedelson:
@schoudry look good now?

schoudry:
Yes looks good

kaushalmall:
@justinedelson a few issues with this. It works, if 

- you open an existing page and and add values in the multi field and then open the properties page again. 

It doesn't work if you add values in the createpagewizard because 

1.  The `/apps/acs-commons/touchui-widgets` clientlib only has categories `cq.authoring.dialog,acs-commons.cq.authoring.dialog`. These categories never get loaded on http://localhost:4508/mnt/overlay/wcm/core/content/sites/createpagewizard.html/content/geometrixx. The node `/libs/wcm/core/content/sites/createpagewizard/jcr:content/head/clientlibs` only loads `coralui2, granite.ui.coral2.foundation, granite.ui.coral.foundation.addon.coral2` categories. So, we can add granite.ui.coral.foundation.addon.coral2 category to the node `/apps/acs-commons/touchui-widgets`. 

2. But, there is another issue, the code 

```
$document.on("click", "[form=cq-sites-properties-form]", function(){
                compositeMultiField.collectDataFromFields();
            });
```
will never get fired either, because on http://localhost:4508/mnt/overlay/wcm/core/content/sites/createpagewizard.html/content/geometrixx, the DOM element `form=cq-sites-properties-form` doesn't exist. The "Create" button element looks like 
```
<button class="foundation-layout-inline2-item foundation-wizard-control coral-Button coral-Button--primary" coral-wizardview-next="" data-foundation-wizard-control-action="next" type="submit" autocomplete="off" is="coral-button" variant="primary" size="M"><coral-button-label class="coral-Button-label">Create</coral-button-label></button>
```
These look like very generic values to me, so I am not sure what we can hook onto for the click action. 

But, I believe if we fix these two issues, it'll work fine.

justinedelson:
@kaushalmall thanks for the feedback. I will take a look, but my immediate inclination is to say that usage in the Create Page Wizard isn't a supported use case and that this only works for existing pages.

justinedelson:
actually, this wasn't as bad as I thought. 

new build: https://files.acrobat.com/a/preview/c5594218-ce20-4810-adca-3d147f85fefb

PR #787, opened 2016-08-31
[Mohit Bansal] Adding hide in sitemap feature. Page marked as "Hide i…

mkbansal:
Adding hide in sitemap feature. Page marked as "Hide in navigation" will be displayed in sitemap. So if you want to hide any page from sitemap then use "hide in sitemap" property.

Hide in navigation is primarily used for navigation only but still this page is valid and should be available for search engine so seeking need of additional property to hide from sitemap.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
@@ -53,6 +53,7 @@ import org.apache.sling.api.servlets.SlingSafeMethodsServlet;
 import org.apache.sling.commons.osgi.PropertiesUtil;
 
 import com.day.cq.commons.Externalizer;
+import com.day.cq.wcm.api.NameConstants;
 import com.day.cq.wcm.api.Page;
 import com.day.cq.wcm.api.PageFilter;
 import com.day.cq.wcm.api.PageManager;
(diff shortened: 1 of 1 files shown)


justinedelson:
Thanks. This looks good, but we shouldn't break backwards compatibility. Please add a configuration flag to disable checking the hide in nav field.

davidjgonzalez:
@justinedelson, @mkbansal - i could see this going 2 ways, add another checkbox that toggles an explicit hideInNav check..  or set the default property value of `hideinsitemap.property` to be `hideInNav` .. If its the default value, then you couldnt check BOTH properties (since its a single value property).  I assume its more sane to have a single property that effects the inclusion vs understanding that a field "Hide in Site Map" *AND* the less semantic "Hide In Nav" can have the same effect.

mkbansal:
@davidjgonzalez , @justinedelson - Your recommendation definitely make sense. I have updated code to support backward compatibility. 
Please note that i have added a condition 
if(StringUtils.isBlank(this.hideProperty)) {
   	this.hideProperty = NameConstants.PN_HIDE_IN_NAV;
}
instead of 
this.hideProperty = PropertiesUtil.toString(properties.get(PROP_HIDE_IN_SITEMAP), NameConstants.PN_HIDE_IN_NAV);

as PropertiesUtil.toString() consider EMPTY value as valid string and saving runmode with empty value will be considered as valid property.

davidjgonzalez:
@mkbansal thanks; so if someone doesnt want to limit what's included in the site map (including `hideInNav` they can add a non-existant property name like `showAllContentInSiteMap` in the OSGi config.

mkbansal:
Thankyou @davidjgonzalez for merging this fix. Please let me know when you will publish the new version of ACSCommons?
Also, who is the right person to update Sitemap documentation with this change?

davidjgonzalez:
@mkbansal I'll try to release this Monday or Tuesday - I can take care of the docs

PR #792, opened 2016-09-14
#783 Fixes IE11 dismissible system notifications

### LINKED ISSUE #783: Dismissable System Notifications redisplayed after visiting Admin page IE 11
Tested on a vanilla AEM 6.1 SP1 instance with ACS AEM Commons 2.6.10:

With IE 11:
1. Start AEM with SP1, ACS Commons and Geometrixx installed
2. Create a dismissable System Notification
3. Visit http://localhost:4502/sites.html. Notification is shown -> Dismiss
4. Open a content page like http://localhost:4502/editor.html/content/geometrixx/en/products.html Notification is shown -> Dismiss
5. Return to the admin page. Notification is shown -> Dismiss

- The same happens using Classic UI.
- With Chrome and Firefox the problem does not occur.

I struggle to debug the issue becuase of the poor Developer Tools in IE but I think the cookie to remember the dismissed notficiations is not set in IE 11.

davidjgonzalez:
Fixed #783 - IE11 cannot dismiss dismissible notifications

### DIFF
--- a/content/src/main/content/jcr_root/apps/acs-commons/components/utilities/system-notifications/notification/clientlibs/notification.js
+++ b/content/src/main/content/jcr_root/apps/acs-commons/components/utilities/system-notifications/notification/clientlibs/notification.js
@@ -50,9 +50,9 @@ $(function() {
     });
     
     function setSessionCookie(name, value) {
-        document.cookie = name + '=' + value + '; expires=0; path=/;';
+        document.cookie = name + '=' + value + '; expires=Tue, 01 Jan 2999 12:00:00 UTC; path=/;';
     }
-    
+
     function getCookieValue(name) {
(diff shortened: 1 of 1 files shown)


content/src/main/content/jcr_root/apps/acs-commons/components/utilities/system-notifications/notification/clientlibs/notification.js
```diff
@@ -50,9 +50,9 @@ $(function() {
     });
     
     function setSessionCookie(name, value) {
```
justinedelson: minor nit - this function is incorrectly named. We aren't setting a "session" cookie. Quite the opposite; we are setting a far future cookie.

PR #800, opened 2016-09-29
flexibile externalizer domains for sitemap

shsteimer:
TL;DR: allow one registration of the sitemap servlet to be addressed at different paths, and use different externalizer domains for each path.

A small change to the sitemap servlet to allow for different externalizer domains at different paths by using an inherited page property.  If either that osgi config prop is left blank, or if the page property doesn't exist, then it will still fall back to using the static externalizer domain in the OSGI config.

I'm not sure how generally useful this is, but I've had this scenarios a few times where I've wanted to use this feature, but couldn't because I had various language copies of a site (same template resource type), but with different externalizer domains (ie. en.mysite.com & fr.mysite.com).  This change allows for that type of setup more easily.

Note that this doesn't currently handle using this new osgi property for externalizing assets, only for pages.


cwoeltge:
Hi Sean, 

did you consider to create /etc/map entries instead? That works for all cases I came across so far. 

Cheers
©

bstopp:
I like the idea and i think it would definitely be useful as i've run into the same problem.

I'm not fan of using /etc/map for _anything_ because it introduces more complexity than is necessary. 

Thoughts on using a OSGi config, mapping incoming hosts to externalizer domains instead of an authored attribute? I am only attempting to try to keep the configurations managed in the same way. Externalizers are a OSGi config; shouldn't defining this also be? 

If kept as an authored item, can you add a ExtJS and Touch UI Input fields for consumers to reference, similar to all the widgets listed [here](https://adobe-consulting-services.github.io/acs-aem-commons/features/widgets.html)?

justinedelson:
@bstopp @shsteimer I tend to agree with @cwoeltge that `/etc/map` is the right solution to this *assuming* that this domain/path relationship is something which spans beyond just the sitemap. Is there a real use case for specifying an alternate domain name *only* for the sitemap? Because I would imagine that you also need to rewrite "internal" links (which `/etc/map` handles already) and (probably) the canonical meta tag. Otherwise, you can easily end up in a state where the URL pointed to in the sitemap and the URLs linked to from the site's HTML will be different thus spreading & diluting your "link juice".

bstopp:
@justinedelson That makes more sense; i made a poor assumption on the goal/use case and how it differs from the OOTB features.

shsteimer:
I didn't realize that /etc/map worked in that way.  I guess I made a poor assumption about how the externalize#externalize call would function when given a domain name if /etc/map entries were in place so came to the incorrect conclusion that I needed both /etc/map and externalizer. TIL.

Closing.

PR #816, opened 2016-10-13
Feature/smtp health check

### LINKED ISSUE #815: Sling Health Check for SMTP access
A near universal integration for AEM projects is configuring an SMTP server in the AEM Mail Service for sending emails. This can be cumbersome to test as it requires engaging in an activity that sends email (ex. Workflow, Registration, etc.)

There should be a Sling Health Check that reports if the SMTP server can be reached and a test email sent.

davidjgonzalez:
Resolves #815 

TBD if the AEM Tools - Operations - Health Checks nodes should be added as part of this PR

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/hc/impl/SMTPMailServiceHealthCheck.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/hc/impl/SMTPMailServiceHealthCheck.java
@@ -1,4 +1,4 @@
-package com.adobe.acs.commons.healthchecks.impl;
+package com.adobe.acs.commons.hc.impl;
 
 import com.day.cq.commons.mail.MailTemplate;
 import com.day.cq.mailer.MessageGateway;
@@ -88,20 +88,12 @@ public class SMTPMailServiceHealthCheck implements HealthCheck {
                     try {
                         messageGateway.send(email);
                         resultLog.info("The E-mail Service appears to be working properly. Verify the health check e-mail was sent to [ {} ]", this.email);
-
                     } catch (Exception e) {
                         resultLog.critical("Failed sending e-mail. Unable to send a test email via the configured E-mail server: " + e.getMessage(), e);
                         log.error("Failed to send E-mail for E-mail Service health check", e);
                     }
 
-                    resultLog.info("SMTP Host: {}", email.getHostName());
-                    resultLog.info("SMTP use SSL: {}", email.isSSL());
-                    if (email.isSSL()) {
-                        resultLog.info("SMTP SSL Port: {}", email.getSslSmtpPort());
-                    } else {
-                        resultLog.info("SMTP Port: {}", email.getSmtpPort());
-                    }
-                    resultLog.info("SMTP From Address: {}", email.getFromAddress());
+                    logMailServiceConfig(resultLog, email);
                 } catch (Exception e) {
                     resultLog.healthCheckError("Sling Health check could not formulate a test email: " + e.getMessage(), e);
                     log.error("Unable to execute E-mail health check", e);
@@ -110,5 +102,16 @@ public class SMTPMailServiceHealthCheck implements HealthCheck {
         }
 
         return new Result(resultLog);
+    }
+
+    private void logMailServiceConfig(FormattingResultLog resultLog, SimpleEmail email) {
+        resultLog.info("SMTP Host: {}", email.getHostName());
+        resultLog.info("SMTP use SSL: {}", email.isSSL());
+        if (email.isSSL()) {
+            resultLog.info("SMTP SSL Port: {}", email.getSslSmtpPort());
+        } else {
+            resultLog.info("SMTP Port: {}", email.getSmtpPort());
+        }
+        resultLog.info("SMTP From Address: {}", email.getFromAddress());
     }
 }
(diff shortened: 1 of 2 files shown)


bundle/src/main/java/com/adobe/acs/commons/hc/impl/SMTPMailServiceHealthCheck.java
```diff
@@ -0,0 +1,117 @@
...
+                resultLog.critical("The AEM Default Mail Service is INACTIVE, thus e-mails cannot be sent.");
+                resultLog.info("Verify the Default Mail Service is active and configured: http://localhost:4502/system/console/components/com.day.cq.mailer.DefaultMailService");
+                log.error("Could not retrieve a SimpleEmail Message Gateway");
+
+            } else {
+                try {
+                    List<InternetAddress> emailAddresses = new ArrayList<InternetAddress>();
+                    emailAddresses.add(new InternetAddress(this.email));
+                    MailTemplate mailTemplate = new MailTemplate(IOUtils.toInputStream(MAIL_TEMPLATE), CharEncoding.UTF_8);
+                    SimpleEmail email = mailTemplate.getEmail(StrLookup.mapLookup(new HashMap<String, String>()), SimpleEmail.class);
```
justinedelson: nit pick, but `Collections.emptyMap()` would be better optimized here.

bundle/src/main/java/com/adobe/acs/commons/hc/impl/SMTPMailServiceHealthCheck.java
```diff
@@ -0,0 +1,117 @@
...
+    protected void activate(Map<String, Object> properties) {
+        this.email = PropertiesUtil.toString(properties.get(PROP_EMAIL), DEFAULT_EMAIL);
+    }
+
+    @Override
+    public Result execute() {
+        final FormattingResultLog resultLog = new FormattingResultLog();
+
+        if (messageGatewayService == null) {
+            resultLog.warn("MessageGatewayService OSGi service could not be found.");
```
justinedelson: why not error? Seems like the HC should fail if this service is null.

justinedelson: Although note that it will never be null since the `@Reference` above isn't optional.

bundle/src/main/java/com/adobe/acs/commons/hc/impl/SMTPMailServiceHealthCheck.java
```diff
@@ -0,0 +1,117 @@
...
+                    List<InternetAddress> emailAddresses = new ArrayList<InternetAddress>();
+                    emailAddresses.add(new InternetAddress(this.email));
+                    MailTemplate mailTemplate = new MailTemplate(IOUtils.toInputStream(MAIL_TEMPLATE), CharEncoding.UTF_8);
+                    SimpleEmail email = mailTemplate.getEmail(StrLookup.mapLookup(new HashMap<String, String>()), SimpleEmail.class);
+
+                    email.setSubject("AEM E-mail Service Health Check");
+                    email.setTo(emailAddresses);
+
+                    try {
+                        messageGateway.send(email);
```
justinedelson: Perhaps ignorance on my part, but is this a synchronous call, i.e. it will fail if the SMTP server is unreachable?

davidjgonzalez:
Example success: 

image

 Example failure: 

image

I did notice the timeouts arent great on socket failures. I in-code set the socket timeouts to 10s (which should be generous) however it appears the message gateway is retrying (not seeing where i can specify no retries)

bundle/src/main/java/com/adobe/acs/commons/hc/impl/SMTPMailServiceHealthCheck.java
davidjgonzalez: yes - see my issue with timeout below though. it keeps retrying, leading to long timeouts (not good for a healthcheck...

image

bundle/src/main/java/com/adobe/acs/commons/hc/impl/SMTPMailServiceHealthCheck.java
```diff
@@ -0,0 +1,124 @@
...
+        final FormattingResultLog resultLog = new FormattingResultLog();
+
+        if (messageGatewayService == null) {
+            resultLog.critical("MessageGatewayService OSGi service could not be found.");
+            resultLog.info("Verify the Default Mail Service is active: http://localhost:4502/system/console/components/com.day.cq.mailer.impl.CqMailingService");
+        } else {
+            final MessageGateway<SimpleEmail> messageGateway = messageGatewayService.getGateway(SimpleEmail.class);
+            if (messageGateway == null) {
+                resultLog.critical("The AEM Default Mail Service is INACTIVE, thus e-mails cannot be sent.");
+                resultLog.info("Verify the Default Mail Service is active and configured: http://localhost:4502/system/console/components/com.day.cq.mailer.DefaultMailService");
```
bstopp: See previous comment.

bundle/src/main/java/com/adobe/acs/commons/hc/impl/SMTPMailServiceHealthCheck.java
```diff
@@ -0,0 +1,124 @@
...
+        this.email = PropertiesUtil.toString(properties.get(PROP_EMAIL), DEFAULT_EMAIL);
+    }
+
+    @Override
+    public Result execute() {
+        final FormattingResultLog resultLog = new FormattingResultLog();
+
+        if (messageGatewayService == null) {
+            resultLog.critical("MessageGatewayService OSGi service could not be found.");
+            resultLog.info("Verify the Default Mail Service is active: http://localhost:4502/system/console/components/com.day.cq.mailer.impl.CqMailingService");
```
bstopp: Shouldn't this use an externalizer to generate a valid link?

PR #828, opened 2016-10-29
Feature/users to csv

### LINKED ISSUE #820: Users to CSV Report
There should be an easy way to export users, basic profile data, and their groups (declared and transitive) from AEM for governance and reporting purposes.

davidjgonzalez:
Resolves #820


davidjgonzalez:
Made some changes from original Issue.

1) Updated so export pages can be saved (so if you want to re-run the same report over time its easier)
2) Custom props can be specified via rel paths off the [rep:User] node

image

[users.export.csv.txt](https://github.com/Adobe-Consulting-Services/acs-aem-commons/files/559757/users.export.csv.txt)

PR #829, opened 2016-10-29
Feature/smtp health check

davidjgonzalez:
This should be off by default.

.. also an valid email address should be provided to receive the test emails.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/hc/impl/SMTPMailServiceHealthCheck.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/hc/impl/SMTPMailServiceHealthCheck.java
@@ -1,4 +1,4 @@
-package com.adobe.acs.commons.healthchecks.impl;
+package com.adobe.acs.commons.hc.impl;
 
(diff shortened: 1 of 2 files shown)


bundle/src/main/java/com/adobe/acs/commons/hc/impl/SMTPMailServiceHealthCheck.java
```diff
@@ -21,7 +21,9 @@
 
 @Component(metatype = true,
         label = "ACS AEM Commons - Health Check - SMTP E-Mail Service",
-        description = "Checks if the AEM E-Mail Service can connect and send mail via the configured SMTP server.")
+        description = "Checks if the AEM E-Mail Service can connect and send mail via the configured SMTP server.",
+        policy = ConfigurationPolicy.REQUIRE
```
justinedelson: @davidjgonzalez was this commit not merged intentionally?

davidjgonzalez: no

Issue #814, opened 2016-10-11
Inline editing Page Properties multifield doesn't work in Touch-UI (AEM 6.1 + ACS Commons 2.7.0)


rubnig:
I've added a multifield from the ACS Commons package in the page properties.

The dialog is working fine, when the page properties are opened up-on the page itself. Multifield properties are stored as a JSON object:

![image](https://cloud.githubusercontent.com/assets/8969453/19263941/0de53b7e-8f9f-11e6-8bb6-df473ba2c0b2.png)

![image](https://cloud.githubusercontent.com/assets/8969453/19263982/33081840-8f9f-11e6-8cc1-a2f604b0d532.png)

But when I edit the same page properties through the Touch UI menu, the multifield dialog shows up normally, but saves the properties in the wrong format. Not as a single JSON object but as two seperate properties:

![image](https://cloud.githubusercontent.com/assets/8969453/19264035/6fa891bc-8f9f-11e6-8a9f-0aa989b1dee7.png)

![image](https://cloud.githubusercontent.com/assets/8969453/19264046/7c634b36-8f9f-11e6-9048-772c86d5e268.png)

The multifield is initialized through the following code:

```
<facets
        jcr:primaryType="nt:unstructured"
        jcr:title="Facets"
        sling:resourceType="granite/ui/components/foundation/section">
    <layout
            jcr:primaryType="nt:unstructured"
            sling:resourceType="granite/ui/components/foundation/layouts/fixedcolumns"
            margin="{Boolean}false"/>
    <items jcr:primaryType="nt:unstructured">
        <column
                jcr:primaryType="nt:unstructured"
                sling:resourceType="granite/ui/components/foundation/container">
            <items jcr:primaryType="nt:unstructured">
                <title
                        jcr:primaryType="nt:unstructured"
                        jcr:title="Search Facets"
                        sling:resourceType="granite/ui/components/foundation/form/fieldset">
                    <items jcr:primaryType="nt:unstructured">
                        <facets_field
                                jcr:primaryType="nt:unstructured"
                                sling:resourceType="granite/ui/components/foundation/form/multifield"
                                class="full-width"
                                fieldDescription="Click '+' to add a new facet"
                                fieldLabel="Facets">
                            <field
                                    jcr:primaryType="nt:unstructured"
                                    sling:resourceType="granite/ui/components/foundation/form/fieldset"
                                    acs-commons-nested="JSON_STORE"
                                    name="./facets">
[...]

PR #830, opened 2016-10-31
Fix for issue 814 (ACS Commons Multifield in Touch UI Page Properties)

rubnig:
#814 occured due to a wrong detection of the edit properties page in AEM (http://localhost:4502/libs/wcm/core/content/sites/properties.html).

Added a new check for this specific page (so we don't interfere with the checks in place).


justinedelson:
Thanks

rubnig:
I've made the check more generic, more page were affected by this bug see #832

PR #840, opened 2016-11-08
Removed unused variable

josephrignanese:
I noticed that `mvn -Panalysis clean install` was failing due to an unused variable.


davidjgonzalez:
Thanks

PR #853, opened 2016-11-28
remove dependency to AbstractAssetWorkflowProcess. fixes #852

### LINKED ISSUE #852: AbstractAssetWorkflowProcess is changing for AEM 6.3
There are a some API changes coming to the AbstractAssetWorkflowProcess in AEM 6.3 which makes the current build of ACS AEM Commons fail to install and run. I think it would be best if we removed this dependency from our code and instead provided a helper class with the required methods.

justinedelson:
This code also has the advantage of removing a usage of the deprecated componentAbstract feature, thus cleaning up a few log entries along the way.

Note that anyone extending the AbstractRenditionModifyingProcess will need to refactor their code (thus the major package increase), but I don't think that is a very common use case.


bundle/src/main/java/com/adobe/acs/commons/dam/AssetWorkflowHelper.java
```diff
@@ -0,0 +1,181 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Utility methods which support the writing of asset-related workflow processes, largely to avoid a
+ * dependency on AbstractAssetWorkflowProcess which tends to change between AEM versions.
+ */
+public final class AssetWorkflowHelper {
```
justinedelson: Good point. I think it is best to consolidate. Updated the PR.

PR #860, opened 2016-12-01
added stylesheet inliner for emails

jshelley-xx:
When using AEM's output for html emails, inlining styles in <style> elements (instead) adds support for additional email clients. See https://www.campaignmonitor.com/css/.

Aside from adding the rewriter to the rewriter pipeline, an "inline-css" selector must be present to cause the transformation.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StylesheetInlinerTransformerFactory.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StylesheetInlinerTransformerFactory.java
@@ -1,0 +1,200 @@
+/*
+ * #%L
+ * ACS AEM Commons Bundle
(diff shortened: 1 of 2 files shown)


jshelley-xx:
I think I need to rebase this from feature/6.2-compatibility, or use the deprecated HtmlLibraryManager packages instead.  Thoughts?

jshelley-xx:
"ACS AEM Commons is expected to be installable on AEM 6.0 and AEM 6.1"

justinedelson:
@jshelley-xx I think this should be OK on the 6.2-compatibility branch.

justinedelson:
@jshelley-xx I think this is an interesting change. Let me know if you are still interesting in pursuing this for the 6.2 branch. I have some additional questions/comments.

PR #861, opened 2016-12-01
added stylesheet inliner for emails

jshelley-xx:
Resubmitting this PR to feature/6.2-compatibility.

====

When using AEM's output for html emails, inlining styles in <style> elements (instead) adds support for additional email clients. See https://www.campaignmonitor.com/css/.

Aside from adding the rewriter to the rewriter pipeline, an "inline-css" selector must be present to cause the transformation.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StylesheetInlinerTransformerFactory.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StylesheetInlinerTransformerFactory.java
@@ -19,19 +19,13 @@
  */
 package com.adobe.acs.commons.rewriter.impl;
 
-import java.io.CharArrayWriter;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.InputStreamReader;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
-import java.util.Map;
 
-import org.apache.commons.lang.StringUtils;
-import org.apache.felix.scr.annotations.Activate;
+import org.apache.commons.io.IOUtils;
 import org.apache.felix.scr.annotations.Component;
-import org.apache.felix.scr.annotations.Deactivate;
 import org.apache.felix.scr.annotations.Properties;
 import org.apache.felix.scr.annotations.Property;
 import org.apache.felix.scr.annotations.Reference;
@@ -63,15 +57,12 @@ import com.adobe.granite.ui.clientlibs.LibraryType;
         label = "Stylesheet Inliner Transformer Factory", 
         description = "Sling Rewriter Transformer Factory which inlines CSS references")
 @Properties({ 
-    @Property( name = "pipeline.type", value = StylesheetInlinerTransformerFactory.SELECTOR, propertyPrivate = true) } )
+    @Property( name = "pipeline.type", value = "inline-css", propertyPrivate = true) } )
 @Service(value = { TransformerFactory.class })
 public final class StylesheetInlinerTransformerFactory implements TransformerFactory {
 
-    public static final String SELECTOR = "inline-css";
-    
-    private static final String NEWLINE = "\n";
+    private static final char[] NEWLINE = new char[] {'\n'};
     private static final String STYLE = "style";
-    private static final String BODY = "body";
     private static final String HEAD = "head";
 
     private static final Logger log = LoggerFactory.getLogger(StylesheetInlinerTransformerFactory.class);
@@ -86,8 +77,7 @@ public final class StylesheetInlinerTransformerFactory implements TransformerFac
 
     private class CSSInlinerTransformer extends AbstractTransformer {
 
-        protected boolean enabled = false;
-        protected boolean inHead = false;
+        protected boolean afterHeadElement = false;
         protected List<String> stylesheetsInHead = new ArrayList<String>();
         private SlingHttpServletRequest slingRequest;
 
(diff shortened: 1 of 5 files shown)


bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StylesheetInlinerTransformerFactory.java
```diff
@@ -0,0 +1,200 @@
...
+
+        @Override
+        public void init(ProcessingContext context, ProcessingComponentConfiguration config) throws IOException {
+            super.init(context, config);
+            slingRequest = context.getRequest();
+
+            String[] selectors = slingRequest.getRequestPathInfo().getSelectors();
+
+            if (selectors.length > 0) {
+                enabled = Arrays.asList(selectors).contains(SELECTOR);
```
justinedelson: I'm not sure why this is necessary. Wouldn't you just this up in a separate pipeline which was triggered by the selector or just add this to the existing newsletter pipeline (default is in /libs/mcm/config/rewriter/campaign-link-rewrite)

bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StylesheetInlinerTransformerFactory.java
```diff
@@ -0,0 +1,200 @@
...
+                        }
+                    } else {
+                        getContentHandler().startElement(namespaceURI, localName, qName, attrs);
+
+                        if (localName.equalsIgnoreCase(HEAD)) {
+                            inHead = true;
+                        } else if (localName.equalsIgnoreCase(BODY)) {
+
+                            // add each of the accumulated stylesheet references
+                            for (String sheet : stylesheetsInHead) {
```
justinedelson: Would it make more sense to inline these just before the `</head>` tag?

bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StylesheetInlinerTransformerFactory.java
```diff
@@ -0,0 +1,200 @@
...
+        final String rel = attrs.getValue("", "rel");
+        final String type = attrs.getValue("", "type");
+        final String href = attrs.getValue("", "href");
+
+        return StringUtils.equals("link", elementName) && StringUtils.equals(rel, "stylesheet")
+                && StringUtils.equals(type, LibraryType.CSS.contentType) && StringUtils.startsWith(href, "/")
+                && !StringUtils.startsWith(href, "//") && StringUtils.endsWith(href, LibraryType.CSS.extension);
+    }
+
+    @Activate
```
justinedelson: these two methods can just be removed

bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StylesheetInlinerTransformerFactory.java
```diff
@@ -0,0 +1,200 @@
...
+        @Override
+        public void endElement(String uri, String localName, String qName) throws SAXException {
+            if (inHead && localName.equalsIgnoreCase(HEAD)) {
+                inHead = false;
+            }
+            super.endElement(uri, localName, qName);
+        }
+    }
+
+    private static boolean isCSS(final String elementName, final Attributes attrs) {
```
justinedelson: I would be ideal to find a way to share this method between this and the StaticReferenceRewriterTransformer.

bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StylesheetInlinerTransformerFactory.java
```diff
@@ -0,0 +1,200 @@
...
+
+                CharArrayWriter caw = new CharArrayWriter(inputStream.available());
+                caw.write(NEWLINE);
+                int read = -1;
+                while ((read = isr.read()) > -1) {
+                    caw.write(read);
+                }
+                getContentHandler().startElement(namespaceURI, STYLE, null, new AttributesImpl());
+                getContentHandler().characters(caw.toCharArray(), 0, caw.size());
+                super.endElement(namespaceURI, STYLE, null);
```
justinedelson: minor nit - for consistency, please make this `getContentHandler().endElement()`

bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StylesheetInlinerTransformerFactory.java
```diff
@@ -0,0 +1,200 @@
...
+
+                if (resource != null) {
+                    inputStream = resource.adaptTo(InputStream.class);
+                }
+            }
+
+            if (inputStream != null) {
+                InputStreamReader isr = new InputStreamReader(inputStream);
+
+                CharArrayWriter caw = new CharArrayWriter(inputStream.available());
```
justinedelson: I think it would be better to use` IOUtils.toCharArray()` here just to avoid the extra few lines of code. That would involve two calls to `characters()`. Or you could use `IOUtils.copy()` if you wanted to preserve the `CharArrayWriter`.

bundle/src/test/java/com/adobe/acs/commons/rewriter/impl/StylesheetInlinerTransformerFactoryTest.java
```diff
@@ -0,0 +1,286 @@
...
+import org.xml.sax.SAXException;
+import org.xml.sax.helpers.AttributesImpl;
+
+import com.adobe.granite.ui.clientlibs.HtmlLibrary;
+import com.adobe.granite.ui.clientlibs.HtmlLibraryManager;
+import com.adobe.granite.ui.clientlibs.LibraryType;
+
+import junitx.util.PrivateAccessor;
+
+@RunWith(MockitoJUnitRunner.class)
```
justinedelson: Great tests!!!

bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StylesheetInlinerTransformerFactory.java
jshelley-xx: I'd not seen that rewriters could be configured to match selectors.  Removing.

bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StylesheetInlinerTransformerFactory.java
jshelley-xx: It'd simplify the code a bit to move it to </head>, and I don't see that there's any difference in support in the campaignmonitor document.  Moving.

jshelley-xx:
I think these should address your comments.  I pulled those isCss and isJavascript tests to a new class in the same package, thinking they'd probably not be used from other packages.

justinedelson:
great work!

PR #867, opened 2016-12-12
Versioned clientlibs transformer factory

### LINKED ISSUE #866: Sling context path does not get stripped with versioned client libs
When a context path is set for AEM/Sling the versioned clientlib rewriter throws an error because the resource path is invalid

`com.adobe.granite.ui.clientlibs.impl.HtmlLibraryManagerImpl No library configured at /context-path/etc/clientlibs/mylib`

royteeuwen:
Fixed a bug for the versioned clientlibs transformer factory when a context path is set

Resolves #866

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/VersionedClientlibsTransformerFactory.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/VersionedClientlibsTransformerFactory.java
@@ -43,6 +43,8 @@ import org.apache.felix.scr.annotations.Reference;
 import org.apache.felix.scr.annotations.Service;
 import org.apache.sling.api.SlingConstants;
 import org.apache.sling.commons.osgi.PropertiesUtil;
+import org.apache.sling.rewriter.ProcessingComponentConfiguration;
+import org.apache.sling.rewriter.ProcessingContext;
 import org.apache.sling.rewriter.Transformer;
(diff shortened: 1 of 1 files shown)


bdhoine:
Thanks, we really needed this 👍

bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/VersionedClientlibsTransformerFactory.java
```diff
@@ -129,25 +131,33 @@ public Transformer createTransformer() {
         return new VersionableClientlibsTransformer();
     }
 
-    private Attributes versionClientLibs(final String elementName, final Attributes attrs) {
+    private Attributes versionClientLibs(final String elementName, final Attributes attrs, String contextPath) {
```
bdhoine: make String final

royteeuwen: done

wiiitek: There are two methods with `String contextPath` argument but it is made `final` in only one of them.
Is it ok?

royteeuwen:
Fixed :)

justinedelson:
Thanks!

PR #868, opened 2016-12-13
New Feature: HTLab - A micro DSL for Useful expression composition in HTL

adamcin:
HTLab
=====
A micro DSL for Useful expression composition in HTL
-----------------------

_adds the com.adobe.acs.commons.htlab namespace_

The `com.adobe.acs.commons.htlab.use.RSUse` class, or "Resource Selector Use", is the entry point for the HTLab micro-DSL within an HTL script. It makes, _ahem_, **creative** use of the HTL map key syntax to expose a simple dollar-delimited pipeline of single-argument function calls beginning with a property selector. The point of this is to provide a more concise means of extending HTL functionality, as compared to the out-of-the-box level of expressiveness which involves typing out a fully-qualified class name for every instance of a custom `Use` class for every combination of inputs and outputs.

To initialize the DSL for the request resource, create an instance of this class using the `data-sly-use` attribute:

`<sly data-sly-use.resource_="com.adobe.acs.commons.htlab.use.RSUse" />`

This allows one to access values from an associated underlying property map and then dollar-pipe them to `HTLabFunction` objects, which can either be registered as OSGi services, or implemented as `Use` classes. For example:

`${resource_['jcr:created $ jsonDate']}` evaluates to an ISO8601-formatted string representing the `Calendar` value of the `jcr:created` property.

(see `com.adobe.acs.commons.htlab.impl.func.JsonDateFunction`)

If a function name is not mapped to an active `HTLabFunction`, the input value is passed through and the function name logged, rather than throwing an exception.

PROPERTY NAME SELECTORS
-----------------
If the wrapped target object of an `RSUse` implements `Map` or adapts to a `ValueMap`, it is possible to select values from the `Map` representation using the keys by themselves. In fact, an attempt will always be made FIRST to check for the presence of a non-empty key in this underlying map representation before other selector rules are evaluated. This means that the keys in a wrapped `Map` will always take precedence over any peculiarities in the DSL syntax. 

THE SELF SELECTOR
-----------------
The target object of an `RSUse` instance can be selected instead of one of its properties by using the pipe operator by itself (the "Self" selector) or followed by a function.

`${resource_[' $ ']}` evaluates to the underlying Resource, whereas `${resource_[' $ pageLink']}` passes the `Resource` to the `com.adobe.acs.commons.htlab.impl.func.PageLinkFunction` service which generates a link to the resourcePage by mapping the containing-Page-path using the resolver and request context objects, and then appending ".html".

RELATIVE PATH SELECTORS
-------------------
The `RSUse` interprets key names beginning with `./` or `../` as potential relative paths and will attempt to resolve a resource located at that path relative to the wrapped target resource. For example:

`${resource_['./par/sling:resourceType $ toString']}` might evaluate to the value of sling:resourceType property of the root parsys in a page component.

In the same script, `${resource_['../jcr:primaryType $ toString']}` might evaluate to "cq:Page", which would be the jcr:primaryType of the parent resource.

If the wrapped target is not a resource, but does implement `Adaptable`, `RSUse` will first attempt `target.adaptTo(Resource.class)` to establish the base for relative path resolution.

**IMPORTANT**: Generally speaking, _Property Name Selectors_ by themselves will IMMEDIATELY evaluate to the best equivalent type matching the boxed JCR primitive value. In these expressions, the first function following the selector will likely be passed a String, Long, Calendar, Boolean, etc. On the other hand, JCR properties resolved by _Relative Path Selectors_ or by the _Self Selector_ MAY NOT be immediately unwrapped from their `Resource` representation. For this reason, a key feature of robust `HTLabFunction` implementations is that a function which accepts an input value of type `String` SHOULD ALSO accept an input value of type `Adaptable` that adapts to String.class. See `com.adobe.acs.commons.htlab.use.ToStringUseFn` for a simple example of this behavior, and see `com.adobe.acs.commons.htlab.impl.func.JsonDateFunction` for a more complex example.

INITIALIZER OPTIONS
-------------------
WRAP OBJECT ("`wrap=`")

To wrap an HTL variable, just specify `wrap=varName` in the `data-sly-use` options. For example,

`data-sly-use.currentPage_="${'com.adobe.acs.commons.htlab.use.RSUse' @ wrap=currentPage}"`

By convention, the RSUse variable should be named `var_` (i.e. "var" followed by underscore), where "var" is the name of the variable that is wrapped.

RESOURCE PATH ("`path=`")

It is also possible to wrap a resource resolved from a path. Simply specify `path='./path/to/resource'`. The path will be resolved relative to the request resource.

PIPE OPERATOR ("`pipe=`")

If the default dollar ("$") pipe operator conflicts with an existing property name in the wrapped map, you can specify a different operator using `pipe='[op]'`.

For example, initializing with `data-sly-use.resI18n_="${'com.adobe.acs.commons.htlab.use.RSUse' @ pipe='%', wrap=resI18n}"` allows for property access like so: `${resI18n_['Big Dishwasher Sale $ $ $ Marquee % truncateTo100']}`, where "Big Dishwasher Sale $ $ $ Marquee" is interpreted as the property name.

USE FUNCTIONS
-------------
HTLab functions can be bound in the `RSUse` initializer so long as the assigned function names do not collide with the argument names specified above (i.e., Use Functions cannot be bound as `wrap`, `path`, or `pipe`).

Specifically, the Use Function must be initialized before the `RSUse` instance, with a valid HTL use variable name:

`data-sly-use.myUseFunc="com.adobe.acs.commons.htlab.use.ToStringUseFn"`

Then, the use variable must be bound to the `RSUse` initializer as an option `name=value` pair:

`data-sly-use.currentPage_="${'com.adobe.acs.commons.htlab.use.RSUse' @ toString=myUseFunc}"`

The function can then be applied using the option name assigned in the `RSUse` initializer:
`${currentPage_['cq:lastModified $ toString']}` evaluates to "java.util.GregorianCalendar[time=..."

An Initializer-bound Use Function will always override an OSGi-registered function of the same name, regardless of the latter's `service.ranking` value.

HTLAB_USE SHORTCUT
------------------
The `com.adobe.acs.commons.htlab.impl.HTLabBindingsValuesProvider` adds the `HTLAB_USE` binding which is a mapping between simple class names and fully-qualified class names for the classes under the .htlab.use package. OSGi configuration binding is required to enable the shortcut in your HTL scripts.

For example,

1. `data-sly-use.resource_="com.adobe.acs.commons.htlab.use.RSUse"` becomes
 `data-sly-use.resource_="${HTLAB_USE.RSUse}"`

2. `data-sly-use.myUseFunc="${'com.adobe.acs.commons.htlab.use.ToStringUseFn' @ onNull='n/a'}"` becomes `data-sly-use.myUseFunc="${HTLAB_USE.ToStringUseFn @ onNull='n/a'}"`

3. `data-sly-use.fnAdaptToRepStatus="${'com.adobe.acs.commons.htlab.use.AdaptToUseFn' @ type='com.day.cq.replication.ReplicationStatus'}"` becomes `data-sly-use.myUseFunc="${HTLAB_USE.AdaptToUseFn @ type='com.day.cq.replication.ReplicationStatus'}"`

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/htlab/HTLabContext.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/htlab/HTLabContext.java
@@ -23,6 +23,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
 import javax.script.Bindings;
 import javax.script.SimpleBindings;
 
@@ -37,16 +38,16 @@ import org.slf4j.Logger;
 
 /**
  * Wraps a {@link Bindings} instance with accessor methods that are analogous to those of
- * {@link SlingBindings}
+ * {@link SlingBindings}.
  */
 public final class HTLabContext extends ValueMapDecorator {
+    public static final HTLabContext EMPTY = new HTLabContext(Collections.<String, Object>emptyMap());
 
-    private final Map<String, Object> bindings;
+    @Nonnull private final Map<String, Object> bindings;
 
-    private HTLabContext(Map<String, Object> base) {
+    private HTLabContext(@Nonnull Map<String, Object> base) {
         super(base);
-        this.bindings = Collections.unmodifiableMap(
-                new HashMap<String, Object>(base));
+        this.bindings = Collections.unmodifiableMap(new HashMap<String, Object>(base));
     }
 
     /**
@@ -63,8 +64,8 @@ public final class HTLabContext extends ValueMapDecorator {
      * @param bindings the input bindings provided to a {@link org.apache.sling.scripting.sightly.pojo.Use} class.
      * @return the new context object
      */
-    public static HTLabContext fromBindings(Bindings bindings) {
-        return new HTLabContext(bindings);
+    @Nonnull public static HTLabContext fromBindings(Bindings bindings) {
+        return new HTLabContext(bindings != null ? bindings : Collections.<String, Object>emptyMap());
     }
 
     /**
--- a/bundle/src/main/java/com/adobe/acs/commons/htlab/HTLabContext.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/htlab/HTLabContext.java
@@ -1,0 +1,124 @@
+/*
+ * #%L
+ * ACS AEM Commons Bundle
+ * %%
+ * Copyright (C) 2014 Adobe
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+package com.adobe.acs.commons.htlab;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.CheckForNull;
+import javax.script.Bindings;
+import javax.script.SimpleBindings;
+
+import org.apache.sling.api.SlingHttpServletRequest;
+import org.apache.sling.api.SlingHttpServletResponse;
+import org.apache.sling.api.resource.Resource;
+import org.apache.sling.api.resource.ResourceResolver;
+import org.apache.sling.api.scripting.SlingBindings;
+import org.apache.sling.api.scripting.SlingScriptHelper;
+import org.apache.sling.api.wrappers.ValueMapDecorator;
+import org.slf4j.Logger;
+
+/**
+ * Wraps a {@link Bindings} instance with accessor methods that are analogous to those of
+ * {@link SlingBindings}
+ */
+public final class HTLabContext extends ValueMapDecorator {
+
+    private final Map<String, Object> bindings;
+
+    private HTLabContext(Map<String, Object> base) {
+        super(base);
+        this.bindings = Collections.unmodifiableMap(
+                new HashMap<String, Object>(base));
+    }
+
+    /**
+     * Create a new {@link Bindings} instance from the underlying map.
+     * @return clone of original bindings
+     */
+    public Bindings cloneBindings() {
+        Map<String, Object> cloned = new HashMap<String, Object>(bindings);
+        return new SimpleBindings(cloned);
+    }
+
+    /**
+     * Wrap a {@link Bindings} instance with a new {@link HTLabContext}.
+     * @param bindings the input bindings provided to a {@link org.apache.sling.scripting.sightly.pojo.Use} class.
+     * @return the new context object
+     */
+    public static HTLabContext fromBindings(Bindings bindings) {
+        return new HTLabContext(bindings);
+    }
+
+    /**
+     * Get the HTTP Request object.
+     * @return the HTTP Request object
+     */
+    @CheckForNull public SlingHttpServletRequest getRequest() {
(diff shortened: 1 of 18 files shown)


gabrielwalt:
Sorry to ask, but what use-cases does this solve that cannot be solved otherwise to an acceptable level?

Also, following features will be available with 6.3, which might be some of the use-cases that this attempts to solve:
* Initialization of arbitrary resources:
  `<sly data-sly-use.myRes="some/path">${myRes.jcr:title}</sly>`
* Formatting of [dates and numbers](https://github.com/Adobe-Marketing-Cloud/htl-spec/blob/1.3/SPECIFICATION.md#122-format):
  `${'dd MMMM yyyy, EEEE' @ format=myDate, locale='de', timezone='UTC'}`
  `${'#,###.00' @ format=1000, locale='de'}`
* Passing of request attributes:
  `<sly data-sly-include="${'script.html' @ requestAttributes=attributesMap}"/>`
  `<sly data-sly-resource="${'some/path' @ requestAttributes=attributesMap}"/>`

adamcin:
@gabrielwalt 

These HTL features are great news and certainly comprise at least 80% of the immediate use cases inspiring the development of this feature.

The other aspect of HTL that this is trying to solve is the backdoor obfuscation of the component content model that I have seen on some projects, in which there is a tight encapsulation of Java Use classes around the component model, such that a Use class serves as both DAO and controller. I think this occurs because of the following reasons: 

1. Java Use class methods cannot take any arguments, which requires that any variables are passed as init options
2. Java class `data-sly-use` attributes are verbose and even simple classes with a single option usually force a developer to add at least one line break to see the next attribute or the end of the element tag, which adds noise when compared to the original HTML template.
3. As Use class complexity grows, the length of init options may exceed the width of a developer's editor window, which leads to the decision to either a) decompose its behavior into multiple classes, which contributes to point (2) above, or b) replace N options that accept scalar values with 1 option that accepts a Resource or ValueMap, or even 0 options when they decide to arbitrarily choose a Binding provided by the WCMUsePojo, like "properties" or "currentPage", instead.

When a critical mass of Java Use classes have been created that extend WCMUsePojo, developers on the project begin to see this as the _expedient_ pattern, and begin copying it everywhere. Then the copying leads to code duplication, which backend Java developers solve by extracting shared functionality into static methods, services, or bespoke model classes, which further abstracts the controller functionality away from the Sling pattern of "Have path -> get Resource -> get ValueMap / Adapter". This ultimately makes it more difficult to understand the relationship between content model and view that HTL seems designed to elevate and enforce. 

Granted, this is not a necessary or inevitable evolution on a project implementation, because with proper code reviews the maintainability of the Java-Use pattern can be preserved, but at the cost of convenience for developers.

With that in mind, the fundamental imperatives of this module are to:
1. Take ownership of the DAO role and decouple it from component-specific Use classes by handling the resource and property resolution in a standard library (`RSUse`) that should only require one verbose data-sly-use call in a script to expose any resource or property any number of times.
2. Enourage the use of declarative syntax in HTL files by elevating the "what" and "where" while omitting the "how" and encourage composition over inheritance in Java Use classes by establishing a concise syntax for assembling expressions that clearly define these relationships at point of render within a template.  
3. Provide a means of progressively incorporating context from site hierarchy and OSGi configuration into existing HTL components without having to fix or rewrite all HTL files at the same time, or having to integrate an osgi service dependency into each Use class directly.

I think the 6.3 enhancements to HTL will certainly help satisfy the first two imperatives I've listed above. For the third point I listed, however, I would still be looking for way to incorporate context of hierarchy or configuration, such that that developers can create and maintain the component while being agnostic to the presence or absence of that context depending on the environment.

The best example I have for this is link rewriting. When a link is rendered within an element and attribute that the default rewriter pipeline is configured for,  I have full orthogonal control over how that link is mapped in any particular environment without having to adjust any component code that produced it, assuming the component simply printed `${resource.path}.html`. However, if the link is rendered in a `data-search-results-href` attribute on a container `div` element, then I'm in a slightly less flexible position. I either have to configure the default rewriter pipeline to send startElement events for `div` or I have to modify the component HTL and/or Use code to integrate a call to a centralized link rewriter service.

PR #882, opened 2017-01-09
Contextual pathbrowser field

HitmanInWis:
Add a contextual pathbrowser field for working on multi-site platforms, so that authors can be limited to selecting paths within the current site.

Currently supports TouchUI only.


HitmanInWis:
This granite UI widget leverages the Page Root Provider (added by Shared Component Properties in the latest ACS Commons version) to determine the root page of the current site, and limits the author's selection to content under that path.

HitmanInWis:
Tested on AEM 6.1 and 6.2

PR #886, opened 2017-01-12
884 - Configure Parsys Placeholder error message fix

### LINKED ISSUE #844: Package Replication Status Updater uses same session for multiple events
We have configured the Package Replication Status Updater (JcrPackageReplicationStatusEventHandler) which works fine as long as there is not more than one package replicated at a time. We have a special case of mass replication where we use a groovy script to generate packages per content section and replicate them asynchronously (we are not using one package to not let them grow too big). 

Currently this causes many of the following WARNs in the log file:

```
12.11.2016 17:24:22.961 *WARN* [sling-threadpool-e7e460fb-28e9-4f10-9928-15d8de079324-(apache-sling-job-thread-pool)-99-<main queue>(acs-commons/replication/package)] org.apache.jackrabbit.oak.jcr.delegate.SessionDelegate Attempted to perform Setting property [/content/ ... path to page ..... /jcr:content/cq:lastReplicationAction] while thread sling-threadpool-e7e460fb-28e9-4f10-9928-15d8de079324-(apache-sling-job-thread-pool)-104-<main queue>(acs-commons/replication/package) was concurrently writing to this session. Blocked until the other thread finished using this session. Please review your code to avoid concurrent use of a session.
```

The root cause is that the class JcrPackageReplicationStatusEventHandler only uses one admin session for all package replication events it receives:
https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/acs-aem-commons-3.4.0/bundle/src/main/java/com/adobe/acs/commons/replication/status/impl/JcrPackageReplicationStatusEventHandler.java#L212

This should be fixed by creating an admin session for each package replication event.

davidjgonzalez:
Fixes #844 

Adds a check to see if the cellSearchPath is exists before attempting to see if it contains a parsys config. Its not unusual that all cellSearchPaths will not exist, and since this feature is "on by default" we should not worry people with console.log errors.

Ideally this feature is off by default, but I expect that ship has sailed until the next major release.

### DIFF
--- a/content/src/main/content/jcr_root/apps/acs-commons/widgets/source/js/classicui-configure-parsys-placeholder.js
+++ b/content/src/main/content/jcr_root/apps/acs-commons/widgets/source/js/classicui-configure-parsys-placeholder.js
@@ -31,6 +31,7 @@
         ACS_PARSYS_BG_COLOR = "acsParsysBackgroundColor",
         ACS_PARSYS_BORDER_COLOR = "acsParsysBorderColor";
 
+
     if( ( pathName !== "/cf" ) && ( pathName.indexOf("/content") !== 0)){
         return;
     }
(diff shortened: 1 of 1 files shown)


veeraswamikandregula:
Hi @davidjgonzalez , we just installed the latest ACS commons release 3.7.0 on our AEM 6.2 SP1 instance and we are still getting the parsys error. Please find the screenshot attached.

Can you please let us know if there is any easy way to switch this feature off.

untitled

davidjgonzalez:
@veeraswamikandregula can you click on the `widgets.js:1729` and provide the exact JS line that is causing this? Unfortunately we have this older JS that is "on by default" and im not sure there's a good way to disable it without breaking everyone that expects it to be on. In the mean time let me update the JS to be super-defensive in its checks.

davidjgonzalez:
Should be fixed in: https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/931

PR #897, opened 2017-01-30
#894 - Correct determination of STALE WF status

### LINKED ISSUE #894: Stale workflows are not removed
I selected to remove completed, aborted, suspended, and stale workflows.  After it successfully ran I see there are still stale workflows.

davidjgonzalez:
Prior to this fix the code looked at JUST the Workflow instance's status property, which works for everything except RUNNING and STALE (STALE workflows are RUNNING workflows with job IDs but not matching jobs).

This fix correctly determines the STALE state.

Fixed #894

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/workflow/bulk/removal/impl/WorkflowInstanceRemoverImpl.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/workflow/bulk/removal/impl/WorkflowInstanceRemoverImpl.java
@@ -21,6 +21,12 @@
 package com.adobe.acs.commons.workflow.bulk.removal.impl;
 
 import com.adobe.acs.commons.workflow.bulk.removal.*;
+import com.adobe.granite.workflow.WorkflowException;
+import com.adobe.granite.workflow.WorkflowSession;
+import com.adobe.granite.workflow.exec.WorkItem;
+import com.adobe.granite.workflow.exec.Workflow;
+import com.adobe.granite.workflow.exec.filter.WorkItemFilter;
+import com.day.cq.workflow.WorkflowService;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.felix.scr.annotations.*;
@@ -83,6 +89,8 @@ public final class WorkflowInstanceRemoverImpl implements WorkflowInstanceRemove
     @Reference
     private JobManager jobManager;
 
+    @Reference
+    private WorkflowService workflowService;
 
     /**
      * {@inheritDoc}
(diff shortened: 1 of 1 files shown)


davidjgonzalez:
SNAPSHOT build for 6.0/6.1 w fix:

[acs-aem-commons-content-2.9.1-SNAPSHOT-min.zip](https://github.com/Adobe-Consulting-Services/acs-aem-commons/files/739579/acs-aem-commons-content-2.9.1-SNAPSHOT-min.zip)

davidjgonzalez:
Worth noting that the avoidance of using WF APIs in this impl is intentional as if WF Model state (versions, etc.) enter unexpected state (cant resolve a WF instance not to a WF obj) then everything falls apart. Direct inspection of WF instance resource properties allows these "malformed" WF instances to be removed.

bundle/src/main/java/com/adobe/acs/commons/workflow/bulk/removal/impl/WorkflowInstanceRemoverImpl.java
```diff
@@ -331,6 +334,69 @@ public int removeWorkflowInstances(final ResourceResolver resourceResolver,
...
+            workflow = workflowSession.getWorkflow(workflowInstanceResource.getPath());
+            if (workflow == null) {
+                throw new WorkflowException(String.format("Workflow instance object is null for [ %s]", workflowInstanceResource.getPath()));
+            }
+        } catch (WorkflowException e) {
+            log.warn("Unable to locate Workflow Instance for [ {} ] So it cannot be RUNNING and must be STALE. ", workflowInstanceResource.getPath(), e);
+            return "STALE";
+        }
+
+        if (workflow != null) {
```
davidjgonzalez: @andrewmkhoury does this look right to you for checking is something has pending items... 

If atleast 1 workitem does NOT have a sling job id OR an end-time, that means its in the inbox waiting. 
If all work items have endtimes, then that means they have all been "completed" and thus there should be atleast 1 matching sling job to indicate its execution, else its STALE.

andrewmkhoury:
@davidjgonzalez It looks fine to me, but it needs testing.  Producing a stale workflow is a little tricky as it means the workflow references a sling job that doesn't exist.

davidjgonzalez:
@andrewmkhoury finally for around to testing this out and it works! thanks for the advice!

Attaching the 2.x builds w/ this fix if anyone is interested...

/cc @bowerjd

[acs-aem-commons-content-2.9.1-SNAPSHOT-min.zip](https://github.com/Adobe-Consulting-Services/acs-aem-commons/files/769514/acs-aem-commons-content-2.9.1-SNAPSHOT-min.zip)
[acs-aem-commons-content-2.9.1-SNAPSHOT.zip](https://github.com/Adobe-Consulting-Services/acs-aem-commons/files/769515/acs-aem-commons-content-2.9.1-SNAPSHOT.zip)

PR #905, opened 2017-02-12
#904 - Node Exists query builder predicate

### LINKED ISSUE #904: NodeExists QB predicate for determining if a relative node exists/notexists
Often, especially in the case of Bulk Workflow Management, it is helpful to identify Nodes where relative nodes do/do not exist.

For example, it is difficult to construct a query in any Query language/DSL that identifies dam:Asset's with an original rendition node, but are missing a web rendition node.

It would be nice to have a QueryBuilder predicate that could be used for these scenarios.  Note, this would be a filtering predicate, so all the usual performance considerations need to be taken into account.

```
nodeExists.or=true
nodeExists.exists=jcr:content/renditions/original
nodeExists.2_exists=jcr:content/renditions/cq5dam.thumbnail.48.48.png
nodeExists.1_notexists=jcr:content/renditions/cq5dam.web.1280.1280.png
nodeExists.2_notexists=jcr:content/renditions/cq5dam.web.600.400.png
```

davidjgonzalez:
Resolves #904 .. 

```
/**
 * This AEM QueryBuilder predicate checks if a JCR node exists, or doesn't exist, off the provided relative path.
 *
 * There are 3 configurations:
 *
 * nodeExists.or = true | false (defaults to false)
 * -- When nodeExists.or = false (the default), all .exists and .notexists conditions for this predicate are AND'd together to determine if the result node is included.
 *
 * nodeExists.#_exists = relative path (relative from the result node) to another node. This relative path mush exist for this expression to return true
 * -- Multiple exists conditions can be present and need to be prefixed via the usual `#_exists` syntax.
 *
 * nodeExists.#_notexists = relative path (relative from the result node) to another node. This relative path mush NOT exist for this expression to return true
 * -- Multiple exists conditions can be present and need to be prefixed via the usual `#_notexists` syntax.
 *
 * nodeExists.or=true
 * nodeExists.exists=jcr:content/renditions/original
 * nodeExists.2_exists=jcr:content/renditions/cq5dam.thumbnail.48.48.png
 * nodeExists.1_notexists=jcr:content/renditions/cq5dam.web.1280.1280.png
 * nodeExists.2_notexists=jcr:content/renditions/cq5dam.web.600.400.png
 */
```

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/search/impl/NodeExistsPredicateEvaluator.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/search/impl/NodeExistsPredicateEvaluator.java
@@ -70,7 +70,12 @@ public class NodeExistsPredicateEvaluator extends AbstractPredicateEvaluator imp
 
     @Override
     public final boolean canFilter(final Predicate predicate, final EvaluationContext context) {
-        return !predicate.getParameters().isEmpty();
+        if (predicate.getParameters().isEmpty()
+                || (predicate.getParameters().size() == 1 && predicate.getParameters().get(OR) != null)) {
+            return false;
+        } else {
+            return true;
+        }
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


davidjgonzalez:
@andrewmkhoury seem reasonable? Main use-case trying to hit is use w BWM to process Assets that are missing Renditions.

andrewmkhoury:
@davidjgonzalez yes, this will work great for the use-case.

PR #906, opened 2017-02-12
Feature/service user manager

### LINKED ISSUE #892: Service User Manager
Basic idea:
* Simplify service user creation and service user configuration
* Ensure that service user exists
* Ensure that service user permissions are correct (edited)

ideally the code should be small and simple, nothing too complex; a set of rules similar to the "Day CQ ACL Setup Service"

-----

WDYT about 2 Config Factories... the first that will add/update .. and the 2nd that removes; not sure if the Remove being separate is overkill... we could have an "operation" property and consolidate?

com.adobe.acs.commons.util.impl.EnsureServiceUser-abc.xml
```
principalName="service-user"
ace="[type=allow;permissions=jcr:read,rep:write;path=/content/foo;rep:glob=/jcr:content/*]"
```

Im thinking this would

1. Create the service user if does not exist
2. Query for all rep:grant/rep:deny with principal name = principalName
3. Remove any ACE's that dont match the config
4. Add any ACE's that are defined in the config but don't exist

... Remove would remove all ACLs and the service user.

/cc @andrewmkhoury @joerghoh

davidjgonzalez:
Fixes #892 

More i think about it, the more I like OSGi based definitions vs in-code Annotation based. Open to debate; or potentially could have both approaches (if the latter is contrib'd later)


davidjgonzalez:
@andrewmkhoury @justinedelson @bstopp  - WDYT?

davidjgonzalez:
thoughts on how to achieve the blacklist? Manually curate it?

andrewmkhoury:
I guess manually curate it.  You could use the json dump feature of JCR compare to get all the user ids from /home/users/system.  Then grep for a list of rep:principalName and generate the code for a HashSet.

justinedelson:
I thought there was a unique index on rep:principalName, so you can't overwrite an existing name.

andrewmkhoury:
They are unique, you can't have two users with the same rep:principalName.
However, I'm more concerned about the ACLs.  To avoid wiping out the ACLs
we would need to check against a list of the OOTB users.

justinedelson:
oh. got it. nevermind then.

davidjgonzalez:
@andrewmkhoury added a protected system users principal name list (based on the list you PM'd me).

davidjgonzalez:
Any last thoughts before merging?

andrewmkhoury:
Nope, seems good to me.

PR #908, opened 2017-02-13
Feature/audit log search

klcodanr:
This is the AEM 6.0 compatible version of the Audit Log Search described in this pull request: https://github.com/Adobe-Consulting-Services/acs-aem-tools/pull/173


content/src/main/content/jcr_root/_oak_index/.content.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
davidjgonzalez: Im thinking this should be a lucene index.. they tend to be more robust than property indexes. 
Does this HAVE to be part of the commit hook?

Also - dont recall the oak version it was introduced, but lucene can now be marked as "nrt" (or near real time) .. which triggers from, though doesnt block, the commit hook.

content/src/main/content/jcr_root/_oak_index/.content.xml
```diff
@@ -0,0 +1,8 @@
...
+<jcr:root xmlns:oak="http://jackrabbit.apache.org/oak/ns/1.0" xmlns:jcr="http://www.jcp.org/jcr/1.0" xmlns:nt="http://www.jcp.org/jcr/nt/1.0" jcr:primaryType="nt:unstructured">
+        <cqAuditEvent
```
davidjgonzalez: @justinedelson what are your thoughts on including an Index as part of the package?

content/src/main/content/jcr_root/_oak_index/.content.xml
klcodanr: @davidjgonzalez -- I don't see why it would have to be part of the commit hook. As long as it updated in a reasonable time.

content/src/main/content/jcr_root/_oak_index/.content.xml
klcodanr: @davidjgonzalez @justinedelson I wonder if it'd make sense to do something like ensure oak index where the user could have the option to install the oak:index if it's not already installed.

davidjgonzalez: I guess my concern w auto-deploying an index is a pretty intrusive operation. If someone doesn't want to use this feature they still have the cost/risk of managing this index...

I wonder if we make a check that displays on the tool page warning if the index isn't created, and in docs we provide an ensure oak index definition that can be deployed to create it?

klcodanr: @davidjgonzalez -- Sounds like a plan. I'm thinking that it might make sense to add a warning or something as well if no index is found because it's pretty worthless without the index. I'll give it a crack and update here when complete

klcodanr:
@davidjgonzalez so I updated it to display a warning if it doesn't find an oak:index for the cq:AuditEvent type and gives the user a button to create the index. I also changed the index to be lucene instead of proerty. After running a couple queries it seems like the lucene index performs about 10x better than property index when filtering by path (given the way the query gets constructed in this tool).

davidjgonzalez:
Replacing a #929

PR #914, opened 2017-02-28
#391: added support for pages without jcr:content node.

nc-andreashaller:


bundle/src/main/java/com/adobe/acs/commons/errorpagehandler/impl/ErrorPageHandlerImpl.java
```diff
@@ -260,31 +260,19 @@ public String findErrorPage(SlingHttpServletRequest request, Resource errorResou
         final ResourceResolver resourceResolver = errorResource.getResourceResolver();
         final String errorResourcePath = errorResource.getPath();
 
-        // Get error page name to look for based on the error code/name
-        String errorsPath = null;
-
-        // Try to find the closest real parent for the requested resource
-        final Resource parent = findFirstRealParentOrSelf(request, errorResource);
```
davidjgonzalez: Isnt this required in the event a `HTTP GET /content/pages/page/jcr:content/root/non-existing-resource.html` is requested? The error resources isn't guarenteed to be a `[cq:Page]/jcr:content` or `[dam:Asset]/jcr:content` - or am i missing something?

PR #919, opened 2017-03-05
searches for the frequency and priority attribute in the current page…

yelouardi:
# Site Map V2

## Issues 
In current site map the values priority and freq they are recoverable from current page if these values are nulls the page is not validate  at  google 

## Solution
If the values priority and freq are nulls then get inheritance value 

## Condition
Check that attribute in property dialog of AEM/ConfigMgr ( Sitemap Servlet ) include.inherit  is true to activate inheritance value

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
@@ -33,9 +33,6 @@ import javax.xml.stream.XMLOutputFactory;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamWriter;
 
-import com.day.cq.commons.jcr.JcrConstants;
-import com.day.cq.dam.api.Asset;
-import com.day.cq.dam.api.DamConstants;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.time.FastDateFormat;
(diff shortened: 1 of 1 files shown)


bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -274,18 +285,33 @@ private void writeAssets(final XMLStreamWriter stream, final Resource assetFolde
...
+                writeElement(stream, elementName, value);
+                break;
+            }
+        }
+    }
+
+    private void writeFirstPropertyValue(final XMLStreamWriter stream, final String elementName,
+            final String[] propertyNames, final InheritanceValueMap properties) throws XMLStreamException {
         for (String prop : propertyNames) {
             String value = properties.get(prop, String.class);
```
davidjgonzalez: Not a big deal, but this line can just be `value = properties.getInherited(prop, String.class)` as `getInherited(..)` tries to get the the local property first, and only traverses up if it cannot be found locally.

https://docs.adobe.com/content/docs/en/aem/6-2/develop/ref/javadoc/com/day/cq/commons/inherit/InheritanceValueMap.html#getInherited(java.lang.String,%20java.lang.Class)

yelouardi: Ok thank for your help

PR #920, opened 2017-03-06
New Feature: Automatic Package Replicator

klcodanr:
Here's something based on what I've done for a couple projects. Instead of either doing a mass-activation or having jobs running in the publishers, this will build and replicate a package of content.

The jobs can be triggered based on:

 - Sling Events
 - Cron Triggers
 - Workflows


klcodanr:
Hey @davidjgonzalez - changes incorporated. Hope it looks good!

PR #924, opened 2017-03-08
Content Migration Framework

joerghoh:
This is the first iteration of a content migration framework. 

In projects you often encounter the situation, that over time the structure of individual components and pages change or need to be modified. Typically this is done by writing a custom servlet, which does that job.
But this is unflexible and often lacks some features like 
* partial migration 
* dry run
* reasonable logging

This Content Migration Framework tries to provide these features, and requires you to implement these 2 aspects (by implementing the ContentMigrationStep interface)
* identifying the resources which need to be changed
* transform a resource from the existing structure to the new structure

The execution of one of the registered ContentMigrationSteps can be triggered using the ContentMIgrationProcessor service. A sample use of the ContentMigrationProcessor can be seen in the ContentMigrationServlet.

At the moment I am still not really confident with the node. I wonder if the naming is good and I am open to suggestions for better names (for both classes and the methods). Also I would like to get some feedback if that's the way how you rewrite your content structures if you need to perform changes.

So please don't merge it right now, at the moment I am just looking for feedback :-) 
Thanks!


toniedzwiedz:
Hi,

I'm just wondering how the proposed solution compares to using the [AEM Groovy Console by Olson Digital](https://github.com/OlsonDigital/aem-groovy-console) to run content migration scripts. We tend to use this approach a lot at Cognifide and TBH, I've never written a Servlet to update instances of a component across the content tree.

The console gives one access to the repository, provides bindings for OSGi Services and common adaptables and runs arbitrary Groovy code. It's easy enough to obtain a JCR session and persist or roll back arbitrary changes.

Have you had a chance to try it?

davidjgonzalez:
@toniedzwiedz is that like aem fiddle or the sling script console (which I believe is groovy based as well) ... and it allows you to run adhoc scripts in aem?

joerghoh:
@toniedzwiedz This tool is not intended to run adhoc scripts, but is rather designed for planned content changes triggered without the need to access Osgi Console; this means, that it doesn't require admin privileges to execute the content migration. 

So technically both solutions can do the same, but in some details (which are important to me) this framework provides better integration into authoring/development/enterprise processes than a generic script console.

toniedzwiedz:
@davidjgonzalez I suppose it's similar, although I'm not very familiar with AEM fiddle and I didn't realise it could also run Groovy.

@joerghoh I see. Thanks.

bundle/src/main/java/com/adobe/acs/commons/cmf/MigrationOptions.java
```diff
@@ -0,0 +1,27 @@
+package com.adobe.acs.commons.cmf;
+
+import javax.jcr.Session;
+
+public class MigrationOptions {
```
davidjgonzalez: Waiting to see how this is used; but the following seem like potential options:

* Commit batch size (if this is desirable; have to deal w multi-resource rollback then)
* Throttle (be it N ms every X resources or engage ACS's ThrottleTaskService)

bundle/src/main/java/com/adobe/acs/commons/cmf/MigrationOptions.java
```diff
@@ -0,0 +1,27 @@
...
+	boolean dryRun;
+	
+	
+	public MigrationOptions() {
+		session = null;
+		dryRun = false;
+	}
+	
+	
+	public MigrationOptions setSession (Session s) {
```
davidjgonzalez: Why not ResourceResolver?

bundle/src/main/java/com/adobe/acs/commons/cmf/impl/ContentMigrationServlet.java
```diff
@@ -0,0 +1,88 @@
...
+		
+		out.print("<ul>");
+		while (iter.hasNext()) {
+			String s = iter.next();
+			out.println("<li>" + s + "</li>");
+		}
+		out.print("</ul>");
+		
+		String usedMigrationStep = request.getParameter(MIGRATION_STEP);
+		String path = request.getParameter(PATH);
```
davidjgonzalez: Should the params be collected in the Steps? 
I assume custom steps will require a number of other params.

bundle/src/main/java/com/adobe/acs/commons/cmf/impl/ContentMigrationServlet.java
```diff
@@ -0,0 +1,88 @@
...
+		
+		out.print(String.format("<p>Starting %s on path %s</p>", new Object[]{usedMigrationStep,path}));
+		
+		try {
+			IdentifiedResources id = cmp.identifyAffectedResources(usedMigrationStep, path, request.getResourceResolver());
```
davidjgonzalez: Maybe a v2 thing, but i've really come to like CSV files for data at scale. Rendering to a web page tends to be 
1) Transient (hit refresh... cya data!) 2) Difficult to review (scrolling, hard to find info)

bundle/src/main/java/com/adobe/acs/commons/cmf/TraversingContentMigrationStep.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.apache.sling.api.resource.Resource;
+import org.apache.sling.api.resource.ResourceResolver;
+
+
+/**
+ * This class implements the Visitor pattern for the identifyResources() method.
+ *
+ */
+public abstract class TraversingContentMigrationStep implements ContentMigrationStep {
```
davidjgonzalez: Definitely like this the abstracts that take out the boiler-plate work

bundle/src/main/java/com/adobe/acs/commons/cmf/TraversingContentMigrationStep.java
```diff
@@ -0,0 +1,60 @@
...
+		List<Resource> result = new LinkedList<Resource>();
+		return visit (result, rootResource);
+	}
+	
+	/**
+	 * Visits the resource r and its sub-resources. 
+	 * @param r
+	 * @return
+	 */
+	private List<Resource> visit (List<Resource> list, Resource r) {
```
davidjgonzalez: I wonder if this should have some filter options... back in 6.0 days i had to rely every heavily on traversing (search was still evolving) and I became close friends with `accepts(..)` in [1] .. i let me decrease my traversals by vast amounts since i typically didnt want to traverse below the [cq:Page] or [dam:Asset] level - once I was at those levels, the underlying structure was well-known and i didnt have traverse everything below.


[1] https://sling.apache.org/apidocs/sling7/org/apache/sling/api/resource/AbstractResourceVisitor.html

davidjgonzalez:
@joerghoh like the idea -- do you have a use case that prompted you to create this you could share? Typically when i think "migration", i think content coming into AEM, and im not seeing facilities for that (maybe thats a custom step?)

davidjgonzalez:
@joerghoh figured its worth mentioning...  is there overlap w Sling Pipes? TBH i havent used them so my understanding is purely academic..

https://sling.apache.org/documentation/bundles/sling-pipes.html

joerghoh:
Thanks @davidjgonzalez for the review. I still have a few things on the agenda for this, but I will definitly look into your fedback.

Regarding the name "migration". I came across the need for such a framework when we pushed back changes for a component because it would have required us to change (or "migrate") every occurrence of this component within the content. You could also name it a "content modification framework" if you like :-)

davidjgonzalez:
@joerghoh gotcha; makes sense -- i still thing there might be something to abstracting out the source side, such that it could get the source data to transform from anywhere -- though that might expand the scope too much.

I think there is a general need to take source data, transform w biz rules, and put someplace else in AEM; not sure if this is too custom to even attempt to build a framework that could handle the variability here.

If the scope is kept to transforming existing resources, the term "migration" is a bit leading; might be better as "modification" or "transformation" as you note in your comment.

joerghoh:
I don't want to make it too broad, the usecase I am looking right now is the evolution of components and templates, which requires changes to already created content (pages).
Let's use  "content modification framework" because then I don't have to change the package name :-)

davidjgonzalez:
@joergh you know there is the template and resource type updates in acs tools that does a pretty nice job of this -- do your transforms require more inspection of the nodes to transform? Could that just be augmented?

cwoeltge:
Hi Jörg,

just saw your PR but can't access code right now. 
Couple of questions/comments from a conceptional pov. 
Did you have a look into Sling Pipes? At least parts of it are intended to solve this as well.

For custom migration code Fiddle is part of ACS and runs with the permissions of the current user (pretty similar to the Groovy Console). Pretty common for any ad-hoc or planned migrations, flexible and straight forward. 

For execution you can run this on package install using our vault-upgrade-hook - to allow execution on install on every instance your code update is deployed to. 

So, how would you position your tool related to the above. 

Regards,
© 


Sent from my iPhone

joerghoh:
@cwoeltge regarding your questions:

* I considered Sling pipes, but the documentation could be improved a lot. I truly believe that it's much more flexible, but also harder to understand.
* And ACS Fiddle is ... well, a fiddle. It's something for adhoc processes, but in my point of view not well for processes which need to be executed at scale with no manual intervention. And even then you still need to write the code which does the changes you would like to perform.

I see this tool as something in between. At a time you might need to change the resourcetypes of pages. You need to change some component nodes at scale. Things you normally write a one-off script for. Which often lacks the obvious things like logging, dry-run etc.

davidjgonzalez:
Replaced w #962

PR #926, opened 2017-03-11
httpcache enhancement: allow caching of non 200 responses

benpeter:
an obvious use case for caching on AEM (as opposed to Dispatcher or above) would be non-200 Responses, which Dispatcher will not cache, so that e.g. redirects can make up a high part of the requests that actually hit the publishers.

this PR brings cachability of responses that are not 200 OK to the engine and the in-memory cache storage. some steps remain to be done on the project side (in particular reconfiguring the engine to not include `CacheOnlyResponse200` and creating rules that make sure that the repsonse code is duely considered). I tried to keep the PR to the minimum that is required to make non-200-caching possible at all.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/httpcache/store/mem/impl/MemTempSinkImpl.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/httpcache/store/mem/impl/MemTempSinkImpl.java
@@ -53,7 +53,7 @@ public class MemTempSinkImpl implements TempSink {
         if (null != sink) {
             return new ByteArrayInputStream(sink);
         } else {
-            throw new HttpCacheDataStreamException("Nothing available in sink.");
+            return new ByteArrayInputStream(new byte[0]);
         }
     }
 
--- a/bundle/src/main/java/com/adobe/acs/commons/httpcache/engine/CacheContent.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/httpcache/engine/CacheContent.java
@@ -62,11 +62,7 @@ public class CacheContent {
     public CacheContent(String charEncoding, String contentType, Map<String, List<String>> headers, InputStream
             dataInputStream) {
 
-        this.status = HttpServletResponse.SC_OK;
-        this.charEncoding = charEncoding;
-        this.contentType = contentType;
-        this.headers = headers;
-        this.dataInputStream = dataInputStream;
+        this(HttpServletResponse.SC_OK, charEncoding, contentType, headers, dataInputStream);
     }
 
     /**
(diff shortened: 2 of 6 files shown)


bundle/src/main/java/com/adobe/acs/commons/httpcache/store/mem/impl/MemTempSinkImpl.java
```diff
@@ -53,7 +53,7 @@ public InputStream createInputStream() throws HttpCacheDataStreamException {
         if (null != sink) {
             return new ByteArrayInputStream(sink);
         } else {
-            throw new HttpCacheDataStreamException("Nothing available in sink.");
```
davidjgonzalez: QQ - what is the implication of not throwing the exception here?

davidjgonzalez:
@benpeter never woulda thought of this!

So to use this, the CacheOnly200 rule would be removed from the configuration, right? 

image

Looks good to me but + @Sivaramvt in case he has thoughts/concerns.

benpeter:
@davidjgonzalez yes, plus adding a custom rule to narrow down on the statūs to actually cache (I cannot think of a use case to cache irrespective of response status, but who knows).
this would need to be done in project space, the aim here was to remove the restriction to 200s.

bundle/src/main/java/com/adobe/acs/commons/httpcache/store/mem/impl/MemTempSinkImpl.java
benpeter: if existing client code makes the assumption that a cachable reponse with `HTTP/1.1 200` but an empty body results in an exception that bubbles up and effectively turns the reponse into a `HTTP/1.1 500`, that assumption would break.

It seems like a particularly unlikely assumption to me, though.

The exception currently occurs for an empty body, as only the call to `getOutputStream()` or `getWriter()` of the cache's respose wrapper creates the underlying `ByteArrayOutputStream` of the in-memory store.

The `TempSink` API contract makes no statement on when to expect the declared `HttpCacheDataStreamException`.

Sivaramvt: `HttpCacheDataStreamException` represents an IOException that may happen while reading/writing body content streams . Originally, the problem we wanted to solve was primary to deal with 200. So, supporting other status codes might require this.

Reg. the above said assumption - Any exception thrown in the http cache would let the request pass through for processing.

Sivaramvt:
Glad to see this PR on http cache. It's been over a year and I barely remember the code.
Good we keep this change crisp retaining the `CacheOnly200` rule config in engine avoiding the risk of caching intermittent 500s caused by server.

davidjgonzalez:
I expect well want a new non-default rule that allows users to specify what StatusCodes to cache.

PR #929, opened 2017-03-14
Feature/audit log search

davidjgonzalez:
Updated @klcodanr with updated traversal checks so we can cut a release.


davidjgonzalez:
@klcodanr chatted w @justinedelson about the the Index creation and the check;

1) For now well just pull out the index into the feature docs so ppl can create and manage it along w all their usual indexes (indexes are just so intrusive)
2) I added a new method to QueryHelper that checks for traversals by checking the Plan

PR #930, opened 2017-03-15
Sling filter to avoid upstream caching with incorrect md5

### LINKED ISSUE #911: Versioned ClientLibs not working with multiple publishers/dispatchers
The root problem is that any md5sum can be used to request a versioned clientlib. This causes an old version of the clientlib to be cached with the md5sum of a new version. I'll explain.

We currently have 4 publishers (P), each fronted with a dispatcher (D). When we deploy, we take the first P+D out of the load balancer, deploy, put them back in the load balancer, and do all other 3 P+D the same way. 

Because of this a newer version of the client libs is deployed on a part of our infrastructure.

Someone requests a webpage from the freshly deployed instance and gets the md5sum of the new clientlibs. The request for the clientlib comes in, but the load balancer requests it from another P+D. This one responds with the old clientlib, but caches it with the new md5sum.

As I can see, we only have 2 options:
1/ we take all P offline and serve stale content while we are deploying, or
2/ some servlet filter is written which checks if the md5sum matches with the cache data (as seen in JMX). If not, it responds with 404, which causes the clientlibs not to be cached with the wrong md5sum.

For us option 2/ would be the most viable option.

Also curious that nobody else has ever noticed this before?

PS: Using sticky load balancer is not an option.

wimsymons:
This fixes #911

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/VersionedClientlibsTransformerFactory.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/VersionedClientlibsTransformerFactory.java
@@ -19,19 +19,14 @@
  */
 package com.adobe.acs.commons.rewriter.impl;
 
-import java.io.IOException;
-import java.nio.charset.Charset;
-import java.util.Map;
-import java.util.concurrent.Callable;
-import java.util.concurrent.ExecutionException;
-
-import javax.management.DynamicMBean;
-import javax.management.NotCompliantMBeanException;
-import javax.management.openmbean.CompositeType;
-import javax.management.openmbean.OpenDataException;
-import javax.management.openmbean.OpenType;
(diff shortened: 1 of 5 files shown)


justinedelson:
@wimsymons can you take a look at #930? I rewrote your pull request to happen inside the TransformerFactory, enabled with a boolean property on the factory component. This both provides the easy option to enable/disable this behavior (it is disabled by default for backwards compatibility reasons) and also avoids leaking the cache as a service.

image

The behavior is the same *except* for the case where there is no entry in the cache and there is no client library (the test named `doFilter_notFoundInCache_NoClientLib`). In your patch, this had the filter return a 404, but I think it should actually be a pass through. Otherwise, we could potentially block the use of non-client library JS/CSS.

wimsymons:
Indeed. See my comment on #945.

justinedelson:
obviated by #945

Issue #864, opened 2016-12-06
Optional inclusion of ACS Commons Clientlibs


davidjgonzalez:
Not sure if the ship has sailed on this, but it might be nice to start optionally including clientlibs that effect (and potentially break) major UIs.

For @HitmanInWis's we gave each clientlibs a discrete name and for someone to use it, they must create a proxy Clientlib that includes our feature clientlib into `cq.widgets` or `cq.authoring.editor` based on their needs.

We could even manage our own "acs-commons.widgets.all" if someone always wanted everything? I dont like that we have a slew of stuff already added to the OOTB categories, but dont htink we can do much about that.

WDYT? good, bad, overkill?

https://adobe-consulting-services.github.io/acs-aem-commons/features/shared-component-properties.html > How to Use > 2. Adding the Client Libraries for Authoring

//cc @justinedelson, @schoudry, @bstopp

schoudry:
@davidjgonzalez i think it's a good idea; though additional configuration needed and some bugs might be missed (as the logic executes only after adding respective proxy clientlib) upside is - it doesn't break UI on hotfix or FP installs (which ACS commons almost never tests with) or upgrades

bstopp:
I like the idea. Auto-including the ACS widgets is akin to auto-loading services. We require configs to be defined to enable any feature. Same should be for custom widgets; yes, maybe it requires an extra development step but creating aggregate clientlibs is almost standard ops now.

Issue #698, opened 2016-04-21
Configure Parsys Placeholder


rcouto-cit:
Hi,

On the version 2.3.2 there is a feature called touchui-configure-parsys-placeholder that tries to request a configuration from a path like that: /etc/designs/geometrixx/jcr:content/contentpage/par

But as I don't have this path configured in my repository, neither I want to use this feature, I believe that this feature should starts turned off and only when I want to use it I should have a way to turn it on.

When this feature starts and tries to access a node that doesn't exist the author can see this error on the console:
granite.min.js:64 GET http://localhost:4502/etc/designs/default/jcr:content/templateWeb/par.2.json?_=1461268087386 404 (Not Found)

Which is the better way to resolve this issue?
![screenshot from 2016-04-21 16 59 49](https://cloud.githubusercontent.com/assets/12274195/14722578/88e63ade-07e2-11e6-92c9-95fcba52e836.png)

stevenwoodson:
I've also just come across this same issue and hoping for a resolution.

justinedelson:
Sorry, perhaps I'm being dense, but why is this an issue? A 404 is a valid response and this doesn't prevent other JS on the page from working.

shams8090:
Hi, i'm facing the same issue. Any fix or work around.
Currently i have installed ACS commons version 2.8.

rao4:
Facing the same issue on 3.0.2 any resolutions found for this issue?

adpv-gm:
Has this issue been resolved? I'm using the latest version of AEM Commons (3.7.0) and still seeing the 404 console errors when hovering over parsys.

davidjgonzalez:
@adpv-gm can you show the line of code that is initiating this XHR? Like @justinedelson requesting something that doesnt existing isnt an error; it can simply be checking to see if it exists, and if it doesn't it will react accordingly. The 5xx class of responses are the ones that you ought to be worried about.

shubhakirani:
We had the same issue. All the jsons with .2.json are being blocked. We commented this function below and issue disappear. Please recommend what this function is for and if we have any alternate solution?

`/*        $.ajax( { url: path + ".2.json", async: false } ).done(function(data){
            if(_.isEmpty(data) || !data[ACS_COMPONENTS_LIMIT]){
                return;
            }

            currentLimit = data[ACS_COMPONENTS_LIMIT];

            var limit = parseInt(data[ACS_COMPONENTS_LIMIT]);

[...]

PR #944, opened 2017-03-28
allow for dynamic inclusion of client libraries in classic UI. Fixes …

justinedelson:
…#864 and #698


justinedelson:
This allows for some of our Classic UI features to be broken out into separate client libraries (currently JS only) and then dynamically included. To start, I broke out the limit parsys and parsys placeholders as those seem to be the most problematic for some.

davidjgonzalez:
To bubble up a key point in this impl, the classicUI widgets will remain "on by default", and a user must actively "exclude all" to disable them.

justinedelson:
Build failure appears unrelated (and transient).

PR #945, opened 2017-03-28
sling filter to avoid upstream caching of incorrect md5 values

### LINKED ISSUE #911: Versioned ClientLibs not working with multiple publishers/dispatchers

justinedelson:
based on patch from @wimsymons, but modified to be tightly coupled to the transformer

fixes #911

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/VersionedClientlibsTransformerFactory.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/VersionedClientlibsTransformerFactory.java
@@ -21,16 +21,27 @@ package com.adobe.acs.commons.rewriter.impl;
 
 import java.io.IOException;
 import java.nio.charset.Charset;
+import java.util.Dictionary;
+import java.util.Hashtable;
 import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import javax.management.DynamicMBean;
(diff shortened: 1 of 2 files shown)


justinedelson:
To enable this filter, there is now a boolean property (`enforce.md5`) in the transformer configuration: 

image

wimsymons:
This is great. Just want to point out a bug we found recently. When using ?debugClientLibs=true, the MD5 should not be checked, otherwise it will return 404 for all .js and .css URI's.

justinedelson:
@wimsymons thanks for pointing that out. I checked that and it seems to be OK now.

PR #948, opened 2017-04-03
Fix MD5 enforcement for paths with dots in them.

justinedelson:
As mentioned in #911, the original implementation of MD5 enforcement failed to take into accounts paths with dots in them. This PR fixes that. It does come with the consequence that *if* you enable the Enforce MD5 option, the paths now get rewritten as

    /path/to/clientlib.ACSHASH<md5>.js

This isn't ideal, but I also don't see another way to differentiate between 

  /path/to/clientlib.with.dots.in.it.md5.js
  /path/to/clientlib.md5.js


henrykuijpers:
... An md5 is always a-f0-9{32}, right? So why can't you use that fact to differentiate?

Also, what use case would there be for having nodes with a dot in the name? Or did you mean selectors?

justinedelson:
Probably not since that could also be part of the node name. And no, I didn't mean selectors.

PR #969, opened 2017-04-17
Feature/957

davidjgonzalez:
Fixes issue with error page handler errorPaths property lookup.

In the previous update to this feature, InheritanceValueMaps were introduced that look at the incorrect path ([cq:Page] vs [cq:Page]/jcr:contnet) 

@nc-andreashaller  can you take a quick look?

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/errorpagehandler/impl/ErrorPageHandlerImpl.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/errorpagehandler/impl/ErrorPageHandlerImpl.java
@@ -306,12 +306,18 @@ public final class ErrorPageHandlerImpl implements ErrorPageHandlerService {
      */
 	private String findErrorsPath(SlingHttpServletRequest request, Resource errorResource) {
 		final String errorResourcePath = errorResource.getPath();
-		final Resource page = findFirstRealParentOrSelf(request, errorResource);
+		Resource real = findFirstRealParentOrSelf(request, errorResource);
 
         String errorsPath = null;
-        if(page != null) {
-            log.trace("Found page is [ {} ]", page.getPath());
-	        final InheritanceValueMap pageProperties = new HierarchyNodeInheritanceValueMap(page);
+        if (real != null) {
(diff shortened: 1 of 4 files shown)


bundle/src/main/java/com/adobe/acs/commons/errorpagehandler/impl/ErrorPageHandlerImpl.java
```diff
@@ -306,12 +306,18 @@ public String findErrorPage(SlingHttpServletRequest request, Resource errorResou
...
-		final Resource page = findFirstRealParentOrSelf(request, errorResource);
+		Resource real = findFirstRealParentOrSelf(request, errorResource);
 
         String errorsPath = null;
-        if(page != null) {
-            log.trace("Found page is [ {} ]", page.getPath());
-	        final InheritanceValueMap pageProperties = new HierarchyNodeInheritanceValueMap(page);
+        if (real != null) {
+            log.trace("Found real resource at [ {} ]", real.getPath());
+            if (!JcrConstants.JCR_CONTENT.equals(real.getName())) {
```
nc-andreashaller: If I see it right this will always be true because `findFirstRealParentOrSelf` will always return the plain page path. Shouldn't this logic be added to the `findFirstRealParentOrSelf` method instead?

bundle/src/main/java/com/adobe/acs/commons/workflow/bulk/execution/impl/runners/AEMTransientWorkflowRunnerImpl.java
```diff
@@ -212,8 +213,13 @@ public void run() {
                     workspace.commit();
                 }
             } catch (Exception e) {
-                log.error("Error processing periodic execution: {}", e);
+                log.error("Error processing periodic execution for job [ {} ] for workspace [ {} ]", new String[]{ jobName, workspace.getPath() }, e);
```
nc-andreashaller: The String array is not needed here `log.error("Error processing periodic execution for job [ {} ] for workspace [ {} ]", jobName, workspace.getPath(), e);` should just work fine.

Issue #974, opened 2017-04-21
Bulk Workflow Manager Transient WF Runner uses admin resourceResolver


davidjgonzalez:
Discovered by Ian R.

Only applies to 3.x.x

davidjgonzalez:
Fixed via #976

PR #976, opened 2017-04-21
#974 - Bulk Workflow Manager - Transient WF Runner to use service res…

davidjgonzalez:
…ource resolver


ireasor:
Looks good - standard service user stuff.  Any reason it caused the Travis build to fail or is that normal?

davidjgonzalez:
@ireasor nope; looks like issue is with ...

```
[ERROR] Failures: 
[ERROR]   MonthlyExpiresHeaderFilterTest.testDoActivateSuccess:221 Values should be different. Actual: 30
[ERROR]   WeeklyExpiresHeaderFilterTest.testDoActivateSuccess:176 Values should be different. Actual: 30
```

Kicking it off again to see if its a glitch or something more sinister

PR #977, opened 2017-04-21
#975 - remove use of the deprecated Felix whiteboard properties

### LINKED ISSUE #975: AEM Environment Indicator does not work on AEM 6.3 TouchUI
image

justinedelson:
This change impacts the Environment filter and the dispatcher Expires/Cache-Control filters.

Testing looks good on 6.2 and 6.3 for the Environment filter and the Cache-Control filter, but can't get the Expires header to work. I see the filter being registered and called, but the expires header always shows up at `Expires: Thu, 01 Jan 1970 00:00:00 GMT`

Hoping @bstopp has a better test setup.

### DIFF
--- a/bundle-twitter/pom.xml
+++ b/bundle-twitter/pom.xml
@@ -85,11 +85,11 @@
         </dependency>
         <dependency>
             <groupId>org.osgi</groupId>
-            <artifactId>org.osgi.compendium</artifactId>
+            <artifactId>osgi.cmpn</artifactId>
         </dependency>
         <dependency>
             <groupId>org.osgi</groupId>
-            <artifactId>org.osgi.core</artifactId>
+            <artifactId>osgi.core</artifactId>
         </dependency>
         <dependency>
             <groupId>org.apache.felix</groupId>
--- a/bundle/pom.xml
+++ b/bundle/pom.xml
@@ -115,11 +115,11 @@
         </dependency>
         <dependency>
             <groupId>org.osgi</groupId>
-            <artifactId>org.osgi.compendium</artifactId>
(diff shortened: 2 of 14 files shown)


davidjgonzalez:
LGTM .. will wait for @bstopp to check out the expires; odd that this change would effect that behavior though.

bstopp:
This looks good to me; i think you may have been forgetting to add the header:
`Server-Agent: Communique-Dispatcher` to your requests. I tested the daily expires via PostMan with:
```
filter.pattern=/content.*
expries.time=23:00
```

Header value: `Expires →Fri, 21 Apr 2017 23:00:00 GMT`

justinedelson:
@bstopp the header is definitely there. Do I need to test this on publish maybe?

bstopp:
I tested mine on Author. I saw a Jan 1 1970 date one time, but wasn't able to recreate after I fixed my request. Let me see if i can recreate, i'll spend some time testing possible causes.

justinedelson:
hmm. yeah, it works in Postman. I was testing in curl.

bstopp:
Ok yes that is weird, i see it not working in cURL and it should; i'll use that and retest.

bstopp:
I tested it through the dispatcher; it looks to be working correctly:

```
-rw-r--r-- 1 apache apache 61854 Apr 21 19:08 en.html
-rw-r--r-- 1 apache apache     0 Apr 23  2017 en.html.ttl
```
The header being wrong via cURL must be a curl issue or some other filter coming into play?

justinedelson:
thanks for the review @bstopp

PR #980, opened 2017-04-21
Created a new workflow process to set replication status on the provi…

ireasor:
…ded node.

Refactored a method from ReplicatedByWorkflowProcess to reuse, resulting in the creation of ReplicationResourceLocator.

During asset migrations, after using a tool like VLT-RCP, this process can be used in conjuction with BWM to mark the assets that were VLTed over as replicated.


bundle/src/main/java/com/adobe/acs/commons/replication/status/impl/ReplicationResourceLocatorImpl.java
```diff
@@ -0,0 +1,52 @@
...
+ * OSGi Service for finding the resource responsible for tracking replication status on a given node.
+ */
+@Component
+@Service
+public class ReplicationResourceLocatorImpl implements ReplicationResourceLocator {
+
+	private static final Logger log = LoggerFactory.getLogger(ReplicationResourceLocatorImpl.class);
+	
+	@Override
+	public Resource getReplicationStatusResource(String path, ResourceResolver resourceResolver) {
```
davidjgonzalez: WDYT about moving this method into `ReplicationStatusManager`?  It seems the only time you'd want to find that resource is you're going to manage the replication status? Feels overkill to have a service that just does this light lookup.

[1] https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/replication/status/impl/ReplicationStatusManagerImpl.java

bundle/src/main/java/com/adobe/acs/commons/replication/status/impl/ReplicationResourceLocatorImpl.java
```diff
@@ -0,0 +1,52 @@
...
+		final Page page = resourceResolver.adaptTo(PageManager.class).getContainingPage(path);
+		final Asset asset = DamUtil.resolveToAsset(resourceResolver.getResource(path));
+
+		Resource resource;
+
+		if (page != null) {
+		    // Page
+		    resource = page.getContentResource();
+		    log.trace("Candidate Page for setting replicateBy is [ {} ]", resource.getPath());
```
davidjgonzalez: "Candidate XXX for setting replicateBy is..." -> "XXX's resource that tracks replication status is..."

bundle/src/main/java/com/adobe/acs/commons/replication/status/impl/ReplicationResourceLocatorImpl.java
```diff
@@ -0,0 +1,52 @@
...
+		} else if (asset != null) {
+		    // DAM Asset
+		    final Resource assetResource = resourceResolver.getResource(asset.getPath());
+		    resource = assetResource.getChild(JcrConstants.JCR_CONTENT);
+		    log.trace("Candidate Asset for setting replicateBy is [ {} ]", resource.getPath());
+		} else {
+		    // Some other resource
+		    resource = resourceResolver.getResource(path);
```
davidjgonzalez: I dont think its necessary, but is it helpful to look around for a jcr:content node? Trying to think of outliers..

Tags -> trackers rep status on the cq:Tag

wondering if something like the following would be useful? ..or maybe too assuming and could cause problems? Just throwing it out there.

```
if (resource.getChild("jcr:content") != null ) {
    resource = resource.getChild("jcr:content");
} else {
    Resource tmp = resource;
     while(!"jcr:content".equals(tmp.getParent().getName()) {
           tmp = tmp.getParent();
     }
    if ("jcr:content".equals(tmp.getName()) {
         resource = tmp;
    }
}
```

bundle/src/main/java/com/adobe/acs/commons/replication/status/impl/ReplicationResourceLocatorImpl.java
ireasor: Sure thing.  I had actually considered that before.  I was trying to avoid a SRP violation, but it's possible that setting the status and determining where to set the status are similar enough concepts to not be a concern.  I'll move it to the existing service.

bundle/src/main/java/com/adobe/acs/commons/replication/status/impl/ReplicationResourceLocatorImpl.java
ireasor: Since this was a refactor of existing code, I tried to change as little as possible.  The log message makes less sense in a generalized method, though.  I'll go ahead and update it.

bundle/src/main/java/com/adobe/acs/commons/replication/status/impl/ReplicationResourceLocatorImpl.java
ireasor: Any concerns about the performance impact of a change like this?  There is also a potential issue with this approach in that you may want to activate a resource that doesn't have a jcr:content node but has a parent or child that does have a jcr:content node.  This would result in not replicating your resource but replicating the parent or child instead.  Seems risky.

ireasor:
One of my tests is non-deterministic.  Will submit an update shortly...

ireasor:
Comments addressed.  Let me know if there's anything else.

davidjgonzalez:
@ireasor LGTM! will merge after travis does his thing

davidjgonzalez:
@ireasor when you have a minute, can you read through [1] and make sure its OK?

https://adobe-consulting-services.github.io/acs-aem-commons/features/workflow-processes/set-replication-status/index.html

PR #981, opened 2017-04-25
Upstream fixes for Fast Action Manager

badvision:
- Updated for Java 8 support
- Fixes to busy detection logic in ThrottledTaskRunnerImpl which is now a lot less CPU intensive
- Refactored retry logic in DeferredActions

This is an upstream cherry-pick from some more good stuff to come out hopefully soon.  Stay tuned!

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/fam/mbean/package-info.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/fam/mbean/package-info.java
@@ -17,5 +17,5 @@
  * limitations under the License.
  * #L%
  */
-@aQute.bnd.annotation.Version("1.1.0")
+@aQute.bnd.annotation.Version("3.0.0")
 package com.adobe.acs.commons.fam.mbean;
--- a/bundle/src/main/java/com/adobe/acs/commons/fam/mbean/package-info.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/fam/mbean/package-info.java
@@ -17,5 +17,5 @@
  * limitations under the License.
  * #L%
  */
-@aQute.bnd.annotation.Version("3.0.0")
+@aQute.bnd.annotation.Version("1.1.0")
 package com.adobe.acs.commons.fam.mbean;
--- a/bundle/src/main/java/com/adobe/acs/commons/fam/ActionManager.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/fam/ActionManager.java
@@ -62,11 +62,23 @@ public interface ActionManager {
     void withResolver(Consumer<ResourceResolver> action) throws Exception;
     
     /**
-     * After scheduling actions withQueryResults or deferredWithResolver, schedule
-     * a cleanup task to close all remaining resource resolvers.
-     * NOTE: This is mandatory!
+     * Register a handler to be fired when the work has completed with no errors.
+     * @param successTask 
      */
-    void addCleanupTask();
+    void onSuccess(Consumer<ResourceResolver> successTask);
+
+    /**
+     * Register a handler to be fired when the work has completed and there was at least one error.
+     * @param failureTask 
+     */
+    void onFailure(BiConsumer<List<Failure>, ResourceResolver> failureTask);
+    
+    /**
+     * Register a handler to be fired when the work is completed, successfully or not.  
+     * Note: These handlers are called after the success/fail handlers.
+     * @param finishHandler 
+     */
+    void onFinish(Runnable finishHandler);
 
     /**
      * Have all actions completed?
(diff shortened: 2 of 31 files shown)


badvision:
Oh and commit cb8ecb1 is awesome -- the cleanup is now automatic AND the action manager now supports ending events.  This allows better features such as chaining one action manager to start when another one has finished.  This is the core of MCP.  :)

justinedelson:
This is a very non-backwards compatible change. I'm assuming this is warranted, but I'm concerned about breaking existing consumers of this API.

Are there any alternatives? Like duplicating the whole package space to `fam2` and starting the versioning over at `1.0.0`?

badvision:
There are not many folks using it yet.  This is necessary evolution.  The
auto-cleanup alone is worth the price of admission.

justinedelson:
I'm not seeing how the auto-cleanup requires any non-backwards compatible API changes.

IIUC, that is isolated to removing a single method from the `ActionManager` interface. That removal isn't necessary; you can just make it a no-op implementation.

badvision:
Yes, that is a point well-taken.  Good call-out there.  I think the actual automatic cleanup made sense to be renamed as well as it makes it clearer that the purpose of that method is for automatic cleanup.  I have updated the interface with the necessary javadoc so that most IDEs will warn the developer it is deprecated and explains it is automatic.

justinedelson:
Well, there are still some other issues -- the switch of instance to static methods is problematic for baseline as are the changes from abstract classes to interfaces of the functions package. The good news is that the bump of the mbean package seems entirely unnecessary, so that change to the `package-info.java` file can just be rolled back.

I think the functions package is going to be very challenging to evolve in a way that doesn't break backwards compatibility. One idea would be to introduce a new set of methods to `ActionManager` and `DeferredActions` which return the standard JDK classes and deprecate the ones which use the `functions` package. We can do a release with those changes and then remove the deprecated methods in a future release. This is no doubt more painful for you as the developer but would at least provide a glidepath for consumers of the API.

badvision:
No can do on the functions packages -- the refactored ones have to stay because the Java functions implementations don't allow exceptions; whereas mine do.  It's a subtle difference but it means code doesn't have to be littered with try/catch logic everywhere.  As for the functions package changing, there's no compile breakage for old vs new code.   New code uses lambdas so the function classes are never directly referenced or even imported for that matter.   It is more an issue that Jetty doesn't handle Java 8 in JSPs so there's no way that inline functions can be declared anymore.  MCP will resolve that issue, because there won't be a need to use FAM from code fiddles anymore anyway.

badvision:
And in order for functional java 8 expressions to work, the classes have to be functional interfaces with default methods such that only one method has to be declared in classes that implement that interface.  Abstract classes are not compatible with lambdas; the functional stuff was only declared as abstract to be temporarily backwards compatible with Java 7.

justinedelson:
> As for the functions package changing, there's no compile breakage for old vs new code. 

How can this be true since previously consumers would extend a base class and now they have to implement an interface? It would be fine to create *new* interfaces, but changing a class to an interface is a non-backwards compatible change which require refactoring by consumers and that's a non-starter from my perspective.

If the change was compatible, you wouldn't need to change the export version's major version. The fact that baseline is telling you to do so is indicative of a non-backwards compatible change.

The goal should be to only increment the minor version.

bundle/src/main/java/com/adobe/acs/commons/fam/mbean/package-info.java
```diff
@@ -17,5 +17,5 @@
  * limitations under the License.
  * #L%
  */
-@aQute.bnd.annotation.Version("1.1.0")
+@aQute.bnd.annotation.Version("3.0.0")
```
justinedelson: this change does not appear to be necessary

badvision:
@justinedelson So aside from the breaking changes on the Functions (which I doubt were getting used much outside of FAM anyway), I was able to make the other changes less major.  Let me know if this is more palatable.

justinedelson:
Please see https://github.com/badvision/acs-aem-commons/pull/10 which restores backwards compatibility. I'm not thrilled with the interface/class names, but perhaps you have better suggestions and can refactor from there.

badvision:
Yeah I'm not terribly thrilled but I think you have a good approach there with the adapter patterns.  Good call!  I think I'll use a separate package name or something instead or something but I'll have to ponder it.  Functional interfaces are funky beasts.  Is the adapter pattern needed though?  Couldn't the abstract classes just be changed to implement the functional interfaces?  I'll play with it a little.

justinedelson:
bq. Is the adapter pattern needed though? Couldn't the abstract classes just be changed to implement the functional interfaces?

I tried that but it created some issues because we actually do want to change the API of `ActionManager` to use the interfaces. So this puts us on a migration path.

justinedelson:
One other thing to think about is whether or not what is now called `IDeferredActions` should be make into an interface with a separate implementation. I don't recall offhand why that pattern wasn't followed in this case, but it obviously a lot easier to do with a new interface.

justinedelson:
I made a few minor cosmetic changes in 8882569 before merging.

thanks for sticking with this.

badvision:
No need to thank me!  Thanks to you for keeping us on the straight and narrow and future-proofing the heck out of this bad boy!  🥇 for putting up with my shenanigans!

PR #987, opened 2017-05-02
#920 - Package re-org to "tool" agnostic specific

davidjgonzalez:
@klcodanr moved some of the files around to be "tool" agnostic, can you do a quick sanity check of these moves?

AFIACT this does not impact the WebUI component at all.


klcodanr:
Looks good to me @davidjgonzalez 👍

PR #990, opened 2017-05-05
Add resource visitors which support lambda expressions

badvision:
Also add configuration hints for Netbeans to use Java 1.8


badvision:
There is now also a batch-retry utility class which lets you group actions together in batches such that they are all run/committed at a time and if any fail then they are retried a number of times together.  This allows some speedup in batches of work involving a lot of commits.

justinedelson:
Please submit separate features as separate pull requests. Otherwise, we'll lose track of what is in each release.

badvision:
Fair 'nuff

PR #991, opened 2017-05-05
Update Shared Component Properties to show option only for applicable components

HitmanInWis:
One thing that has bothered me since the launch of Shared Component Properties was that the "Shared Properties" and "Global Properties" controls should show up on ALL components, when likely only a fraction of your components use those dialogs.  This is misleading to authors and can even break the authoring UI requiring a page refresh if an author clicks the button for a non-applicable component.

This PR resolves the issue (for both the Touch UI and the Classic UI).  Now a component will only show the "Shared Properties" button if it has a shared props dialog, and will only show the "Global Properties" button if it has a global props dialog.


bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesPageInfoProvider.java
```diff
@@ -23,42 +23,88 @@
...
+
+    /**
+     * Add a "sharedComponentProperties" section to pageInfo so that JS libs in the authoring interface
+     * can determine whether or not to enable shared/global properties for a component on a site.
+     */
     @Override
     public void updatePageInfo(SlingHttpServletRequest request, JSONObject info, Resource resource)
             throws JSONException {
+        if (scheduledSharedComponentsMapUpdate != null && Calendar.getInstance().after(scheduledSharedComponentsMapUpdate)) {
+            scheduledSharedComponentsMapUpdate = null;
```
HitmanInWis: Understood that this is not 100% threadsafe and that 2 exactly concurrent requests could trigger this to run twice, but the operation runs in well less than a second so not worried about it.

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesPageInfoProvider.java
```diff
@@ -23,42 +23,88 @@
...
+    private ResourceResolverFactory resourceResolverFactory;
+
+    @Reference
+    private SlingRepository repository;
+
+    private Session respositorySession;
+
+    private ObservationManager observationManager;
+
+    private Map<String, List<Boolean>> componentsWithSharedProperties;
```
davidjgonzalez: ConcurrentHashMap?

HitmanInWis: Not sure it's necessary b/c we never modify the internals of the map - the variable is only ever set via an atomic assignment (=).

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesPageInfoProvider.java
```diff
@@ -23,42 +23,88 @@
...
- * are enabled.  Note that this provider requires Page Root Provider to
- * be configured.
- *
+ * are enabled, and for which component types.  Note that this provider
+ * requires Page Root Provider to be configured.
+ * <p>
  * https://docs.adobe.com/docs/en/cq/5-6-1/developing/pageinfo.html#Creating a Page Information Provider
  */
 @org.apache.felix.scr.annotations.Component
 @Service
```
justinedelson: This should have the PageInfoProvider to be explicit about the service interface now that the class implements two interfaces.

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesPageInfoProvider.java
```diff
@@ -87,4 +134,151 @@ public void updatePageInfo(SlingHttpServletRequest request, JSONObject info, Res
...
+                    boolean hasGlobalDialogForTouch = componentHasTouchDialog(component, "dialogglobal");
+                    boolean hasSharedDialogForClassic = componentHasClassicDialog(component, "dialog_shared");
+                    boolean hasGlobalDialogForClassic = componentHasClassicDialog(component, "dialog_global");
+                    if (hasSharedDialogForTouch || hasGlobalDialogForTouch || hasSharedDialogForClassic || hasGlobalDialogForClassic) {
+                        localComponentsWithSharedProperties.put(component.getResourceType(),
+                                Arrays.asList(hasSharedDialogForTouch, hasGlobalDialogForTouch, hasSharedDialogForClassic, hasGlobalDialogForClassic));
+                    }
+                }
+            }
+            componentsWithSharedProperties = localComponentsWithSharedProperties;
```
justinedelson: this should be

    componentsWithSharedProperties = Collections.unmodifiableMap(localComponentsWithSharedProperties);

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesPageInfoProvider.java
```diff
@@ -23,42 +23,88 @@
...
+    @Reference
+    private SlingRepository repository;
+
+    private Session respositorySession;
+
+    private ObservationManager observationManager;
+
+    private Map<String, List<Boolean>> componentsWithSharedProperties;
+
+    private Calendar scheduledSharedComponentsMapUpdate = null;
```
justinedelson: Could we use a timestamp (a `long`) instead of a Calendar object.?

HitmanInWis: doh...one more refactor incoming...can just use `System.currentTimeInMillis()` now...

HitmanInWis:
Thanks @justinedelson @davidjgonzalez for the quick review and merge.

Should I make an update in https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io to call out the update for the next version of ACS Commons?  This feels worth mentioning since it could solve a blocker for people using the feature.  Not sure if you guys just kinda set those comments in https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io or if you want me to.

davidjgonzalez:
That'd be great! Thanks!

HitmanInWis:
Sounds good, submitted PR.

PR #994, opened 2017-05-05
This utility allows processing actions in batches, such that multiple changes can be retried if any of them fail and break the commit.

badvision:
This allows greater overall concurrency and prevents race conditions from causing unrecoverable data integrity issues.


davidjgonzalez:
@badvision any idea why the visitors are showing as new files; i merged 993 already which contained them. I think this should be ok, but i dont understand why GH is saying these are new files/lines when they (AFIACT) already exist via the merged #993

badvision:
Yeah I was taking a refactor and split it into different parallel branches.  Let me see if I can clean up the PR.

badvision:
I merged it back with master but it still shows the other commits.  No conflicts though.  I think it's just a side-effect of how I split up the feature branches.

bundle/src/main/java/com/adobe/acs/commons/fam/impl/ActionManagerImpl.java
```diff
@@ -144,7 +144,7 @@ public void withResolver(CheckedConsumer<ResourceResolver> action) throws Except
         resolver.setCurrentItem(currentPath.get());
         try {
             action.accept(resolver.getResolver());
-        } catch (Exception ex) {
+        } catch (Throwable ex) {
```
justinedelson: What's the reason for switching to `Throwable` here? In general, catching `Throwable` is a bad idea since that also catches `Error` whose javadoc essentially says not to catch it.

badvision: If you look at the purpose of the catch logic it is supposed to report ANY cause of failure.  Most Error subclass stuff will likely halt the JVM before it hits this but NOT looking for problems causes other issues down the line when the action manager finishes, hasn't caught fatal issues, and acts like everything is hunky dory.  That is very bad; because without knowing there was a fatal problem somewhere, other code down the line will continue blissfully ignorant of the issue.  So by catching and logging that any Throwable is encountered, this makes for a more seamless and reliable transition from simple one-thread iterative logic into the multi-threaded rapid-fire logic that FAM allows.

justinedelson: sorry, I meant to comment on the *next* change, where the exception isn't thrown. Here I think you don't need the catch block at all and this should just be rewritten to 

   try {
       action.accept(resolver.getResolver());
   } finally {
       ... finally block...
   }

bundle/src/main/java/com/adobe/acs/commons/fam/impl/ActionManagerImpl.java
```diff
@@ -278,9 +278,10 @@ private void deferredWithResolver(
                 if (!closesResolver) {
                     logCompletetion();
                 }
-            } catch (Exception ex) {
+            } catch (Throwable t) {
```
justinedelson: So here is where I think Throwable is wrong since you aren't rethrowing the exception. I think `catch (Exception|RuntimeException) ex) ` would be a better choice.

badvision: I see -- well point taken.  I still want to know if any failure occurs but the RuntimeException bit might be a better idea.  I was having trouble with it catching checked exceptions correctly.

justinedelson:
Merged with a few changes:
* Changed logging to use SLF4J for consistency
* When a fatal exception happens in a retry, the exception gets thrown regardless of how many more retry attempts are available.

I also added some notes to the shiny new CHANGELOG.md file. Please review and edit as you see fit.

PR #998, opened 2017-05-08
update to latest sling mock versions

justinedelson:


badvision:
It looks like some of the tests are failing with the same root cause:
`java.lang.RuntimeException: Unable to invoke method 'bindAdapterFactory' for class org.apache.sling.testing.mock.sling.MockAdapterManagerImpl
Caused by: java.lang.NoSuchMethodError: org.apache.sling.adapter.internal.AdapterFactoryDescriptor.<init>`

justinedelson:
@badvision do you have a specific example? Does that happen on the CLI or in your IDE?

badvision:
No I was just reading the log from the failed build via Travis CI -- I haven't tried to build with this change yet.

justinedelson:
weird. I can't reproduce locally. I'll try moving the dependencies around

justinedelson:
looks like it is working now after removing the old testing dependency

PR #1000, opened 2017-05-08
Feature/984

davidjgonzalez:
Added to Synthetic Workflow forms (Single and Multi-threaded)

image

Allows for specifying your own custom or picking from 2 pre-defined options.


ireasor:
Seems like there is a lot of code duplication between components/utilities/bulk-workflow-manager/includes/fast-action-manager/form.jsp and components/utilities/bulk-workflow-manager/includes/synthetic-workflow/form.jsp.  Any way that we could extract part of this markup into another file that we include in both?

Issue #1005, opened 2017-05-18
Assets Folder Properties support


davidjgonzalez:
It would be nice to be able to provide client-specific simple properties on Assets Folders via the OOT Folder Properties Web UI.

![image](https://cloud.githubusercontent.com/assets/1451868/26218396/d3bf4cf0-3bd8-11e7-81e1-e38c5a9dee14.png)

at this time it is difficult to do some due to how data is written and read from the sling:XXXFolder and sling:XXXFolder/jcr:content nodes.

davidjgonzalez:
This implementation would admittedly be on the less elegant side and somewhat limited due to the OOTB plumbing that drives Folder Properties. There would  be 2 concerns:

1) Writing new properties -> Requires a Sling Filter to write whitelists property values (initially only single-val String support; could be extended to support typeHints); Unfortunately Sling POST Processors are not a viable solution due to the implementation of the OOTB FolderShareHandler PostOperation.

2) Reading new properties -> The reading of folder properties is ~ hardcoded into folderproperties.jsp OOTB; This would require a supporting servlet that exposes raw node properties to the granite.ui.forms.values ValueMap to allow new form fields to accessed saved data.

This allows for something like the following to be added to Assets Folder Properties.

![image](https://cloud.githubusercontent.com/assets/1451868/26218602/a299df4a-3bd9-11e7-8b99-16454e61ff12.png)

davidjgonzalez:
Was stewing on this some more and an alternate approach is to create a 1) Felix filter that hides the operation param 2) let sling post servlet handle normal property setting, 3) post processor that invokes the FolderShareHandler PostOperation

PR #1006, opened 2017-05-18
#1005 - Assets Folder Properties Support

davidjgonzalez:
Initial pass; Will write tests if there isnt a better way.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/dam/impl/AssetsFolderPropertiesSupport.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/dam/impl/AssetsFolderPropertiesSupport.java
@@ -1,0 +1,206 @@
+package com.adobe.acs.commons.dam.impl;
+
+import com.day.text.Text;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.felix.scr.annotations.*;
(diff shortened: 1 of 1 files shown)


bundle/src/main/java/com/adobe/acs/commons/dam/impl/AssetsFolderPropertiesSupport.java
```diff
@@ -0,0 +1,206 @@
...
+        if (log.isTraceEnabled()) {
+            log.trace("Accepted by AssetsFolderPropertiesSupport [ {} ]", request.getResource().getPath());
+        }
+
+        for (final String property : properties) {
+            log.debug("Looking for request parameter named [ {} ] in [ {} ]", property, StringUtils.join(request.getRequestParameterMap().keySet()));
+            boolean deleteHint = false;
+            boolean propertyParamExists = false;
+
+            if (request.getRequestParameterMap().keySet().contains(property)) {
```
bstopp: I'd recommend grabbing keySet into a var, the number of nested calls and the fact you have to do it twice.

davidjgonzalez:
@bstopp yup - the OOTB PostOperation performs its own commit. I suppose i could move it to run before and drafting (with a safety check) on its commit...

PR #1014, opened 2017-05-24
Feature/1005 - Assets Folder Properties Support

### LINKED ISSUE #1005: Assets Folder Properties support

davidjgonzalez:
Resolves #1005 

Re-factored to leverage the default Sling POST Servlet, and invokes the OOTB FolderShareHandler via a PostProcessor.

* Config policy required
* Granite UI names are off the sling:Folder and NOT the [sling:*Folder]/jcr:content

Example property for setting Brand Portal mcConfig property

image

[acs-aem-commons-assets-folder-properties-support-1.0.0.zip](https://github.com/Adobe-Consulting-Services/acs-aem-commons/files/1025865/acs-aem-commons-assets-folder-properties-support-1.0.0.zip)

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/dam/impl/AssetsFolderPropertiesSupport.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/dam/impl/AssetsFolderPropertiesSupport.java
@@ -87,6 +87,12 @@ public class AssetsFolderPropertiesSupport extends SlingSafeMethodsServlet imple
     private static final String GRANITE_UI_FORM_VALUES = "granite.ui.form.values";
 
     /**
+     * The is a reference to the OOTB AEM PostOperation that handles updates for Folder Properties; This is used below in process(..) to ensure that all OOTB behaviors are executed.
+     */
+    @Reference(target="&(sling.post.operation=dam.share.folder)(sling.servlet.methods=POST)")
+    private PostOperation folderShareHandler;
+
+    /**
      * This method is responsible for post processing POSTs to the FolderShareHandler PostOperation (:operation = dam.share.folder).
      * This method will store a whitelisted set of request parameters to their relative location off of the [sling:*Folder] node.
      *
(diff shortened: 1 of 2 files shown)


bundle/src/main/java/com/adobe/acs/commons/dam/impl/AssetsFolderPropertiesSupport.java
```diff
@@ -0,0 +1,247 @@
...
+     * @throws IOException
+     * @throws ServletException
+     */
+    public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse, FilterChain chain) throws IOException, ServletException {
+        final SlingHttpServletRequest request = (SlingHttpServletRequest) servletRequest;
+        final SlingHttpServletResponse response = (SlingHttpServletResponse) servletResponse;
+
+        if (!accepts(request)) {
+            chain.doFilter(request, response);
+            return;
```
ireasor: Maybe a personal style difference, but I usually feel like an if/else is more readable than a return statement inside an if clause followed by code outside the if clause.

bundle/src/main/java/com/adobe/acs/commons/dam/impl/AssetsFolderPropertiesSupport.java
```diff
@@ -0,0 +1,247 @@
...
+    public void init(FilterConfig filterConfig) throws ServletException {
+        // Do Nothing
+    }
+
+    public void destroy() {
+        // Do Nothing
+    }
+
+    @Reference(target="&(sling.post.operation=dam.share.folder)(sling.servlet.methods=POST)")
+    private PostOperation folderShareHandler;
```
ireasor: I usually declare all my references at the top.  Is that not a standard pattern?

davidjgonzalez:
@ireasor thanks for the review! will adjust

bundle/src/main/java/com/adobe/acs/commons/dam/impl/AssetsFolderPropertiesSupport.java
davidjgonzalez: ill leave this one; oddly(?) i find nested if/else statements to be less comprehendible (especially as depth increases) -- i think because having a flat-linear decision spaces reduces the amount of state i have to remember the further down the function i go.

davidjgonzalez:
@ireasor also removed the null check on the service reference object in process(..) since the reference requires it to be satisfied for the @Component to become active.

PR #1024, opened 2017-06-13
Vanity path look-up support on mapped requests

SachinMali:
This feature attempts to solve the problem of when using 
* JCR RR for outgoing mapping (ex. removing /content/sample/en from URLs)
* httpd rewriting for incoming url mapping (ex. pre-fixing /content/sample/en onto URLs before passing to AEM)

The above model, while hugely popular and simpler than /etc/map, breaks the ability to use vanity paths in AEM, since vanity paths "naturally" mangled by the httpd-level path prefixing (ex. if the sling vanity path is `/r/b`, httpd will mangle it to `/content/sample/en/r/b` or `/content/sample/en/r/b.html` before sending to AEM), and naturally AEM is unable to resolve a mangled vanity path to a real vanity path.

This feature hooks into the sling 404 handling to un-mangle requests and then forward them back through the Sling stack for vanity resolution. This handles both internal vanity redirects and external.

@joerghoh, @bstopp, @justinedelson thoughts?


bundle/src/main/java/com/adobe/acs/commons/wcm/vanity/impl/VanityURLServiceImpl.java
```diff
@@ -0,0 +1,113 @@
...
+	 * @param request SlingHttpServletRequest object used for performing query/lookup
+	 * @return return true if the vanityPath is a registered sling:vanityPath under /content
+	 */
+	protected boolean isVanityPath(String vanityPath, SlingHttpServletRequest request) throws RepositoryException {
+		final long start = System.currentTimeMillis();
+
+		final Map<String, String> params = new HashMap<String, String>();
+
+		// Limit to /content as this could show up in /jcr:system version nodes, etc.
+		params.put("path", "/content");
```
bstopp: To prevent multi-tenancy issues; thoughts on providing the removed prefix path to this method to form the base path of the search? This will force root context to be the site on which the original request was submitted (e.g. `/content/sample/en`), this should eliminate  cross-contamination.

bstopp:
With this, thoughts on updating the `VanityDuplicateCheckServlet` to take into consideration the site's context root, as an optional config (eg. depth?) so that vanities can be reused, in different contexts?

bundle/src/main/java/com/adobe/acs/commons/wcm/vanity/impl/VanityURLServiceImpl.java
```diff
@@ -0,0 +1,113 @@
...
+
+		// Manually strip off any selectors or extensions from the URL
+		String candidateVanity = StringUtils.substringBefore(requestPathInfo.getResourcePath(), ".");
+		// Map the incoming URL to remove any prefix
+		candidateVanity = request.getResourceResolver().map(candidateVanity);
+
+		log.debug("Candidate vanity URL to check and dispatch: [ {} }", candidateVanity);
+
+		// Check if...
+		// 1) the candidateVanity and the requestURI are the same; If they are it means the request has already
```
davidjgonzalez: it wouldnt? If the Request attribute is set to true, that means this has somehow gone through this loop already. Maybe the req attribute name could use some love? "acs-aem-commons__vanity-check-loop-detection" ?

bundle/src/main/java/com/adobe/acs/commons/wcm/vanity/impl/VanityURLServiceImpl.java
davidjgonzalez: so we thought about this... it would get you 1/2 way there; you still have the problem of vanities across multiple tenants being problematic. since we didnt a great way to FULLY handle multi-tenant vanities, we left out that consideration for now. If its massive boon to have partial support, we could add it back in. 

Also, it begs the question; are we ever using vanities for things that ARENT being re-written (i dont think we are..)

davidjgonzalez:
@bstopp I think someone else updating VanityDuplicateCheckServlet sounds like a great idea 👍

bundle/src/main/java/com/adobe/acs/commons/wcm/vanity/impl/VanityURLServiceImpl.java
bstopp: The assumption is that all of these requests are going through:
1. An outbound only Resource Resolver mapping.
1. An Apache ProxyPass rule to prepend the root to resolve (1) back to a repository path. 

So every request should be fully qualified NonExistingResources. if we document that this feature should only be used in conjunction with the vanity duplication checker; it would complete the feature set.

bundle/src/main/java/com/adobe/acs/commons/wcm/vanity/impl/VanityURLServiceImpl.java
```diff
@@ -0,0 +1,113 @@
...
+		params.put("property.value", vanityPath);
+		params.put("p.limit", "1");
+		params.put("p.guessTotal", "true");
+
+		final Query query = queryBuilder.createQuery(PredicateGroup.create(params), request.getResourceResolver().adaptTo(Session.class));
+		final SearchResult result = query.getResult();
+
+		if (result.getTotalMatches() > 0) {
+			if (log.isDebugEnabled()) {
+				log.debug("Found matching Sling vanity path [ {} ] on [ {} ]", vanityPath, result.getHits().get(0).getPath());
```
davidjgonzalez: its either resolve to 1 or none; figured 1 is better than none?

bundle/src/main/java/com/adobe/acs/commons/wcm/vanity/impl/VanityURLServiceImpl.java
davidjgonzalez: Added one @SachinMali merges my PR should come through ..

bundle/src/main/java/com/adobe/acs/commons/wcm/vanity/impl/VanityURLServiceImpl.java
```diff
@@ -0,0 +1,118 @@
...
+			return false;
+		}
+
+		request.setAttribute(VANITY_DISPATCH_CHECK_ATTR, true);
+
+		final String requestURI = request.getRequestURI();
+		final RequestPathInfo requestPathInfo = new PathInfo(request.getResourceResolver(), requestURI);
+
+		// Manually strip off any selectors or extensions from the URL
+		final String resourcePath = StringUtils.substringBefore(requestPathInfo.getResourcePath(), ".");
```
joerghoh: Is this string operation really sufficient to achieve this? Resources containing a "." in their name?

bundle/src/main/java/com/adobe/acs/commons/wcm/vanity/impl/VanityURLServiceImpl.java
```diff
@@ -0,0 +1,118 @@
...
+		// Limit to <pathScope> to get 1/2 to multi-tenant support
+		params.put("path", StringUtils.defaultIfEmpty(pathScope, DEFAULT_PATH_SCOPE));
+		params.put("property", NameConstants.PN_SLING_VANITY_PATH);
+		params.put("property.value", vanityPath);
+		params.put("p.limit", "1");
+		params.put("p.guessTotal", "true");
+
+		log.debug("Searching for vanity path [ {} ] under [ {} ]", vanityPath, pathScope);
+
+		final Query query = queryBuilder.createQuery(PredicateGroup.create(params), request.getResourceResolver().adaptTo(Session.class));
```
joerghoh: If I understood correctly, this query is executed on every request. Is my understanding correct? In that case this would have a huge performance impact even if this query is backed by an optimized Oak index.

davidjgonzalez: This query will only execute on 404'ing requests and use an property index that indexes on the property and should have ~0 or 1 results.

SachinMali: Yes this query will be executed on 'every request that is delegated to 404 page'. Regarding performance, David & I talked about it, I will let him explain.

davidjgonzalez: @joerghoh FWIW - AEM is constantly querying the uuid property index (just flip your logging to debug on queries and click do anything on Pub); its very fast since there are 0 or 1 results.

@chetanmeh any thoughts on the impact of having 404ing requests perform a query against a property index that will return 0 or 1 results (almost always unless someone has mis-authored content). Is executing a query of this profile problematic? It seems to run in 0 to 1ms for me; i expected this was ~ an look-up in a hash table that might have to be read from disk.

joerghoh: 404s happen quite often on AEM publishs. I know instances, where 50% of the traffic hitting the publishs cause 404 (because of the good hit-ration on publishs). I would try to avoid it or convince me by providing benchmark data :-)

davidjgonzalez: @joerghoh thoughts on how to avoid it? I could see adding some partial guards like:

- Provide a configurable request URI pattern that whitelists vanity path patterns (ex. ^/[a-z0-9]{3-20}) before querying

I don't think we have access to Sling's Bloom Filter which would be nice to leverage here.

 Im interested in what @chetanmeh has to say wrt to perf vs. that of the queries on UUID which are in every request (and aren't a problem).

Let's stew on this some more though. I'd love to benchmark, but i'm not sure i have the infra to do it justice :)

davidjgonzalez: oh ... i have an idea.

bundle/src/main/java/com/adobe/acs/commons/wcm/vanity/impl/VanityURLServiceImpl.java
davidjgonzalez: @joerghoh think i fixed this in the pending PR to come through... Basically, im just letting rr.map(..) do its thing and it leaves any selectors/extensions on.. but when passed through the vanity handling chain it seems to work.

That being said, do you have an example in mind? Keep in mind, at the point that this is executing a resource resolution on the actual URI has already happened and determined no resource can be resolved at the requestURI.

bundle/src/main/java/com/adobe/acs/commons/wcm/vanity/impl/VanityURLServiceImpl.java
chetanmeh: > Still interested wrt to the performance profile property indexes w ~0-1 or 1 matches.

It very much depends on query. 

    select * from [nt:base] where [jcr:uuid] = 'foo'

If the index used is a unique index and its a Mongo based setup then it would result in one read calls in Mongo for "/oak:index/uuid/:index/<uuid entry>". 

    select * from [nt:base] where [sling:resourceType] = 'type1'

However if its a property index then it may result in multiple Mongo calls as the index storage mirrors the content path of indexed content. So here it would have to read a structure like

```
oak:index
  + slinResourceType
     + :index
        + type1
            + content
               + en
                  ... //node structure mirroring indexed content
```
So for reading the first result for 'type1' it would need to read all those child nodes (if they have not been found in cache).

davidjgonzalez:
Replaced by #1067

PR #1036, opened 2017-07-03
Add source attachments for jar files

stefanseifert:
Having source attachments is useful when browsing class files in the Java IDE.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -86,6 +86,11 @@
                     <version>2.5</version>
                 </plugin>
                 <plugin>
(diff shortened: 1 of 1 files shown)


davidjgonzalez:
Should back-port to compat/6.0 branch

justinedelson:
We already create and deploy source JARs on release. They just aren't available on repo.adobe.com since that blocks the distribution of source JARs. They are available on bintray.

justinedelson:
As commented, this is already done on release.

PR #1040, opened 2017-07-06
#1039 - Health Check Status Emailer

### LINKED ISSUE #1039: Health Check Status E-mailer
It would be nice to have a way to have to periodically run sets of of Sling Health Checks and e-mail the results to a set of users. It would provide a simple, consistent digest across any AEM environment (even lower envs that might not enjoy prod-level monitoring). I can see two major uses:

1) Run a set of health checks frequently (every hour?)  and ONLY send e-mails if at least 1 failure occurs. (So as not to numb people to these emails)
2) Run a set of health checks less frequently (once a day/week?) and send the full status (All Success and Failures)

This has the achilles heel of if the SMTP server is unavailable, no e-mail will be sent. but hopefully the absence of emails from use case #2 is sufficient to identify this (along w other monitoring).

This should be an OSGI config factory so different schedulers w/ different configurations can be setup based on the multiple (2) uses outlined above.

## OSGi Factory Config
image

## Example E-mail

email

_note this screenshot has the success and failure #'s in the subject line flipped_

davidjgonzalez:
Fixed #1039

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/hc/impl/HealthCheckStatusEmailer.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/hc/impl/HealthCheckStatusEmailer.java
@@ -81,7 +81,7 @@ public class HealthCheckStatusEmailer implements Runnable {
     private static final int HEALTH_CHECK_STATUS_PADDING = 20;
     private static final int NUM_DASHES = 100;
 
-    private Calendar nextEmailTime = null;
+    private Calendar nextEmailTime = Calendar.getInstance();
 
     /* OSGi Properties */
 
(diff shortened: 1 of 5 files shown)


bundle/src/main/java/com/adobe/acs/commons/hc/impl/HealthCheckStatusEmailer.java
```diff
@@ -0,0 +1,246 @@
...
+@Properties({
+        @Property(
+                label = "Cron expression defining when this Scheduled Service will run",
+                name = "scheduler.expression",
+                description = "Every weekday @ 8am = [ 0 0 8 ? * MON-FRI * ] Visit www.cronmaker.com to generate cron expressions.",
+                value = "0 0 8 ? * MON-FRI *"
+        ),
+        @Property(
+                name = "scheduler.concurrent",
+                boolValue = false,
```
davidjgonzalez: Should this be false if it's a configuration factory? If something it set to run every 10 mins and something else is set to run every 24 hours, there could be a concurrency conflict; will this setting result in one of them not firing?

davidjgonzalez:
Should also add configuration for maxTimeout for these for longer running system connectivity HCs (like adobe.io!).

PR #1041, opened 2017-07-07
Feature/qrcode

davidjgonzalez:
Minor refactoring of edubey1's QR COde that was originally PR'd to ACS Tools.
This adds a QR Code button/panel to the AEM Sites' Page Editor (assuming its enabled). This allows for much easier QA (or review) of pages on mobile devices as the mobile device can scan the QR code, and then go directly to that page on AEM Publish without navigation, etc.

@edubey1 im not sure why i cant PR through your repo; but GH isnt giving me that option; your commits are part of this though.

image

image


content/src/main/content/jcr_root/apps/acs-commons/components/utilities/qr-code/clientlibs/js/app.js
```diff
@@ -0,0 +1,134 @@
...
+                'Content-Type': 'application/x-www-form-urlencoded'
+            }
+        }).
+        success(function (data, status, headers, config) {
+            $scope.app.running = false;
+            NotificationsService.add('success', "Success", "Your configuration has been saved");
+            NotificationsService.running($scope.app.running);
+
+            if ($scope.form.enable) {
+                $http({
```
davidjgonzalez: @justinedelson interested to get your thoughts on this little work around ;) .. I wanted to allow this to be enabled w/out developer support (ie. Dev deploying a proxy cq:ClientLibraryFolder node) but at the same time ensuring there is ZERO JS/CSS introduced into the Page Editor if this is not enabled.

justinedelson: @davidjgonzalez Possibly I'm missing something and need to test this locally, but at first glance... wouldn't this category property get blown away on the next deployment of ACS AEM Commons?

justinedelson:
Sorry if I'm repeating something I've already asked, but why is this configuration necessary? This would seem to be entirely duplicative of what is already configured by the Externalizer. I don't see a reason to have these mappings. Just look up the publish host from the Externalizer. Duplication of configuration leads to incorrect configuration.

davidjgonzalez:
image

* Re-wrote to use externalizer (pretty extensive :()
* Adjusted a few things to provide better isolation of the configured properties; So `/etc/acs-commons/qr-code/jcr:content/config` and `/etc/acs-commons/qr-code/jcr:content/clientlib-authoring` are both explicitly `mode=merge` .. it works (on 6.3) but will test on 6.2 since [1] implies it shouldn't (i think).

[1] http://jackrabbit.apache.org/filevault/importmode.html

davidjgonzalez:
@justinedelson any concerns w/ the way the clientlibs are "enabled" in this feature? 

The submission of the button actually updated the categories of a `cq:ClientLibraryFolder` to be `cq.authoring.editor.hook`  .. and the clientlib node itself is marked as "merge" in the filter.xml

Its a bit unorthodox, thus the bump of the q.

justinedelson:
@davidjgonzalez agree it is a bit unorthodox, but it seems like it works, so let's go for it.

Issue #1043, opened 2017-07-13
JcrPackages not closed in JcrPackageReplicationStatusEventHandler


kwin:
The packages being opened in https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/replication/status/impl/JcrPackageReplicationStatusEventHandler.java#L306 are nowhere being closed which leads to resource leakage.

PR #1045, opened 2017-07-13
Feature/1043 1044 - Package Replication improvements

davidjgonzalez:
Combines updated to JCR Package Replication #1043 and #1043

* Closes package after opening to prevent resource leakage
* Improves "replicated by" to default to the actual user that performs the replication. Legacy OSGi configuration available to allow a named user to be assigned for all package replication (leave this value blank to use the real user)

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/replication/status/impl/JcrPackageReplicationStatusEventHandler.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/replication/status/impl/JcrPackageReplicationStatusEventHandler.java
@@ -24,13 +24,13 @@ import com.adobe.acs.commons.packaging.PackageHelper;
 import com.adobe.acs.commons.replication.status.ReplicationStatusManager;
 import com.day.cq.jcrclustersupport.ClusterAware;
 import com.day.cq.replication.ReplicationAction;
+import com.day.cq.replication.ReplicationEvent;
 import com.day.cq.replication.ReplicationStatus;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.felix.scr.annotations.Activate;
 import org.apache.felix.scr.annotations.Component;
 import org.apache.felix.scr.annotations.ConfigurationPolicy;
-import org.apache.felix.scr.annotations.Deactivate;
 import org.apache.felix.scr.annotations.Properties;
 import org.apache.felix.scr.annotations.Property;
 import org.apache.felix.scr.annotations.PropertyOption;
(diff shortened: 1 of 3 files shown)


davidjgonzalez:
@kwin WDYT?

kwin:
Wow, that was fast. Thanks a lot. 

I think you close the `JcrPackage` too soon though. You should close it only once you no longer use the reference towards `JcrPackage`. Apart from that the method `getJcrPackageLastModified(...)` should rather rely on the method `JcrPackage.getPackage().getCreated()` instead of directly accessing nodes/properties as this returns the date when the package has been last built.

bundle/src/main/java/com/adobe/acs/commons/replication/status/impl/JcrPackageReplicationStatusEventHandler.java
```diff
@@ -231,7 +245,7 @@ public final JobResult process(final Job job) {
 
                     if (resources.size() > 0) {
                         replicationStatusManager.setReplicationStatus(resourceResolver,
-                                this.replicatedBy,
+                                StringUtils.defaultIfEmpty(replicatedBy, "Package Replication"),
```
kwin: Why do you need a fallback here as well, the property `PROPERTY_REPLICATED_BY` is never null/empty.

bundle/src/main/java/com/adobe/acs/commons/replication/status/impl/JcrPackageReplicationStatusEventHandler.java
```diff
@@ -309,6 +319,10 @@ private boolean containsJcrPackagePath(final String[] paths) {
                 }
             } catch (RepositoryException e) {
                 log.warn("Error checking if the path [ {} ] is a JCR Package.", path);
+            } finally {
+                if (jcrPackage != null) {
+                    jcrPackage.close();
```
kwin: this is too early, you are still referencing the jcr package afterwards

davidjgonzalez:
@kwin - FYI - tried this on 6.2 and replication event is null (fromRequest(..)) .. doing this between flights so haven't had a chance to look yet.

Thanks for other comments; will fix when we get to cruising altitude :)

davidjgonzalez:
@kwin so, getting replication action from the osgi event is only supported for grantite replication events, and not the old cq replication topics; might have to do it a bit more manually to handle both cases

kwin:
This PR looks much better now, but I would still recommend to fix the method to get the last built date (and rely internally on `JcrPackage.getPackage().getCreated()`). Right now the modification date on the package rather reflects the time the package definition has last been modified (i.e. via an updated filter) but not when the package has been last rebuilt.

davidjgonzalez:
@kwin date should be fixed; thanks

PR #1047, opened 2017-07-21
Feature/mcp (version 1)  "Greetings, programs!"

badvision:
This has the java-side of MCP ready to use, and with plenty of implementations to be immediately useful:
- Deep prune is a recursive folder removal tool, also useful for stopping sling jobs, purging workflow, etc.
- Asset ingestor (aka Anteater V2.0) is a fully-functional high-speed asset ingestion tool
- Folder relocator allows you to move or rename very large folders
- Asset reporter allows you to get a detailed report on how much space binaries for assets are using

But not only does this pull request contain these tools, but the reusable framework (MCP) which makes building these tools incredibly easy.


bundle/pom.xml
```diff
@@ -57,13 +57,16 @@
                             org.apache.sling.models.spi.injectorspecific.*;resolution:=optional, <!-- that package was added with Sling Models API 1.1 -->
                             com.day.cq.wcm.workflow.process;version="[6.0,8)", <!-- using a wider version range for forward compatibility -->
                             com.day.cq.mailer;version="[5.9,7)", <!-- using a wider version range for forward compatibility -->
+                            com.adobe.cq.sightly;version="[2.5,4)",
+                            com.day.cq.dam.api;version="[1.16,2)", <!-- using a wider version range for forward compatibility -->
```
justinedelson: This is almost certainly the wrong import range. It should probably be `[1.16,1.18)` since any change to the `Asset` interface after `1.17` could break this code.

bundle/src/main/java/com/adobe/acs/commons/fam/Failure.java
```diff
@@ -26,11 +27,15 @@
 public final class Failure {
     
     private String nodePath;
-    private Exception exception;
+    private String error;
+    private String stackTrace;
+    private Long time;
+    transient private Exception exception;
```
justinedelson: since Failure isn't serializable, why does it need transient fields?

bundle/src/main/java/com/adobe/acs/commons/fam/impl/ActionManagerImpl.java
```diff
@@ -48,14 +50,14 @@
 /**
  * Manages a pool of reusable resource resolvers and injects them into tasks
  */
-class ActionManagerImpl implements ActionManager {
+class ActionManagerImpl extends CancelHandler implements ActionManager {
 
-    private static final Logger LOG = LoggerFactory.getLogger(ActionManagerImpl.class);
+    transient private static final Logger LOG = LoggerFactory.getLogger(ActionManagerImpl.class);
```
justinedelson: since ActionManagerImpl isn't serializable (is it?), why does it need transient fields?

bundle/src/main/java/com/adobe/acs/commons/fam/package-info.java
```diff
@@ -17,5 +17,5 @@
  * limitations under the License.
  * #L%
  */
-@aQute.bnd.annotation.Version("2.2.0")
+@aQute.bnd.annotation.Version("3.0.0")
```
justinedelson: It is hard to tell, but I'm guessing that this is due to `ThrottledTaskManager` missing a `@ProviderType` annotation.

bundle/src/main/java/com/adobe/acs/commons/mcp/form/CheckboxComponent.java
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.adobe.acs.commons.mcp.form;
+
+/**
+ * Radio button selector component
+ */
+public class CheckboxComponent extends FieldComponent {
```
justinedelson: Does this class or really any other class in this package need to be exported?

justinedelson: Nevermind. I see them referenced from the annotation. Bummer. But the bulk of them look like `@ProviderType` to me.

badvision: I am assuming that people want to make their own MCP-based utilities so yes.  It has to be exported if people make their own process definitions.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/ControlledProcessManagerServlet.java
```diff
@@ -0,0 +1,166 @@
...
+import org.apache.sling.api.SlingHttpServletRequest;
+import org.apache.sling.api.SlingHttpServletResponse;
+import org.apache.sling.api.servlets.SlingSafeMethodsServlet;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Servlet for interacting with MCP.
+ */
+@SlingServlet(paths = "/bin/mcp")
```
justinedelson: Why is this using a path-mapped servlet (which are discouraged)?

bundle/src/main/java/com/adobe/acs/commons/mcp/form/CheckboxComponent.java
badvision: I had limited options to work with via annotations.  I had to choose between strings and class references.  With class references, the classes have to be present and exported.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/ControlledProcessManagerServlet.java
badvision: Were they?  I can change that.  What's the preference, selector-based?

bundle/src/main/java/com/adobe/acs/commons/mcp/model/ArchivedProcessFailure.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.apache.sling.api.resource.Resource;
+import org.apache.sling.models.annotations.DefaultInjectionStrategy;
+import org.apache.sling.models.annotations.Model;
+
+/**
+ * Persisted failure information.
+ */
+@Model(adaptables = Resource.class, defaultInjectionStrategy = DefaultInjectionStrategy.OPTIONAL)
+public class ArchivedProcessFailure {
```
justinedelson: Does this need to be exported?

bundle/src/main/java/com/adobe/acs/commons/mcp/model/ArchivedProcessInstance.java
```diff
@@ -0,0 +1,129 @@
...
+import org.apache.sling.models.annotations.Model;
+import org.apache.sling.models.annotations.Via;
+
+/**
+ *
+ */
+@Model(adaptables = Resource.class, defaultInjectionStrategy = DefaultInjectionStrategy.OPTIONAL)
+public class ArchivedProcessInstance implements ProcessInstance {
+    @Inject
+    transient private Resource resource;
```
justinedelson: not serializable == no transient fields

bundle/src/main/java/com/adobe/acs/commons/mcp/model/ArchivedProcessFailure.java
badvision: Actually no -- but it is a model bean of sorts.  Should I make an IMPL sub-folder under model and hide it in there?

bundle/src/main/java/com/adobe/acs/commons/mcp/model/AvailableProcessDefinitions.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.commons.lang3.reflect.FieldUtils;
+import org.apache.sling.api.scripting.SlingScriptHelper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Produce a list of available process definitions that can be started by the
+ * user.
+ */
+public class AvailableProcessDefinitions extends WCMUsePojo {
```
justinedelson: IMO, using WCMUsePojo and Sling Models in the same page is a bit of a weird programming model, but I guess it works.

bundle/src/main/java/com/adobe/acs/commons/mcp/model/ArchivedProcessInstance.java
badvision: This was because of how GSON works.  If I don't mark things as transient it tries to serialize way too much of the object graph and gets hung in infinite loops.

bundle/src/main/java/com/adobe/acs/commons/mcp/model/AvailableProcessDefinitions.java
badvision: It is very, very weird.  But the goal was to handle the ugliness in java so that the component side could be expressed cleanly.  Had to pay the price somewhere.  Sightly doesn't give you a lot of other options.

bundle/src/main/java/com/adobe/acs/commons/fam/package-info.java
badvision: I did introduce new FAM behaviors in this update: The task manager and action managers now support a way to halt a running task dead in its tracks.  This changed method signatures but was absolutely necessary to support a stop button.

bundle/src/main/java/com/adobe/acs/commons/mcp/processes/AssetReport.java
```diff
@@ -0,0 +1,296 @@
...
+                FrozenAsset assetVersion = new FrozenAsset(asset, rev);
+                String versionHash = assetVersion.getMetadataValue(SHA1);
+                if (versionHash != null) {
+                    if (observedHashes.contains(versionHash)) {
+                        return;
+                    } else {
+                        observedHashes.add(versionHash);
+                    }
+                }
+                Long size = getTotalAssetSize(assetVersion);
```
justinedelson: should be `long` not `Long`

bundle/src/main/java/com/adobe/acs/commons/fam/impl/ActionManagerImpl.java
badvision: So that GSON doesn't serialize it in the mcp rest servlet.  The way it works is that GSON serializes a bunch of stuff and serves it up to the front-end so that it can do live status reporting, etc.

bundle/src/main/java/com/adobe/acs/commons/fam/Failure.java
badvision: Because GSON (sorry, explained it in a couple of other places already)

bundle/pom.xml
badvision: 6.3 uses 1.22.0 but when I tried to use 1.22.0) it still didn't work.   Should I try [1.16,1.23) ?

bundle/src/main/java/com/adobe/acs/commons/mcp/processes/AssetReport.java
badvision: Are we going to get in an auto-boxing match?  (j/k, I'll change it... lol)

bundle/src/main/java/com/adobe/acs/commons/mcp/processes/AssetReport.java
```diff
@@ -0,0 +1,296 @@
...
+                tabulate(folderPath, Column.subasset_count, 1);
+                long size = subasset.getRenditions().stream().collect(Collectors.summingLong(Rendition::getSize));
+                tabulate(folderPath, Column.subasset_size, size);
+                tabulate(folderPath, Column.combined_size, size);
+            });
+        }
+        if (includeVersions) {
+            for (Revision rev : asset.getRevisions(null)) {
+                tabulate(folderPath, Column.version_count, 1);
+                FrozenAsset assetVersion = new FrozenAsset(asset, rev);
```
justinedelson: I'm not convinced that `FrozenAsset` needs to exist. It looks like the only reason this object is created is to calculate the size of an asset version and get the SHA1 property. And while there is some obvious convenience to being able to use the same method (`getTotalAssetSize()`) to calculate a current version and an older version, if the cost that comes at is potential changes every time the `Asset` interface changes, I'm not sure the juice is worth the squeeze.

badvision: I adopted this approach because Clint already did the hard part of implementing this, so it seemed the shortest path.  My original approach was to walk the tree the hard way.  Yes the overall goal is to count the size of binaries both for the versioned asset's renditions as well as its versioned subassets.  When you consider how much stuff is being stored in a version you'll quickly see why keeping too many versions is a very bad thing.

badvision: Or, to put it another way, I did it this way because I had a total of 80 hours to write the whole solution and it was either this approach or not have enough time for tabulating versions and leave that feature out.

justinedelson:
So high level comments...

1. The API surface here is gigantic and I'm not sure all of it is necessary. I'd ask that you take a second look at each exported class and make sure it is absolutely necessary. It is generally a bad sign when you have `@Component` classes in an exported package.
2. Coverage is going in the wrong direction.
3. The FrozenAsset thing is really problematic. I'll think a little about a different solution.
4. Missing ProviderType annotations (I think), although this goes along with the first item

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/ControlledProcessManagerServlet.java
justinedelson: See https://sling.apache.org/documentation/the-sling-engine/servlets.html

bundle/src/main/java/com/adobe/acs/commons/fam/package-info.java
justinedelson: sure, but why do those changes require a major version change? It seems like those changes only pose a backwards compatibility issue for providers (i.e. FAM itself), so the major version change is likely due to missing metadata.

badvision:
It is an API for creating new solutions so most (if not all) of the exports are intentional.  I think that some of the model beans can be hidden (such as the archived ones) since I already export the interface.  What is the recommended package naming convention for that? ...model.impl?

bundle/src/main/java/com/adobe/acs/commons/fam/impl/ActionManagerImpl.java
justinedelson: ah. would it make more sense to use `@Expose` and then intentionally expose the important fields? IMO, using transient is just hella non-intuitive (although when you explain it, it makes sense)

badvision: In some places I have to make the logger transient because of GSON but it makes sense to make sure that the class is serializable so that it is more clear what is happening.   Other places that field is transient in classes that are not serialized ever so it makes so sense.  I've done a pass through and cleaned that up.  The PR will be updated shortly.

bundle/src/main/java/com/adobe/acs/commons/fam/package-info.java
badvision: Yep, adding @ProviderType let me use 2.3.0 as a version number.  Thanks!

bundle/src/main/java/com/adobe/acs/commons/mcp/form/CheckboxComponent.java
badvision: I hadn't looked at it from an objective lens of Consumer vs. Provider type but yeah, most of this is ProviderType with only few exceptions.  I'll update those.

bundle/pom.xml
justinedelson: Yes, that's what I meant. But see https://github.com/badvision/acs-aem-commons/pull/18 which gets rid of this need.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/ControlledProcessManagerImpl.java
```diff
@@ -0,0 +1,199 @@
...
+@Component
+@Service(ControlledProcessManager.class)
+@Property(name = "jmx.objectname", value = "com.adobe.acs.commons:type=Manage Controlled Processes")
+public class ControlledProcessManagerImpl implements ControlledProcessManager {
+    private static final org.slf4j.Logger LOG = LoggerFactory.getLogger(ControlledProcessManagerImpl.class);
+
+    private static final String SERVICE_NAME = "manage-controlled-processes";
+    private static final Map<String, Object> AUTH_INFO;
+    @Reference(cardinality = ReferenceCardinality.MANDATORY_MULTIPLE, bind = "bindDefinition", unbind = "unbindDefinition", referenceInterface = ProcessDefinition.class, policy = ReferencePolicy.DYNAMIC)
+    private final List<ProcessDefinition> processDefinitions = Collections.synchronizedList(new ArrayList<>());
```
justinedelson: would CopyOnWriteArrayList be better?

justinedelson:
Yeah, I understand it is going to have a large API. I'm just questioning how large. Keep in mind that it is always easier (trivial in fact) to take something which isn't API and make it API but the reverse is never true.

I think `model.impl` is fine.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/ControlledProcessManagerServlet.java
davidjgonzalez: Typically we attach the POST behavior to the resource that is exposing this UI view via a selector (or occasionally sub-resource). So if this tool is exposed via say `/etc/acs-commons/mcp` ... `HTTP POST` to `/etc/acs-commons/mcp/jcr:content.process.json` (and that  `jcr:content` has a `'sling:resourceType` that maps to this servlet + selector). 

Since we have to bring the resource to expose the WebUI, it makes sense to use that for other interactions w the tool as well.

badvision:
Ok I have coveralls happy now, @justinedelson and @davidjgonzalez  --- any other showstoppers?

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/ControlledProcessManagerServlet.java
```diff
@@ -0,0 +1,166 @@
...
+    ControlledProcessManager manager;
+
+    @Override
+    protected void doGet(SlingHttpServletRequest request, SlingHttpServletResponse response) throws ServletException, IOException {
+        String action = request.getParameter("action");
+        Object result = null;
+        try {
+            switch (action) {
+                case "start":
+                    result = doStartProcess(request);
```
justinedelson: this (and several of these for that matter) are not idempotent and thus shouldn't happen in the context of a GET request. probably simpler to just change this whole thing to POST than splitting it up

justinedelson:
I somehow missed the front-end code in my initial review. Sorry about that. I see a lot of coral classes used. Are you sure that all of these are "public" classes. Per https://docs.adobe.com/docs/en/aem/6-3/develop/ref/coral-ui/coralui3/styles.html, only public classes are considered API. For example, I don't see `coral-table-inner` on https://docs.adobe.com/docs/en/aem/6-3/develop/ref/coral-ui/coralui3/styles.html#coral-Table. Just want to make sure we aren't opening ourselves up to a compatibility issue later.

PR #1052, opened 2017-07-27
implement email service timeouts. fixes #1051

### LINKED ISSUE #1051: Emails sent via EmailService do not have connection/socket timeouts
Emails sent using `com.adobe.acs.commons.email.EmailService` do not have the appropriate timeout properties set on the `org.apache.commons.mail.Email` object and there is no way for a client to set them since the `Email` object is managed entirely internally.

justinedelson:
ugh. not sure what's up with the formatting changes in the Email test. I really just added two tests.


justinedelson:
giving up on getting coveralls to recognize that EmailServiceImpl is tested (at least for now)

Issue #1010, opened 2017-05-23
Error page handler cache serves content as ISO-8601 when original content is UTF-8


davidjgonzalez:
Error page handler should set the char encoding on the served response using the same format as the cached response.

justinedelson:
@davidjgonzalez this is fixed, right?

davidjgonzalez:
Nope  - killed the PR cause it was wrong

davidjgonzalez:
We can defer to 3.12.0

joerghoh:
What is the current status here?

davidjgonzalez:
Status: not fixed

PR #1062, opened 2017-07-29
#1010 - Support caching of contentType for error pages

davidjgonzalez:
Fixes issue with contentType setting in Error Page Handler. #1010

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/errorpagehandler/cache/impl/CacheEntry.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/errorpagehandler/cache/impl/CacheEntry.java
@@ -26,6 +26,8 @@ import java.util.Date;
 import java.util.concurrent.atomic.AtomicInteger;
 
 class CacheEntry {
+    private String contentType;
+
     private String data;
 
     private final AtomicInteger hits;
@@ -39,6 +41,7 @@ class CacheEntry {
         this.misses = new AtomicInteger();
         this.data = "";
         this.expiresAt = new Date(0);
+        this.contentType = "UTF-8";
     }
 
     public final String getData() {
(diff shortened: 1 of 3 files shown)


bundle/src/main/java/com/adobe/acs/commons/errorpagehandler/cache/impl/ErrorPageCacheImpl.java
```diff
@@ -98,6 +96,7 @@ public String get(final String path,
             }
 
             cacheEntry.setData(data);
+            cacheEntry.setContentType(request.getContentType());
```
justinedelson: `request.getContentType()` returns the content type of the request body. I don't think that's what you want. As far as I can tell, getting the response content type would require a new method on `ResourceDataUtil` which would return a tuple of the response body and the response content type.

bundle/src/main/java/com/adobe/acs/commons/errorpagehandler/cache/impl/CacheEntry.java
```diff
@@ -39,6 +41,7 @@ public CacheEntry() {
         this.misses = new AtomicInteger();
         this.data = "";
         this.expiresAt = new Date(0);
+        this.contentType = "UTF-8";
```
justinedelson: is the intention to hold the content type (mime type)? or the character encoding? Seems like maybe both, in which case they should be stored separately so that the encoding can be used in the `getBytes()` method.

bundle/src/main/java/com/adobe/acs/commons/errorpagehandler/cache/impl/CacheEntry.java
```diff
@@ -106,6 +115,8 @@ final int getTotal() {
     }
 
     final int getBytes() {
-        return getData().getBytes(Charset.forName("UTF-8")).length;
+        return getData().getBytes(Charset.forName(getContentType())).length;
```
justinedelson: seems like this will fail if contentType is actually set to a mime type.

davidjgonzalez:
Rejected/Closing for this release

Issue #1065, opened 2017-08-02
Bulk page relocator utility


badvision:
As an author I would like a more convenient way to move pages and also have a report of what was moved. I don't want it to work unless I have all the right permissions, and finally I want more control over how the pages are activated at the end.

PR #1066, opened 2017-08-02
Feature/page relocator

badvision:
#1065 : Page Relocator utility

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/fam/impl/ActionManagerImpl.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/fam/impl/ActionManagerImpl.java
@@ -52,6 +52,7 @@ import org.slf4j.LoggerFactory;
  * Manages a pool of reusable resource resolvers and injects them into tasks
  */
 class ActionManagerImpl extends CancelHandler implements ActionManager, Serializable {
+
     private static final long serialVersionUID = 7526472295622776150L;
 
     transient private static final Logger LOG = LoggerFactory.getLogger(ActionManagerImpl.class);
@@ -94,7 +95,6 @@ class ActionManagerImpl extends CancelHandler implements ActionManager, Serializ
     public String getName() {
         return name;
     }
-
 
     @Override
     public int getAddedCount() {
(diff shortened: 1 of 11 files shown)


badvision:
I had to work around some serious limitations of how PageManager works today.  I still have to see if references are all being handled or if I need to add additional stuff to check for references before using PageManager... it's not clear at all in the javadocs.  But as for the rest of it: It does move pages, and it does have the ability to tackle referral of replication to the end.  And hey, test coverage!  What's not to love? :)

bundle/src/main/java/com/adobe/acs/commons/fam/impl/ActionManagerImpl.java
```diff
@@ -52,6 +52,7 @@
  * Manages a pool of reusable resource resolvers and injects them into tasks
  */
 class ActionManagerImpl extends CancelHandler implements ActionManager, Serializable {
+
     private static final long serialVersionUID = 7526472295622776150L;
 
     transient private static final Logger LOG = LoggerFactory.getLogger(ActionManagerImpl.class);
```
davidjgonzalez: project uses "log" vs "LOG"

badvision: hmm... I usually capitalize statics... I can change it.

badvision: But TBH I'd rather just change that outside of the PR since the PR didn't introduce that code per se.

badvision: The reason I touched ActionManagerImpl in this PR was to change the order of when it does work vs. when it updates its "added" counter.  That makes a difference in how well the junit test performs.  It has no real effect on the real operation since it's async anyway.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/ReplicatorQueue.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import javax.jcr.Session;
+
+/**
+ * Collect a list of replication events for later examination and/or replay.
+ */
+public class ReplicatorQueue implements Replicator {
+    Map<String, ReplicationOptions> deactivateOperations = new LinkedHashMap<>();
```
davidjgonzalez: Is this class's state thread safe? It looks like any PageRelocator processes will be reading/writing from this same object in an unsynchronized manner.

badvision: Good catch.  Used the Collections adapters to fix that.

Issue #1063, opened 2017-07-31
Email Service adds <pre></pre> tag around templates


sagarmandal:
If someone wants to use full HTML (`<html><head>...</head><body>...</body></html>`) template, then whole HTML gets wrapped in `<pre></pre>` tags, which breaks the HTML formatting/styles of he template.

justinedelson:
based on the feedback in #1070, this is a product bug.

pat-lego:
This is because it uses it uses this https://github.com/apache/commons-email/blob/master/src/main/java/org/apache/commons/mail/HtmlEmail.java#L90-L92 and it hard coded to wrap it with pre.

PR #1070, opened 2017-08-09
Custom MailTemplate added to to use HtmlEmail.setHtmlMsg(), if HTML t…

sagarmandal:
Custom MailTemplate added in acs-common to to use HtmlEmail.setHtmlMsg(), if HTML template contains full HTML


justinedelson:
@sagarmandal Is the implication of this PR that the MailTemplate class in AEM is the source of this #1063?

sagarmandal:
@justinedelson Yes. MailTemplate class in AEM is the source of this #1063

justinedelson:
If there is a bug in AEM, please report it to DayCare. ACS AEM Commons is not really the place for that.

PR #1076, opened 2017-08-14
update changelog for PageRelocator

justinedelson:
@badvision - is this OK?


badvision:
Yep... I think I spotted a UX bug... minor one.  Shouldn't take much to fix it.

PR #1083, opened 2017-08-17
Fix reflection issue

badvision:
MCP is unable to start processes defined in other bundles due to a naive useage of Class.forName


bundle/src/main/java/com/adobe/acs/commons/mcp/impl/ControlledProcessManagerImpl.java
```diff
@@ -133,7 +133,15 @@ public ProcessDefinition findDefinitionByNameOrPath(String nameOrPath) throws Re
     }
 
     private ProcessDefinition findDefinitionByName(String name) throws ReflectiveOperationException {
-        Class defClass = Class.forName(name);
+        Class clazz;
+        try {
+            clazz = Class.forName(name);
```
justinedelson: I recognize that `Class.forName()` was already used here, so perhaps this should be a separate issue, but... is there a reason to use this vs. just looking up the processDefinition list? IIOW, should the code in the `catch` block be the *only* code path?

The problem is that `Class.forName()` won't work if the class is outside of the bundle. We can deal with that using `Dynamic-Package-Import`, but that's a pretty heavy weight solution.

badvision: Well, technically Class.forName is faster for classes in the current bundle but I think you're right -- the catch block realistically is the only thing needed and class.forName can go away.

badvision:
Ok, everything checks out now.  Does this look better now @justinedelson?

PR #1085, opened 2017-08-18
remove use of reflection to create new ProcessDefinition instances. i…

justinedelson:
…nstead, use the factory pattern


justinedelson:
@badvision WDYT about this as an alternate to #1083?

badvision:
I think this is supposedly what the factory pattern is supposed to do, right?  However, the getName method is supposed to be a friendly string for the display in the UI, not the class name.  Otherwise I think it is functionally equivalent and breaks fewer rules than the evil stuff I wrote.  Is there a way we can let getName live in the ProcessDefinitionFactory and somehow nix it from the ProcessDefinition?  That's one challenge.  The other challenge is the way I used HiddenProcessDefinition as a kind of hook for later use.  Right now it takes no methods but if the definition implements it, that hides it from Admin -- if using a factory, I think the factory has to take that concern as well.

justinedelson:
at present, `getName()` has to be part of the `ProcessDefinition` API since it is used in the output of `ProcesssInstanceImpl`

justinedelson:
I did, however, switch the permission model to pass the actual `User` object into the definition factory. I created two default implementations -- one for just the `admin` user and one for the `admin` user plus members of the `administrators` group.

badvision:
Yeah I like that part -- people were asking for more flexibility in there and your approach makes total sense.  I revised my pull request to incorporate your ideas as well.  However, I changed it a little to use generics and abstract classes so I could enforce that a name is the same in both places -- convention is fine sometimes but in this case I don't want the developer to have the opportunity to mess it up.

davidjgonzalez:
@justinedelson @badvision whats the story w this and #1083 ? Do we want to have this block release? I'd like to cut a release now.

justinedelson:
@davidjgonzalez should be good to release now. thanks.

davidjgonzalez:
@justinedelson thanks!!

PR #1091, opened 2017-08-28
Detect a multiselect within a composite multifield. Fixes #1090.

### LINKED ISSUE #1090: Allow multi-select inside a composite mutlifield
Composite multifield assumes that a field value of type array means that the value is again a multifield. This is not always true; a multi-select field also has a value of type array. As a result, multi-select fields are not correctly populated upon load. The code should distinguish between the two.

FrankDeGrootIsobar:
Detect a multiselect within a composite multifield, and populate accordingly on load. Fixes #1090.

### DIFF
--- a/content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/composite-multifield/source/touchui-composite-multifield.js
+++ b/content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/composite-multifield/source/touchui-composite-multifield.js
@@ -198,7 +198,7 @@
                                 cmf.setWidgetValue($field.siblings( "input.autocomplete-has-suggestion-btn"), rValue);
                             }
 
-                            if (_.isArray(rValue) && !_.isEmpty(rValue) && $field.prop("type") != "select-multiple") {
+                            if (_.isArray(rValue) && !_.isEmpty(rValue) && !cmf.isSelectMultiple($field)) {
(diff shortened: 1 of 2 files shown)


content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/composite-multifield/source/touchui-composite-multifield.js
```diff
@@ -198,7 +198,7 @@
                                 cmf.setWidgetValue($field.siblings( "input.autocomplete-has-suggestion-btn"), rValue);
                             }
 
-                            if (_.isArray(rValue) && !_.isEmpty(rValue)) {
+                            if (_.isArray(rValue) && !_.isEmpty(rValue) && $field.prop("type") != "select-multiple") {
```
justinedelson: minor issue - is there a way to remove the duplication of this check with the code in `touchui-widgets-init.js`? i.e. something like `&& !isSelectMultiple($field)` ?

FrankDeGrootIsobar: Updated to call isSelectMultiple() instead of duplicating the test

justinedelson:
thanks for making that change. Looks good to me. @schoudry any thoughts before merging?

schoudry:
yes looks good

justinedelson:
thanks @FrankDeGrootIsobar

Issue #1095, opened 2017-08-30
Move ACS Tools Tag Maker to use MCP


davidjgonzalez:
It would be  nice to move ACS Tools Tag Maker to ACS Commons and execute it via MCP.

PR #1096, opened 2017-08-30
#1095 - Tag Creator process for MCP (previously ACS Tools Tag Maker)

davidjgonzalez:


bundle/src/main/java/com/adobe/acs/commons/mcp/form/FieldComponent.java
```diff
@@ -133,7 +133,7 @@ public String getName() {
 
     public boolean hasOption(String optionName) {
         return Stream.of(formField.options())
-                .filter(s -> s.equalsIgnoreCase(optionName))
+                .filter(s -> s.equalsIgnoreCase(optionName) || s.startsWith(optionName + "="))
```
davidjgonzalez: @badvision does this look OK?

bundle/src/main/java/com/adobe/acs/commons/mcp/util/AnnotatedFieldDeserializer.java
```diff
@@ -84,7 +86,13 @@ private static void parseInput(Object target, ValueMap input, Field field) throw
             if (value.getClass().isArray()) {
                 val = ((Object[]) value)[0];
             }
-            parseInputValue(target, String.valueOf(val), field);
+
+            if (val instanceof InputStream) {
```
davidjgonzalez: @badvision Changed this to handle file uploads.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/ControlledProcessManagerServlet.java
```diff
@@ -159,12 +161,29 @@ private ProcessInstance getProcessFromRequest(SlingHttpServletRequest request) {
...
-    private Map<String, Object> convertRequestMap(Map<String, String[]> parameterMap) {
-        return parameterMap.entrySet().stream()
+    private Map<String, Object> convertRequestMap(RequestParameterMap requestParameterMap) {
+        return requestParameterMap.entrySet().stream()
                 .filter(entry -> !ignoredInputs.contains(entry.getKey()))
                 .collect(Collectors.toMap(
                         entry -> entry.getKey(),
-                        entry -> (Object) (entry.getValue().length == 1 ? entry.getValue()[0] : entry.getValue())
+                        entry -> {
+                            final RequestParameter[] values = entry.getValue();
```
davidjgonzalez: @badvision Changed this to handle file uploads.

bundle/src/main/java/com/adobe/acs/commons/mcp/form/FieldComponent.java
badvision: Yes, absolutely!  Thanks.  That was a future area of expansion and the next logical step for that bit.  I didn't like having to use a string array but the fact is that annotations provide you very few options so it was a necessary evil in order to pass less-structured data along.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/ControlledProcessManagerServlet.java
badvision: Oh, good thinking!  Just keep in mind it was designed such that single value = object, array => Array of objects, etc.  I haven't hooked up a component that generates multiple values as inputs into the servlet so this was an area of future expansion for multivalue fields, etc.

bundle/src/main/java/com/adobe/acs/commons/mcp/util/AnnotatedFieldDeserializer.java
badvision: Makes sense.

badvision:
Looks great and I like that you added even more form components for folks to use!  Very cool.  Note that there's no need for NAME to be a static field on the process definition.  I was able to re-work things such that only the factory's getName() method has to return that value so it's not needed in more than one spot.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/ControlledProcessManagerServlet.java
davidjgonzalez: Not sure I follow this ...

content/src/main/content/jcr_root/apps/acs-commons/components/utilities/manage-controlled-processes/clientlibs/js/scriptRunner.js
```diff
@@ -109,17 +109,20 @@ var ScriptRunner = {
         });
     },
     startProcess: function () {
-        var data = {};
-        jQuery("#processDefinitionInput form", window.top.document).serializeArray().map(function (x) {
-            data[x.name] = x.value;
-        });
+        /* Use FormData object to support file uploads */
```
davidjgonzalez: @badvision - just want to make sure you saw this change as well; I didn't quite grok why you were doing the map to the data object.

badvision:
It all looks good to me but I don't have the final say to accept and merge the PR.

Issue #1094, opened 2017-08-29
QR code contains bug that makes it log WARN/ERROR all the time


royteeuwen:
With a standard AEM 6.3 CFP2, when opening for example the /content/we-retail/us/en/men, you default get following logs in the error.log, making the logs very bloated and wrong. This should definitely be an ERROR, and I don't know why a WARN would be logged either on a standard with no configuration

```
29.08.2017 23:12:20.908 *WARN* [0:0:0:0:0:0:0:1 [1504041140907] GET /etc/acs-commons/qr-code/_jcr_content/config.json/content/we-retail/us/en/user.html HTTP/1.1] com.day.cq.wcm.core.impl.components.ComponentCacheImpl No component node found at /apps/acs-commons/components/utilities/qr-code/config
29.08.2017 23:12:20.908 *ERROR* [0:0:0:0:0:0:0:1 [1504041140907] GET /etc/acs-commons/qr-code/_jcr_content/config.json/content/we-retail/us/en/user.html HTTP/1.1] com.adobe.acs.commons.wcm.impl.QrCodeServlet Externalizer configuration for AEM Publish did not yield a valid URL
```

davidjgonzalez:
@royteeuwen is this without QR Code enabled at all?

davidjgonzalez:
@royteeuwen can you try this [1] out? You may need to toggle QR Code on and off [1] (to clear the clientlib category setting) and also clear you browser cache.

[1] [acs-aem-commons-content-3.10.1-SNAPSHOT-min.zip](https://github.com/Adobe-Consulting-Services/acs-aem-commons/files/1262122/acs-aem-commons-content-3.10.1-SNAPSHOT-min.zip)
[2] http://localhost:4502/etc/acs-commons/qr-code.html

royteeuwen:
Perfect, this seems to fix it, thanks a lot @davidjgonzalez

davidjgonzalez:
@royteeuwen thanks for confirming - will likely cut a build this weekend w the fix. Don't have much time after work this week to do much :(

royteeuwen:
@davidjgonzalez no worries, I know the feeling, thanks a lot :)!

PR #1097, opened 2017-08-30
#1094 - QR Code bug

davidjgonzalez:
* Defaulted QR Code to be off (cq.author.editor.hook clientlib category was incorrectly set by default)
* May require toggling QR Code button to clear existing category
* Dropped some log levels to WARN


davidjgonzalez:
@royteeuwen WDYT?

Issue #1101, opened 2017-09-01
Add feature to grab publish url of an asset or a page


kunalgaba:
Business users quickly want to grab publish url for sharing or for using it in external tools like Target. 

<img width="1132" alt="screen shot 2017-08-31 at 5 13 27 pm" src="https://user-images.githubusercontent.com/388029/29964714-96a7b354-8f28-11e7-966f-01881be2b809.png">

davidjgonzalez:
Makes sense to me; I assume his only shows of the asset is published? Or has some indication that it is/isn't from this UI?

kunalgaba:
Yes the button will be hidden if the asset is not published.

davidjgonzalez:
Looks like a cool feature! TBH surprised ita not part of the product :)

kunalgaba:
Yes. I was surprised too. I can commit this if you grant me commit rights.

davidjgonzalez:
@kunalgaba can you PR it as described here?

https://adobe-consulting-services.github.io/acs-aem-commons/pages/contribute.html

PR #1106, opened 2017-09-05
#1101 Added a button to grab the publish url of a published asset.

kunalgaba:


content/src/main/content/META-INF/vault/filter.xml
```diff
@@ -83,4 +83,5 @@
     <filter root="/etc/workflow/instances/rep:policy"/>
     <filter root="/home/users/rep:policy"/>
     <filter root="/var/acs-commons/rep:policy"/>
+    <filter root="/apps/dam/gui/content/assets"/>
```
davidjgonzalez: This feels too intrusive to include by default. How can we make this an "opt-in" feature?

content/src/main/content/jcr_root/apps/acs-commons/authoring/dam-copy-publishurl/.content.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<jcr:root xmlns:cq="http://www.day.com/jcr/cq/1.0" xmlns:jcr="http://www.jcp.org/jcr/1.0"
+    jcr:primaryType="cq:ClientLibraryFolder"
+    categories="dam.copy.publishurl"/>
```
davidjgonzalez: we typically prefix our clientlibs with "acs-aem-commons" to namespace them; its unlikely someone else would use "dam.copy.publishurl" but not entirely unreasonable. Check out some of the similar clientlibs for the CL naming conventions.

content/src/main/content/jcr_root/apps/acs-commons/authoring/dam-copy-publishurl/dialog.css
```diff
@@ -0,0 +1,12 @@
+.publish-url-button {
```
davidjgonzalez: can we namespace these classes?

```
.acs-aem-commons__dam-copy-published-url__button { ... }
.acs-aem-commons__dam-copy-published-url__text { ... }
```

This ~ guarantees no collisions.

content/src/main/content/jcr_root/apps/dam/gui/content/assets/metadataeditor/.content.xml
```diff
@@ -0,0 +1,85 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
davidjgonzalez: Same as before; how can we make this feature opt-in?

content/src/main/content/META-INF/vault/filter.xml
kunalgaba: I think then we can create a feature flag for this and write a render condition to show/hide this button based on the feature flag. 
https://sling.apache.org/documentation/the-sling-engine/featureflags.html 

What do you think ? But I will have to overlay the nodes from libs to add the button in the properties UI.

justinedelson: I like the feature flag approach, but this is a very broad filter. It should be as narrow as possible to avoid conflicting with other customizations.

content/src/main/content/jcr_root/apps/dam/gui/content/assets/metadataeditor/.content.xml
```diff
@@ -0,0 +1,85 @@
...
+                            variant="primary">
+                            <parentConfig
+                                jcr:primaryType="nt:unstructured"
+                                close="{Boolean}true"/>
+                        </dismiss>
+                    </footer>
+                </dialogpublishurl>
+            </items>
+        </content>
+        <head jcr:primaryType="nt:unstructured">
```
justinedelson: does this really need to be here? again, this potentially conflicts with other customizations.

content/src/main/content/jcr_root/apps/acs-commons/authoring/dam-copy-publishurl/.content.xml
justinedelson: I think this should be added to one of the existing clientlibs used on this page already. That way, the `head` node doesn't need to be overlayed.

justinedelson: `dam.gui.coral.metadataeditor` perhaps

content/src/main/content/jcr_root/apps/dam/gui/content/assets/metadataeditor/.content.xml
kunalgaba: I have added a feature flag to show/hide this button.

content/src/main/content/jcr_root/apps/acs-commons/renderconditons/isfeatureenabled/isfeatureenabled.jsp
```diff
@@ -0,0 +1,33 @@
+<%
```
justinedelson: I don't think this is needed. There is already `granite/ui/components/foundation/renderconditions/feature`

bundle/src/main/java/com/adobe/acs/commons/dam/impl/CopyAssetPublishUrlFeature.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.sling.featureflags.Feature;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Map;
+
+@Service(value = Feature.class)
+@Properties({@Property(name = "service.ranking", intValue = 200)})
+@Component(label = CopyAssetPublishUrlFeature.LABEL, description = CopyAssetPublishUrlFeature.DESCRIPTION, metatype = true)
+public class CopyAssetPublishUrlFeature implements Feature {
```
justinedelson: Can't we just use the Configured Feature for this and not have any code?

content/src/main/content/jcr_root/apps/acs-commons/renderconditons/isfeatureenabled/isfeatureenabled.jsp
```diff
@@ -0,0 +1,33 @@
...
+    Features features = sling.getService(org.apache.sling.featureflags.Features.class);
+    ExpressionHelper expressionHelper = new ExpressionHelper(sling.getService(ExpressionResolver.class), pageContext);
+
+    boolean vote = false;
+    if (resolver != null) {
+
+         Config cfg = new Config(resource);
+         String featureFlag = expressionHelper.getString(cfg.get("featureFlag", String.class));
+
+         if (features.isEnabled(featureFlag)) {
```
andyklimczak: could we simplify with `vote = features.isEnabled(featureFlag)`?

bundle/src/main/java/com/adobe/acs/commons/dam/impl/CopyAssetPublishUrlFeature.java
kunalgaba: I did not get it. Which feature you are talking about ?

justinedelson: There's a generic component in Sling which allows for features to be defined purely via configuration. The PID for this is `org.apache.sling.featureflags.impl.ConfiguredFeature`, so you create a configuration like `org.apache.sling.featureflags.impl.ConfiguredFeature-myfeature`, set the `enabled` property to true and the `name` property to the feature name.

justinedelson:
Thanks for making all of these changes.

justinedelson:
Merged with a few changes, mostly formatting and minor renaming.

The two significant changes:
* Removed the ConfiguredFeature config. This will have to be done as an opt-in (and needs to be documented).
* Removed the exclusion for the iPad showing the copy button and instead am hiding the button based on browser capability.

PR #1111, opened 2017-09-13
Add a dependency to the cq-content bundle

### LINKED ISSUE #1110: ACS Commons is installed before the primary types are available
Hey,

I have the following setup:

The AEM jars gets extracted with java -jar aem.jar -unpack
The install folder is made in crx-quickstart
All my dependencies are put in the install folder so that they get installed on startup

The problem that acs-aem-commons (and acs-aem-tools) is having is that no dependency is set on AEM itself, so the content gets installed before even de cq-content is available, producing following errors:

13.09.2017 12:06:59.763 *ERROR* [OsgiInstallerImpl] org.apache.jackrabbit.vault.fs.io.Importer E /etc/acs-commons/manage-controlled-processes (javax.jcr.nodetype.NoSuchNodeTypeException: Node type cq:Page does not exist)
13.09.2017 12:06:59.750 *ERROR* [OsgiInstallerImpl] org.apache.jackrabbit.vault.fs.io.Importer E /etc/clientlibs/acs-commons/vendor/qrcode (javax.jcr.nodetype.NoSuchNodeTypeException: Node type cq:ClientLibraryFolder does not exist)

A dependency should be set on cq-content to make sure that the bundle is installed in the right order

royteeuwen:
Add a dependency to the cq-content bundle as to make sure the primary… types are available before actually installing this bundle

fixes #1110

### DIFF
--- a/content/pom.xml
+++ b/content/pom.xml
@@ -68,6 +68,13 @@
                             <target>/apps/acs-commons/install</target>
                         </embedded>
                     </embeddeds>
+                    <dependencies>
+                        <dependency>
+                            <group>day/cq60/product</group>
+                            <name>cq-content</name>
+                            <version>[6.3.64,)</version>
+                        </dependency>
+                    </dependencies>
                     <targetURL>http://${crx.host}:${crx.port}/crx/packmgr/service.jsp</targetURL>
                 </configuration>
             </plugin>


justinedelson:
This would limit installation to 6.3 and above. I don't think that's desired.

@davidjgonzalez are we good to limit installation to 6.2 and above?

royteeuwen:
Perfect, would have gladly done it myself @justinedelson ! I was just starting up an AEM 6.2 to see which version was needed :)

PR #1121, opened 2017-09-26
MCP task to report broken references 

YegorKozlov:
Reporting broken references is a common request which periodically pops up in the mailing lists. This patch provides a MCP task for that. 
The code walks the specified content tree, finds properties that reference other resources and checks for their existence.  The checked references can be one of:

1. A string property containing a reference, e.g, fileReference=/content/dam/image.png
2. An array of strings, e.g, fileReference=[/content/dam/image1.png, /content/dam/image2.png]. Al lof the array elements will be validated
3. A property containing html with links , e.g, 
```
<p>
<a href='/content/page.html'>hello</a>
<img src='/content/dam/image1.png'/>
</p>
```
deep check in html is on by default, but can be disabled on the report page

The report contains the source path and the value of the broken reference:

image

I also tweaked GenericReportExcelServlet to apply auto-filter and column auto-sizing. This way the generated spreadsheet  looks much better.


davidjgonzalez:
@YegorKozlov good to have you back :)

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/BrokenLinksReport.java
```diff
@@ -0,0 +1,211 @@
...
+        report.persist(rr, instance.getPath() + "/jcr:content/report");
+
+    }
+
+    public void buildReport(ActionManager manager) {
+        TreeFilteringResourceVisitor visitor = new TreeFilteringResourceVisitor();
+        visitor.setBreadthFirstMode();
+        visitor.setTraversalFilter(null);
+        visitor.setResourceVisitor((resource, depth) -> {
+            Map<String, List<String>> brokenRefs = collectBrokenReferences(resource, regex, excludeList, deepCheckList);
```
badvision: This part should be wrapped by manager.deferredWithResolver as well, this is how you can divide the work and really speed things up in parallel.  The first part being deferred is mostly so that you benefit from error handling more than anything else.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/BrokenLinksReport.java
```diff
@@ -0,0 +1,211 @@
...
+                            List<String> brokenPaths =  collectPaths(entry, htmlFields)
+                                    .filter(href -> regex.matcher(href).matches())
+                                    .filter(path -> ResourceUtil.isNonExistingResource(resource.getResourceResolver().resolve(path)))
+                                    .collect(Collectors.toList());
+                            return brokenPaths;
+                        })).entrySet().stream().filter(e -> !e.getValue().isEmpty())
+                .collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue()));
+     }
+        // access from unit tests
+    Map<String, EnumMap<REPORT, Object>> getReportData() {
```
badvision: Needs to be a synchronized map otherwise race conditions will cause concurrent modification exceptions

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/BrokenLinksReport.java
YegorKozlov: A good catch. With this block of code wrapped in manager.deferredWithResolver the number of completed tasks is proportional to the number of visited ValueMaps. Each visit is a unit of work and can run in a pooled thread. Previously all the processing was single-threaded.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/BrokenLinksReport.java
YegorKozlov: OK, I changed TreeMap to ConcurrentHashMap.  This way the pooled threads can safely put into it.

bundle/pom.xml
```diff
@@ -281,16 +282,25 @@
         <dependency>
             <groupId>org.apache.poi</groupId>
             <artifactId>poi-ooxml</artifactId>
-            <version>3.6</version>
+            <version>3.16</version>
```
justinedelson: is this compatible with AEM 6.2? As far as I can tell, the POI version shipped with 6.2 was 3.11.

bundle/pom.xml
```diff
@@ -59,6 +59,7 @@
                             com.day.cq.wcm.workflow.process;version="[6.0,8)", <!-- using a wider version range for forward compatibility -->
                             com.day.cq.mailer;version="[5.9,7)", <!-- using a wider version range for forward compatibility -->
                             com.adobe.cq.sightly;version="[2.5,4)",
+                            javax.annotation;resolution:=optional, <!-- pulled by the tika-parsers dependency -->
```
justinedelson: Hmmm. Why does adding a dependency to tika-parsers require a change to the Import-Package configuration here? That doesn't sound right. tika isn't embedded, right?

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/GenericReportExcelServlet.java
```diff
@@ -17,23 +17,25 @@
...
-import java.util.List;
-import javax.servlet.ServletException;
-import javax.servlet.ServletOutputStream;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.felix.scr.annotations.sling.SlingServlet;
-import org.apache.poi.ss.usermodel.Workbook;
-import org.apache.poi.xssf.usermodel.XSSFCell;
-import org.apache.poi.xssf.usermodel.XSSFRow;
-import org.apache.poi.xssf.usermodel.XSSFSheet;
+import org.apache.poi.ss.usermodel.*;
```
justinedelson: please avoid the use of wildcard imports.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/BrokenLinksReport.java
```diff
@@ -0,0 +1,211 @@
...
+import java.io.Serializable;
+import java.util.*;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Broken Links Checker MCP task
+ *
+ * @author Yegor Kozlov
```
justinedelson: Please remove. We don't use `@author` tags in this project.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/BrokenLinksReport.java
```diff
@@ -0,0 +1,211 @@
...
+import org.apache.tika.metadata.Metadata;
+import org.apache.tika.parser.ParseContext;
+import org.apache.tika.parser.html.HtmlParser;
+import org.apache.tika.sax.Link;
+import org.apache.tika.sax.LinkContentHandler;
+
+import javax.jcr.RepositoryException;
+import java.io.ByteArrayInputStream;
+import java.io.Serializable;
+import java.util.*;
```
justinedelson: Please avoid the use of wildcard imports.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/BrokenLinksReport.java
```diff
@@ -0,0 +1,211 @@
...
+    static Map<String, List<String>> collectBrokenReferences(Resource resource, Pattern regex, Set<String> skipList, Set<String> htmlFields) {
+
+        return resource.getValueMap().entrySet().stream()
+                .filter(entry -> !skipList.contains(entry.getKey()))
+                .collect(Collectors.toMap(
+                        entry -> resource.getPath() + "/" + entry.getKey(),
+                        entry -> {
+                            List<String> brokenPaths =  collectPaths(entry, htmlFields)
+                                    .filter(href -> regex.matcher(href).matches())
+                                    .filter(path -> ResourceUtil.isNonExistingResource(resource.getResourceResolver().resolve(path)))
```
justinedelson: Is there actually a need to use `ResourceResolver.resolve()` here vs. `ResourceResolver.getResource()`? I would expect that links in content are to the actual resource path, not the mapped path.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/BrokenLinksReport.java
```diff
@@ -0,0 +1,211 @@
...
+
+    @FormField(
+            name = "Deep check in html",
+            description = "If checked, links will be extracted from html field",
+            component = CheckboxComponent.class,
+            options = {"checked"}
+    )
+    private boolean deepCheck = false;
+
+    @FormField(name = "Fields containing html",
```
justinedelson: The name here should probably be "Properties containing HTML", not "Fields".

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/BrokenLinksReport.java
```diff
@@ -0,0 +1,211 @@
...
+            description = "Properties containing html to extract links",
```
justinedelson: Comma-separate list, right?

bundle/pom.xml
YegorKozlov: AEM 6.3 ships with 3.15 and for my project Adobe provided an update with 3.16.  I can downgrade it to 3.11, it shouldn't be a problem. 
Compatibility shouldn't be a problem. I'm telling it with my Apache POI committer hat on.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/BrokenLinksReport.java
YegorKozlov: I wrote this tool to detect potential 404s which can come from mapped paths as well. For completeness sake I would like to keep ResourceResolver.resolve()

bundle/pom.xml
justinedelson: Ah, I see the problem now. In 6.2 and 6.3, this package is exported by the system bundle with a minimal version (`0.0.0.1_008_JavaSE` or something like that). Currently, ACS AEM Commons imports this package with no version (so an effective version of 0 to infinity because we get it from a non-OSGI jar. When Tika brings in `javax.annotation:javax.annotation-api`, that jar has OSGi metadata which causes the range to be `[1.2,2)`.

Rather than using `exclusions` I think the correct approach would be to set the scope as `provided`. This should be done for all dependencies, but it looks like @badvision forgot to do this for POI. I'll fix that in master.

justinedelson:
Thanks!

PR #1133, opened 2017-10-03
initial implementation of an S3 ingestor for MCP

justinedelson:
includes some internal refactoring to avoid code duplication between S3 and File ingestor, although some still exists


justinedelson:
Here's the problem with the StringUtils class... S3 Mock depends upon a method added in the 1.10.77 release of the AWS SDK. Naturally, AEM uses 1.10.76. So with just the AEM version on the classpath, S3 Mock fails miserably. So I took the latest copy of the StringUtils class and incorporated in into the codebase (as a test file) so that it would get picked up at test time. Would it have been better if S3 Mock used Commons StringUtils? sure. But I don't think that's likely to change.

justinedelson:
@badvision @davidjgonzalez I think this is now ready to merge.

PR #1138, opened 2017-10-15
New framework: JcrPersist - non-instrusive ORM for JCR

sangupta:
Hello - I am raising this PR to add a new framework/feature and to get a discussion started to see if-and-how it fits the world of the `AEM Commons` project. Below are more details on the framework. Please feel free to review/comment/ask-questions.

The code below is currently devoid of unit-tests, however, the code as part of this PR is 100% fully functional and tested using many support-classes. Am already working to re-create them into proper unit-tests, but would like to hear the feedback before completing that to ensure that any changes to this can be incorporated.

## JcrPersist

`JcrPersist` is a simple no-intrusion ORM style framework for JCR/CRX. It allows you to read/write
simple plain Java POJO on the JCR using Java reflection techniques. It needs no annotations
to start with and provides multiple custom extension points for usage.

All functionality of **JcrPersist** that can be used via annotations, can also be used programmatically using the **JcrPersist** APIs. This allows one to work with POJO objects whose source cannot be modified, such as from 3rd party frameworks like Twitter, Instagram, Facebook SDKs and all.

Following are the primary differences from **Sling Models**:

* It is 100% non-intrusive - no need to modify your POJO objects (extend from base-class/add-annotations) to work with
* There are no build time steps to be made (no POM changes)
* Any POJO can be used to both read/write objects
* Collections (including implicit collections) are supported
* Nested and composed objects are supported as well
* All customizations can be done either by use of annotations (intrusive) or programmatically via `JcrPerist` APIs (non-intrusive).

A simplest hello-world kind of example is as follows. Consider the following 2 value-objects:

```java
public class Stocks {

	public long lastUpdated;
	
	public List<Stock> children;

}

public class Stock {

	public String name;
	
	public double price;

}
``` 

To read an object from the JCR,

```java
Stocks stocks = JcrPersist.read("/content/mysite/stocks", Stocks.class, resourceResolver);
```

To write the same updated `Stocks` instance back to JCR:

```java
JcrPersist.persist("/content/mysite/stocks", stocks, resourceResolver);
``` 

More details can be found in the [docs/jcrpersist.md](https://github.com/sangupta/acs-aem-commons/blob/feature/jcr-persist/docs/jcr-persist.md) file (which is a part of this PR).

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/jcrpersist/JcrPersist.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/jcrpersist/JcrPersist.java
@@ -51,8 +51,6 @@ import com.day.cq.wcm.api.Page;
  * using the JCR without any intrusion. This also allows the system
  * to be more extensible.
  * 
- * @author sangupta
- *
  */
 public class JcrPersist {
 	
--- a/bundle/src/main/java/com/adobe/acs/commons/jcrpersist/JcrPersist.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/jcrpersist/JcrPersist.java
@@ -1,0 +1,264 @@
+package com.adobe.acs.commons.jcrpersist;
+
+import java.lang.reflect.Field;
+import java.math.BigDecimal;
+import java.net.URI;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import javax.jcr.RepositoryException;
+
+import org.apache.sling.api.resource.PersistenceException;
+import org.apache.sling.api.resource.Resource;
+import org.apache.sling.api.resource.ResourceResolver;
+import org.apache.sling.api.resource.ValueMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.adobe.acs.commons.jcrpersist.annotation.AEMProperty;
+import com.adobe.acs.commons.jcrpersist.annotation.AEMPropertyExclude;
+import com.adobe.acs.commons.jcrpersist.extension.ValueMapReader;
+import com.adobe.acs.commons.jcrpersist.util.ReflectionUtils;
+import com.day.cq.commons.jcr.JcrConstants;
+import com.day.cq.wcm.api.Page;
+
+/**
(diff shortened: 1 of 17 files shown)


justinedelson:
@sangupta do you intend to resolve the coverage issues?

justinedelson:
Some general comments:

* Please review the API exposed. Based on the documentation, there is much more being exported here than necessary.
* Missing package-info files for the exported packages.
* Missing license headers
* Should not include `@author` tag.
* Please check formatting. I see some inconsistent indentation. The project generally uses spaces instead of tabs, although I see that is not something which is formally documented. Regardless of spaces vs. tabs, indentation should be consistent in a single file. Take a look at the ReflectionUtils.java file specifically.
* The AssertUtils class seems to duplicate functionality which already exists in commons-lang. Also a little odd to use the term "assert" here when that generally has a different meaning elsewhere in Java.
* It is fine to have the documentation in this project to start, but eventually it needs to be moved over to the actual website.

sangupta:
@justinedelson Am already working on adding unit test cases for everything in this PR.

* My thoughts behind exposing everything as an API or as an annotation were to make this as non-intrusive as possible. Help guide me to what is extraneous and I shall remove it.
* Will add package-info files
* Will add license headers
* Will remove `@author` tags
* Will format using white-spaces, looks like my Eclipse settings were broken
* Will remove `AssertUtils` and use `commons-lang`
* Help me with the process on how to get documentation for a feature in - is there a separate repo that I need to raise PR against - and how do we keep both in sync.

Thanks for taking time to review.

justinedelson:
Regarding the API, it looks to me like the only classes which need to be exported are `JcrPersist` and the annotations. Everything else appears to be only used internally and so shouldn't be in an exported package. I guess having the package-private classes in an exported package is OK, although that does create a bit of a inconvenience when generating the JavaDoc since they will be included (I think) but can't actually be used.

sangupta:
@justinedelson If I understand correctly, its to fix the OSGi bundle information with right classes in the exported packages, right?

justinedelson:
It's really two things:
* Put only the classes you want to be the API in the exported package. Everything else needs to go in a package named `impl`. There are other ways to do this (both by convention and configuration), but this is the way this project is set up.
* Create a `package-info.java` file in the exported package(s) so that the package version is properly managed.

PR #1142, opened 2017-10-18
Make sure report tabulation is thread-safe

badvision:
Added synchronization to the tabulation method to avoid race conditions

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/AssetReport.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/AssetReport.java
@@ -197,7 +197,7 @@ public class AssetReport extends ProcessDefinition implements Serializable {
         }
     }
 
-    private void tabulate(String path, Column counter, long amount) {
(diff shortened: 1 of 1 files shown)


bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/AssetReport.java
```diff
@@ -197,7 +197,7 @@ private void setValue(String path, Column field, long amount) {
         }
     }
 
-    private void tabulate(String path, Column counter, long amount) {
+    private synchronized void tabulate(String path, Column counter, long amount) {
         if (getDepth(path) < depthLimit && path.length() >= baseFolder.length()) {
             synchronized (report) {
```
justinedelson: is this still necessary if the whole method is synchronized?

justinedelson:
@badvision if you want to disable stalebot, you can add the pinned label (see https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/.github/stale.yml)

Although I'm not 100% sure why this just hasn't been merged...

PR #1145, opened 2017-10-20
New Feature: Redirect Map Manager

klcodanr:
This is based on some concepts we've been exploring at a client. Rather than using vanities which have several limitations including:

- Performance
- Maintainability (e.g. how do I see all the vanities)
- Cross-site / domain capabilities

Or httpd Redirect Map text files, which have limitations including:

- Requiring apache restarts to update
- Requiring CAB / infrastructure support to update
- Performance for very large files

We instead used properties in AEM to produce a redirect map and then have a script to download the file to an Apache Redirect Map DBM file as such:

`wget http://{SERVER}:4503/etc/acs-commons/redirect-map/{instance}.redirectmap.txt -O /tmp/redirectmap.txt >> /var/log/update-redirect-map.log 2>&1`
`httxt2dbm -i /tmp/redirectmap.txt -o /etc/httpd/conf/redirectmap.map >> /var/log/update-redirect-map.log 2>&1`

Finally, we can reference this file in our Apache configurations:

`# Rewrite rules`
`RewriteMap map.legacy dbm:/etc/httpd/conf/redirectmap.map`
`RewriteMap lc int:tolower`
`RewriteCond ${map.legacy:${lc:$1}} !=""`
`RewriteRule ^(.*)$ 		${map.legacy:${lc:$1}|/} [L,R=301]`

This has the following benefits:

- Since we can create multiple of these configurations, we can manage the redirects / vanities for different sites through a single mechanism
- Since it uses a DBM instead of TXT configuration, the redirect maps can be updated as quickly as activating a file
- We can specify using the same mechanism:
  - Vanity (same site)
  - Legacy (old site -> AEM)
  - External Redirects 
- Redirect maps can be previewed prior to activating the files / pages
- AEM entries containing whitespace (which cause issues in Redirect Maps) are excluded and displayed in a warning in preview


justinedelson:
@klcodanr this looks very cool. is it possible to increase the coverage rate?

klcodanr:
@justinedelson Updated to increase code coverage. Lots of mocks ;-)

klcodanr:
Hey @davidjgonzalez and @justinedelson any updates on this PR?

justinedelson:
sorry @klcodanr . been a busy few weeks. It's on my todo list...

klcodanr:
Know the feeling @justinedelson, thanks!

justinedelson:
@klcodanr sorry for the delay. This looks pretty good to me. I did find it a little non-intuitive at first. WDYT about making the following changes

1. Add some text above the upload button explaining the file format. Per https://httpd.apache.org/docs/current/rewrite/rewritemap.html#txt, this could be something as simple as "This file should be a space-delimited file with the first column containing the source path and the second column containing the redirect destination."
2. Add some explanatory text to what the path and property fields do ("The property and path fields are used to form a query to find available redirects. For example, specifying / as the path and sling:vanityPath as the property will load all vanity paths in the system.")
3. Add a copyable link to the redirect file itself. Maybe even use the externalizer to create the publish URL.

justinedelson:
Thanks!

justinedelson:
@klcodanr now that this has been released, can you submit a PR for https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io with the documentation? thx.

klcodanr:
Awesome will do! Probably tomorrow.

klcodanr:
@justinedelson provided in: https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io/pull/108

Issue #1150, opened 2017-10-24
Empty date/time value of the composite multifield will show as "Invalid date" in the touchui dialog


victor994:
The empty value of datetime field from the composite multifield will show as "invalid date" in the touchui dialog.

PR #1151, opened 2017-10-24
Issue #1150: Attempt to fix the invalid date in the dialog for the empty datetime value

victor994:
The issue is that if the datetime field has an empty value, moment format method will return "Invalid date". The fix is to check first if the date object is valid. If invalid, it should return value as is without modifying.


justinedelson:
Thanks!

victor994:
no problems. can you please advise when the next release will be?

PR #1153, opened 2017-10-24
remove current page from list of referenced pages. fixes #1143

### LINKED ISSUE #1143: Page will not create version on activation if it contains link to itself
We have a content page at a level in node structure which contains a nav component. This nav component has a link to itself, so sub-pages can use the nav content for their own nav. 

When the page is modified and activated, the reference provider shows the page as outdated and allows user to include as part of activation. When this is selected, the jcr:content for the page is activated, but then the subsequent page activation doesn't see it as modified, and no version is created for the activation. 

Ideally, the reference provider class can check if it's itself, and not prompt as a reference.

justinedelson:


justinedelson:
@davidjgonzalez the actual fix is pretty trivial, but I went ahead and rewrote the test to use Sling Mocks.

PR #1155, opened 2017-10-25
New Feature: A granite UI (touch UI) widget for editing HTML

ahmed-musallam:
This is a simple granite UI widget, for Touch UI dialogs, used for editing HTML with code highlighting using [CodeFlask](https://github.com/kazzkiq/CodeFlask.js).

Also adds two vendor clientlibs: [Prism.js](http://prismjs.com/) (only HTML language) and [CodeFlask](https://github.com/kazzkiq/CodeFlask.js)


justinedelson:
TBH, this makes me a little nervous. While the widget in and of itself isn't a problem, it seems to encourage a pattern of allowing authors to hand-edit HTML which is known to create security risks, specifically XSS. @ahmed-musallam how do you forsee the XSS risk resulting from the usage of this widget being mitigated?

ahmed-musallam:
@justinedelson fair point, let me look into using sling's [XSSFilter](https://sling.apache.org/apidocs/sling9/org/apache/sling/xss/XSSFilter.html) with an appropriate context to remove potential XSS  risks. Alternatively, I can look into a JS validation solution to warn against XSS risks.

justinedelson:
generally, Adobe's best practice (and I think the industry best practice as well) is to filter on output, so it isn't really an issue for this widget per se. But can you verify that https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/1155/files#diff-f628067e555e0c13693083f40ba13facR52 doesn't require additional XSS filtering?

ahmed-musallam:
@justinedelson The JavaDoc for [AttrBuilder](https://docs.adobe.com/docs/en/aem/6-3/develop/ref/javadoc/com/adobe/granite/ui/components/AttrBuilder.html#add(java.lang.String,%20java.lang.String))  does mention that attributes are protected with XSSAPI.
> A builder to generate HTML attributes. This builder is designed to be secured using [XSSAPI](https://docs.adobe.com/docs/en/aem/6-3/develop/ref/javadoc/com/adobe/granite/xss/XSSAPI.html). It will encode the value automatically. If the value is null, it will be ignored.

However, I think it may be using `XSSAPI#encodeForHTMLAttr` which will only encode the html and protect against script-running attributes like `onload` or `onclick`, but not necessarily against an HTML string containing a `<script>` I have tested it, and it does not filter out `<script></script>` for example.

I don't think I should filter XSS issues for the widget since authors should see exactly what they authored in the dialog and no XSS will happen on author side since the code protects against it: encodes HTML in the attribute and does not add the HTML to the dialog DOM as-is.

Like you said, XSS prevention should be done on output, i.e. in the **component** that is rendering the HTML string to the page: with HTL display context or in JSP with sling XSSFilter, which is what is done today for the rich text editor (even though it has frontend XSS protection).

Let me know your thoughts.

justinedelson:
@ahmed-musallam I think as long as the widget doesn't display the *rendered* HTML, this will be OK. The problem isn't the text `<script>alert("owed");</script>`, it's what happens when the browser interprets as HTML :)

I want to think about it a bit more and make sure I'm not missing anything. Been a busy week but I'll try to get back to this by mid-next week.

PR #1157, opened 2017-10-30
defer loading of Granite.author to be as late as possible

justinedelson:
This replaces #1156


bkowalk:
Looks like a good replacement to me.  Testing shows it also fixes the image uploader issue.  Thanks for the quick turnaround!

justinedelson:
@schoudry would you be able to review this?

@paul-bjorkstrand would also be interested in your feedback.

schoudry:
@justinedelson looks good

PR #1161, opened 2017-11-02
class attribute fix composite multifield touchui

victor994:
Fixed the selector to allow custom class attribute such as
"section class="your-own-class coral-Form-fieldset" data-name="./data" data-acs-commons-nested="JSON_STORE"


justinedelson:
Thanks!

PR #1168, opened 2017-11-09
Feature/multifield datepicker value not loaded in touchui dialog

victor994:
Like richtext field, the datepicker field should load the name from the first input instead of the second. The second input name is for typehint.


justinedelson:
Thanks!

PR #1170, opened 2017-11-13
improve exception handling

joerghoh:
I improved the exception handling a bit
* do not log the exception separately, but just wrap and throw it
* include the payload in the exception.

Additional observation: I am not really sure, if we should close the resourceresolver. If I understand the ```getResourceResolver()```method correctly, it just creates a new resourceresolver based on an already existing session. Is it save to this close this resourceresolver or does it close the associated (JCR) session as well?


justinedelson:
It is a good idea to close out the `ResourceResolver` in this case since it is created in this context. The `JcrResourceProvider` will "remember" that the `ResourceResolver` was created with an existing `Session` object and not logout the session.

justinedelson:
Merged with some minor formatting changes

PR #1173, opened 2017-11-13
 Only publish coverage data for en_US build

justinedelson:


justinedelson:
seems to have worked.

Issue #1175, opened 2017-11-14
Add Permission Sensitive Cache Servlet


kaushalmall:
Currently, this code gets copied and pasted from one project to another and every developer seem to have their own version. Might be good to have one version in Commons that people can reuse. 

Idea is for this servlet to have two configurable properties 

- The servlet path
- The list of extensions to be removed from the URI so it resolves to a valid resource. 

Thanks @justinedelson for pointing this out.

justinedelson:
PR has been merged

PR #1177, opened 2017-11-16
Feature/Permission Sensitive Cache Servlet

kaushalmall:
@justinedelson @davidjgonzalez @bstopp this is for issue #1175 

Not sure its safe to merge though, @bstopp is having issues building the feature branch from my personal fork. Can either one of you confirm?

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/dispatcher/impl/PermissionSensitiveCacheServlet.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/dispatcher/impl/PermissionSensitiveCacheServlet.java
@@ -71,7 +71,7 @@ public class PermissionSensitiveCacheServlet extends SlingSafeMethodsServlet {
             }
 
         } catch(Exception e) {
-            log.error("authchecker servlet exception: " + e.getMessage());
+            log.error("Authchecker servlet exception", e);
             response.setStatus( SlingHttpServletResponse.SC_UNAUTHORIZED );
         }
     }
--- a/bundle/src/main/java/com/adobe/acs/commons/dispatcher/impl/PermissionSensitiveCacheServlet.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/dispatcher/impl/PermissionSensitiveCacheServlet.java
@@ -63,7 +63,7 @@ public class PermissionSensitiveCacheServlet extends SlingSafeMethodsServlet {
             Resource requestedResource = resourceResolver.resolve( request, requestUri );
 
             if( !ResourceUtil.isNonExistingResource( requestedResource ) ){
-                log.info("Current Session has access to {}", requestUri );
+                log.debug("Current Session has access to {}", requestUri );
                 response.setStatus(SlingHttpServletResponse.SC_OK);
             } else {
                 log.info("Current Session does not have access to {}", requestUri );
--- a/bundle/src/main/java/com/adobe/acs/commons/dispatcher/impl/PermissionSensitiveCacheServlet.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/dispatcher/impl/PermissionSensitiveCacheServlet.java
@@ -1,0 +1,121 @@
+/*
+ * #%L
+ * ACS AEM Commons Bundle
+ * %%
+ * Copyright (C) 2017 - Adobe
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+package com.adobe.acs.commons.dispatcher.impl;
+
+import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.felix.scr.annotations.*;
+import org.apache.sling.api.SlingHttpServletRequest;
+import org.apache.sling.api.SlingHttpServletResponse;
+import org.apache.sling.api.resource.Resource;
+import org.apache.sling.api.servlets.SlingSafeMethodsServlet;
+import org.apache.sling.commons.osgi.PropertiesUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+import static com.adobe.acs.commons.dispatcher.impl.PermissionSensitiveCacheServlet.REMOVEABLE_EXTENSIONS;
+
+@Component(
+        label = "ACS AEM Commons - Permission Sensitive Cache Servlet",
+        description = "Servlet that checks if the current sessions has access to a cached object",
+        metatype = true, immediate = true
+)
+@Properties({
+        @Property(
+                label = "File extensions to be removed from the URI",
+                name = REMOVEABLE_EXTENSIONS,
+                description = "Extensions that can be removed from the URI. For example, .html",
+                cardinality = Integer.MAX_VALUE,
+                value = {"html"}
+        ),
+        @Property(
+                name = "sling.servlet.paths",
+                cardinality = Integer.MAX_VALUE,
+                label = "Sling Servlet Paths",
+                description = "Paths that this servlet will resolve to"
+        )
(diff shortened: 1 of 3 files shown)


bundle/src/main/java/com/adobe/acs/commons/dispatcher/impl/PermissionSensitiveCacheServlet.java
```diff
@@ -0,0 +1,121 @@
...
+            String requestUri = request.getParameter( "uri" );
+            log.debug( "Request URI {}", requestUri );
+
+            if(StringUtils.isNotEmpty( requestUri ) ){
+
+                String uri = getUri( requestUri );
+
+                log.info( "Checking permissions for {}", uri );
+
+                Resource requestedResource = request.getResourceResolver().getResource( uri );
```
davidjgonzalez: Would it be better to use `rr.resolve(request, uri)` ? It seems like this would simplify this servlet and handle other corner cases:

* selectors
* use of /etc/map
* vanities (cant recall if getResource() will work for them, but resolve def does)

It comes at the cost of a single call to resolve (which seems like a reasonable trade-off).

bundle/src/main/java/com/adobe/acs/commons/dispatcher/impl/PermissionSensitiveCacheServlet.java
```diff
@@ -0,0 +1,121 @@
...
+                response.setStatus(SlingHttpServletResponse.SC_UNAUTHORIZED);
+            }
+        } catch(Exception e){
+            log.error("authchecker servlet exception: " + e.getMessage());
+        }
+    }
+
+    public String getUri(String requestUri){
+
+        String extension = FilenameUtils.getExtension( requestUri );
```
davidjgonzalez: Using `FilenameUtils` feels wrong in that the param is  URI and not a file... If we cant just user` rr.resolve(..)` (described above) which makes this trimming moot, how about using `PathInfo` to parse the URI at which point you have direct access to the resource path (`pathInfo.getResourcePath()`) ... 

https://docs.adobe.com/docs/en/aem/6-3/develop/ref/javadoc/com/day/cq/commons/PathInfo.html

bundle/src/main/java/com/adobe/acs/commons/dispatcher/impl/PermissionSensitiveCacheServlet.java
```diff
@@ -0,0 +1,121 @@
...
+
+            if(StringUtils.isNotEmpty( requestUri ) ){
+
+                String uri = getUri( requestUri );
+
+                log.info( "Checking permissions for {}", uri );
+
+                Resource requestedResource = request.getResourceResolver().getResource( uri );
+
+                if (requestedResource != null) {
```
davidjgonzalez: If using resolve, i believe this would need a "isNonExistingResource()" check.

bundle/src/main/java/com/adobe/acs/commons/dispatcher/impl/PermissionSensitiveCacheServlet.java
```diff
@@ -0,0 +1,121 @@
...
+
+    public void doHead(SlingHttpServletRequest request, SlingHttpServletResponse response) {
+        doHeadRequest(request, response);
+    }
+
+    public void doHeadRequest(SlingHttpServletRequest request, SlingHttpServletResponse response) {
+        try{
+            //retrieve the requested URL
+            String requestUri = request.getParameter( "uri" );
+            log.debug( "Request URI {}", requestUri );
```
davidjgonzalez: Feels more like a debug or trace; ie "hi - im being handled by this servlet"

bundle/src/main/java/com/adobe/acs/commons/dispatcher/impl/PermissionSensitiveCacheServlet.java
```diff
@@ -0,0 +1,121 @@
...
+                    response.setStatus(SlingHttpServletResponse.SC_OK);
+                } else {
+                    log.debug("Current Session does not have access to {}", uri );
+                    response.setStatus(SlingHttpServletResponse.SC_UNAUTHORIZED);
+                }
+            } else {
+                response.setStatus(SlingHttpServletResponse.SC_UNAUTHORIZED);
+            }
+        } catch(Exception e){
+            log.error("authchecker servlet exception: " + e.getMessage());
```
davidjgonzalez: Im not sure what the expected behavior is if there is an exception, but its going to return a 200 if this catch block is hit.

bundle/src/main/java/com/adobe/acs/commons/dispatcher/impl/PermissionSensitiveCacheServlet.java
```diff
@@ -0,0 +1,121 @@
...
+
+                log.info( "Checking permissions for {}", uri );
+
+                Resource requestedResource = request.getResourceResolver().getResource( uri );
+
+                if (requestedResource != null) {
+                    log.debug("Current Session has access to {}", uri );
+                    response.setStatus(SlingHttpServletResponse.SC_OK);
+                } else {
+                    log.debug("Current Session does not have access to {}", uri );
```
davidjgonzalez: See above comment wrt to INFO, this feels more like INFO than a DEBUG; Or maybe this is DEBUG and above is TRACE.

kaushalmall:
@davidjgonzalez did all the changes, take a look and let me know what you think. Thanks for the optimization ideas.

bundle/src/main/java/com/adobe/acs/commons/dispatcher/impl/PermissionSensitiveCacheServlet.java
```diff
@@ -0,0 +1,78 @@
...
+            //retrieve the requested URL
+            ResourceResolver resourceResolver = request.getResourceResolver();
+            String requestUri = request.getParameter( "uri" );
+
+            log.debug( "Checking access for URI {}", requestUri );
+
+            Resource requestedResource = resourceResolver.resolve( request, requestUri );
+
+            if( !ResourceUtil.isNonExistingResource( requestedResource ) ){
+                log.info("Current Session has access to {}", requestUri );
```
justinedelson: I know @davidjgonzalez suggested to change this from debug to info, but I'm questioning that suggestion :)

Doesn't the auth checker servlet get called on every request to dispatcher? If so, this will be quite noisy.

kaushalmall: `Doesn't the auth checker servlet get called on every request to dispatcher? If so, this will be quite noisy.`
Yes it does

justinedelson: maybe split the difference and log at debug for allowed and info for denied?

kaushalmall: done.

bundle/src/main/java/com/adobe/acs/commons/dispatcher/impl/PermissionSensitiveCacheServlet.java
```diff
@@ -0,0 +1,78 @@
...
+    public void doHead(SlingHttpServletRequest request, SlingHttpServletResponse response) {
+        try{
+
+            //retrieve the requested URL
+            ResourceResolver resourceResolver = request.getResourceResolver();
+            String requestUri = request.getParameter( "uri" );
+
+            log.debug( "Checking access for URI {}", requestUri );
+
+            Resource requestedResource = resourceResolver.resolve( request, requestUri );
```
davidjgonzalez: @kaushalmall what happens when requestUri is null? Nothing bad right?

kaushalmall: @davidjgonzalez based on [0]

absPath - The absolute path to be resolved to a resource. `If this parameter is null, it is assumed to address the root of the resource tree`. If the path is relative it is assumed relative to the root, that is a slash is prepended to the path before resolving it.

Assuming this will be ok. I can put a null check too, if that's a better way of handling it. 

[0]https://sling.apache.org/apidocs/sling7/org/apache/sling/api/resource/ResourceResolver.html#resolve-javax.servlet.http.HttpServletRequest-java.lang.String-

davidjgonzalez: makes sense to me; easy enough check

kaushalmall: done. @davidjgonzalez

kaushalmall:
@justinedelson @davidjgonzalez FYI, submitted a documentation PR for this.

Issue #1174, opened 2017-11-14
Add codeclimate


justinedelson:
We really should be doing codeclimate scans for pull requests. Of course, first we need to dramatically bring down the number of problems in our codebase. I've started to work on this and will be occasionally committing work in the issue/codeclimate branch.

davidjgonzalez:
Fixed via #1178

PR #1178, opened 2017-11-16
Add Codeclimate configuration and resolve all Java issues

justinedelson:
As mentioned in #1174, I would like to enable codeclimate on this repository. This commit adds the configuration necessary and resolves all of the identified Java issues so we start with a clean slate. I'll deal with JS and CSS/LESS in subsequent PRs -- those analyzers are currently disabled (and I'll probably switch from csslint to stylelint to get LESS support).

No API changes are involved. I did use a lot of warning suppression. In some cases, this was needed to keep the APIs backwards compatible. In other cases, it was for safety.


justinedelson:
Note - when this is merged, we should squash the commits. No need for my individual "down to XXX" commits to muddy the history

Issue #1166, opened 2017-11-06
Clientlib WARN messages for ACS AEM Commons clientlibs


justinedelson:
In testing #1165, I see the following WARN messages on a clean 6.3 instance + ACS AEM Commons 3.11.0:

    06.11.2017 13:03:03.869 *WARN* [sling-oak-observation-27] com.adobe.granite.ui.clientlibs.impl.FileBundle Referenced path in /etc/clientlibs/acs-commons/angularCoral/js.txt does not exist: /etc/clientlibs/acs-commons/angularCoral/js/select.js
    06.11.2017 13:03:03.902 *WARN* [sling-oak-observation-27] com.adobe.granite.ui.clientlibs.impl.FileBundle Referenced path in /etc/clientlibs/acs-commons/coralui-wrapped/css.txt does not exist: /etc/clientlibs/granite/coralui/less/cui-wrapped.less
    06.11.2017 13:03:14.208 *WARN* [0:0:0:0:0:0:0:1 [1509991394186] GET /content/we-retail/us/en.html HTTP/1.1] com.adobe.granite.ui.clientlibs.impl.LibraryCacheImpl Library /apps/acs-commons/components/utilities/errorpagehandler/preview/clientlibs specified embedded that could not be resolved: coralui
    06.11.2017 13:03:14.208 *WARN* [0:0:0:0:0:0:0:1 [1509991394186] GET /content/we-retail/us/en.html HTTP/1.1] com.adobe.granite.ui.clientlibs.impl.LibraryCacheImpl Library /apps/acs-commons/components/utilities/exporters/users-to-csv/clientlibs specified dependencies that could not be resolved: acs-commons.angularjs.v1.3
    06.11.2017 13:03:14.208 *WARN* [0:0:0:0:0:0:0:1 [1509991394186] GET /content/we-retail/us/en.html HTTP/1.1] com.adobe.granite.ui.clientlibs.impl.LibraryCacheImpl Library /apps/acs-commons/components/utilities/quickly/clientlibs specified embedded that could not be resolved: acs-commons.json2

justinedelson:
@davidjgonzalez can you take a look at this?

davidjgonzalez:
Sure thing

PR #1180, opened 2017-11-17
#1166 - Fixed issues w/ clientlibs (and broken Quickly filter)

davidjgonzalez:


justinedelson:
can you fix the thing codeclimate is complaining about?

Issue #1176, opened 2017-11-15
Composite MultiField in case of NodeStore cannot restore value from deep property


dzmitry-kankalovich:
Composite MultiField works fine with root node specified like ```./items```. However, if you specify it like ```./something/items``` it will be able to save values in that structure, but won't be able to restore them once dialog opened again.

The problem is in this line https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/acs-aem-commons-3.12.0/content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/composite-multifield/source/touchui-composite-multifield-nodestore.js#L183

Here ```nName``` would be ```something/items``` which cannot be resolved as JSON object property. Finding that deep property in ```data``` object seem to be solving problem.

PR #1181, opened 2017-11-17
Fixes 1176

schoudry:
@justinedelson @davidjgonzalez fixed 1176, nested property storage, composite fields with property something like ./acs-commons/users


justinedelson:
linking to #1176

Issue #1154, opened 2017-10-25
Select in Nested Multi-field is not stored correctly


bearcherian:
When a nested multi-field has a select widget, and the values are stored as JSON, the value is not stored correctly. Up to version 3.9.0, this worked fine. Starting in 3.10.0, this now stores the value under the key using the `*@Delete` name instead of the actual property name. 

For example, in the structure below, the JSON String should have a property `selectProperty`, but in 3.10.0 and 3.11.0 the value is stored as `selectProperty@Delete`

Sample widget structure:
```
<myMultifield
    jcr:primaryType="nt:unstructured"
    sling:resourceType="granite/ui/components/foundation/form/multifield"
    class="groupOfLinks-multifield"
    expand="{Boolean}true"
    fieldLabel="My Multifield">
    <field
        jcr:primaryType="nt:unstructured"
        sling:resourceType="granite/ui/components/foundation/form/fieldset"
        acs-commons-nested="JSON_STORE"
        name="./myMultifield">
        <layout
            jcr:primaryType="nt:unstructured"
            sling:resourceType="granite/ui/components/foundation/layouts/fixedcolumns"
            method="absolute"/>
        <items jcr:primaryType="nt:unstructured">
            <column
                jcr:primaryType="nt:unstructured"
                sling:resourceType="granite/ui/components/foundation/container">
                <items jcr:primaryType="nt:unstructured">
                    <linkStyle
                        jcr:primaryType="nt:unstructured"
                        sling:resourceType="granite/ui/components/foundation/form/select"
                        name="./selectProperty">
                        <items jcr:primaryType="nt:unstructured">
                            <option1
                                jcr:primaryType="nt:unstructured"
                                text="Option 1"
                                value="option1"/>
                            <option2
                                jcr:primaryType="nt:unstructured"
                                text="Option 2"
                                value="option2"/>    
                            <option3
                                jcr:primaryType="nt:unstructured"
                                text="Option 3"
                                value="option3"/>
                        </items>
                    </linkStyle>
                </items>
            </column>
        </items>
    </field>
</myMultifield>
```

justinedelson:
Can you attach a sample component which illustrates this behavior?

bearcherian:
Do you want a content package with the above dialog snippet, or just the XML content of the full dialog?

justinedelson:
A content package which contains a component which has this in the dialog.

tnebe:
[...]

PR #1182, opened 2017-11-17
Fixes 1154

schoudry:


srinm:
Hi @schoudry , thank you.
This fixed the issue with storing to jcr, Now i see no additional characters added 
but the dialog would not get populated on edit. Not sure if it has any dependency with version i'm using version 3.10.

schoudry:
@srinm please attach the component

srinm:
@schoudry  please find attached component code.
[evet-calendar-1.0.zip](https://github.com/Adobe-Consulting-Services/acs-aem-commons/files/1490709/evet-calendar-1.0.zip)

justinedelson:
@schoudry did you confirm that this fixes the issue with @srinm's component?

tearora:
I am also facing this issue  while using ACS commons 3.10.0 as dependency in the project.It is storing the values of Checkbox as <property_name>@Delete and value is also not retained in dialog edit.,

srinm:
@justinedelson , @schoudry This pull request has resolved my issue with Json store regarding storgae of additional string(@) to jcr.
However i also had another issue with retaining the values in dialog after applying this fix. The dialog issue was resolved after resolving the dependencies.  I have already sent an email regarding the same last week. Thank you again for your support and quick responses.

justinedelson:
@srinm I must have missed that email, but that's why we should try to communicate on this issue or #1154.

In any case, I don't really understand what you mean by "after resolving the dependencies". Are those dependencies of ACS AEM Commons or something else?

justinedelson:
I've tested this locally and I think the patch is working correctly. So merging.

I did note that the attached package embeds a copy of several files from ACS AEM Commons. I had to remove these in order to get the fix working. I'm not sure why those files were copied, but that would make any fix impossible in this specific case.

PR #1183, opened 2017-11-18
Use Java7 idiom 

joerghoh:
Sonar complained about the use of the workaround, which was suggested at SO for Java6. As we don't have the dependency on Java 6 anymore, get rid of it.

Additionally we should not swallow the exception, but just throw it.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/dam/audio/impl/FFMpegAudioUtils.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/dam/audio/impl/FFMpegAudioUtils.java
@@ -25,6 +25,8 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 
 public class FFMpegAudioUtils {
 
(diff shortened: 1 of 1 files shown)


davidjgonzalez:
@joerghoh thanks for the clean-up - could you resolve that conflict if you have a moment?

bundle/src/main/java/com/adobe/acs/commons/dam/audio/impl/FFMpegAudioUtils.java
```diff
@@ -65,18 +67,11 @@ public static File resolveWorkingDir(String slingHome, String path) {
...
-                throw new IOException("Unable to delete temp directory.");
-            }
-            if (!tempDir.mkdir()) {
-                throw new IOException("Unable to create temp directory.");
-            }
-        } catch (IOException e) {
-            log.warn("could not create temp directory in the [{}] with the exception", parentDir, e);
+        Path tempPath = Files.createTempDirectory(parentDir.toPath(),"cqdam");
+        File tmpDir = tempPath.toFile();
+        if (!tmpDir.mkdir()) {
```
justinedelson: I'm not sure why this is necessary -- the call to `Files.createTempDirectory` should actually create the temp directory.

justinedelson:
Doesn't seem necessary anymore.

Issue #1187, opened 2017-11-28
DynamicClassicUIClientLibraryServlet does not take into consideration context path


bdhoine:
when fetching the clientlibs the context path is not in the response

```
// 20171128124815
// http://*:4502/MB027200/bin/acs-commons/dynamic-classicui-clientlibs.json

{
  "js": [
    "/apps/acs-commons/widgets/configure-limit-parsys.min.js",
    "/apps/acs-commons/widgets/configure-parsys-placeholder.min.js"
  ],
  "css": [
    
  ]
}
```

so when the dynamic loader javascript inserts the scripts they can not be found

```
(function() {
    CQ.HTTP.get("/bin/acs-commons/dynamic-classicui-clientlibs.json", function(options, success, response) {
        var data, i, el,
            head = document.getElementsByTagName("head")[0];
        if (success) {
            data = JSON.parse(response.responseText);
            if (data.js) {
                for (i = 0; i < data.js.length; i++) {
                    el = document.createElement("script");
                    el.type = "text/javascript";
                    el.src = data.js[i];
                    head.appendChild(el);
                }
            }
            if (data.css) {
                for (i = 0; i < data.css.length; i++) {
                    el = document.createElement("link");
                    el.rel = "stylesheet";
                    el.type = "text/css";
                    el.href = data.css[i];
                    head.appendChild(el);
                }
            }
        }
    });
})();
```

justinedelson:
fixed with #1188

PR #1188, opened 2017-11-28
Add externaliser to DynamicClassicUiClientLibraryServlet #1187

Simsteve7:


bundle/src/main/java/com/adobe/acs/commons/wcm/impl/DynamicClassicUiClientLibraryServlet.java
```diff
@@ -83,7 +87,7 @@ protected void doGet(@Nonnull SlingHttpServletRequest request, @Nonnull SlingHtt
         if (!excludeAll) {
             Collection<ClientLibrary> libraries = htmlLibraryManager.getLibraries(categories, LibraryType.JS, true, true);
             for (ClientLibrary library : libraries) {
-                writer.value(resourceResolver.map(library.getIncludePath(LibraryType.JS, htmlLibraryManager.isMinifyEnabled())));
+                writer.value(externalizer.relativeLink(request, resourceResolver.map(library.getIncludePath(LibraryType.JS, htmlLibraryManager.isMinifyEnabled()))));
```
justinedelson: According to the JavaDoc, this is equivalent to passing the request into the `map()` method. Would you mind rewriting this as 

```
writer.value(resourceResolver.map(request, library.getIncludePath(LibraryType.JS, htmlLibraryManager.isMinifyEnabled())));
```
(and the same on the line below?)

Simsteve7: @justinedelson Thanks for reviewing. I updated it like requested.

justinedelson:
Thanks!

PR #1189, opened 2017-11-28
Remove references to oak.commons

justinedelson:
There a few references in our codebase to oak.commons. This is really not intended for use by clients of Oak. I also noticed that a couple of the ServiceUser tests weren't properly annotated so I fixed those.


justinedelson:
I don't think this is changelog worthy :)

PR #1192, opened 2017-11-29
HTTPCache: Added compatiblity for JSP includes, although this only su…

Sc0rpic0m:
Same as master branch request.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/httpcache/engine/impl/HttpCacheEngineImpl.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/httpcache/engine/impl/HttpCacheEngineImpl.java
@@ -398,10 +398,7 @@ public class HttpCacheEngineImpl extends AnnotatedStandardMBean implements HttpC
 
         // Copy the cached data into the servlet output stream.
         try {
-            IOUtils.copy(cacheContent.getInputDataStream(), response.getOutputStream());
-            if (log.isDebugEnabled()) {
-                log.debug("Response delivered from cache for the url [ {} ]", request.getRequestURI());
-            }
+            serveCacheContentIntoResponse(response, cacheContent);
 
             return true;
         } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


bundle/src/main/java/com/adobe/acs/commons/httpcache/engine/impl/HttpCacheEngineImpl.java
```diff
@@ -623,4 +620,23 @@ public TabularData getRegisteredPersistenceStores() throws OpenDataException {
...
+     * Has a fallback for wrappers that do not support the outputstream.
+     * @param response
+     * @param cacheContent
+     * @throws IOException
+     */
+    private void serveCacheContentIntoResponse(SlingHttpServletResponse response, CacheContent cacheContent)
+            throws IOException
+    {
+        try{
+            IOUtils.copy(cacheContent.getInputDataStream(), response.getOutputStream());
```
davidjgonzalez: Perhaps it would be better to check the `response.getContentType() => text/html` to see if the writer() can be used?

Feels weird using the exception to control flow of code.

@justinedelson thoughts on the best/definitive check for which to use?

Sc0rpic0m: Can we be 100% sure that if the response type is text/html, the print writer is supported?
If so I guess that's indeed the best.

justinedelson: I don't see why this would be specific to the case where the response type is `text/html`. To @davidjgonzalez 's question, although I agree in general that exceptions shouldn't be used for control flow, in this case I don't really see an alternative. The state of the response object (i.e. whether the output stream or writer has been returned) is not available.

justinedelson:
I think this is dependent upon #1010. Copying from an InputStream to a Writer without a character set is not a good idea (the copy method used has been deprecated in Commons IO 2.5)

Sc0rpic0m:
Should I do it then like : 

 IOUtils.copy(cacheContent.getInputDataStream(), response.getWriter(), response.getCharacterEncoding());   

?

justinedelson:
> IOUtils.copy(cacheContent.getInputDataStream(), response.getWriter(), response.getCharacterEncoding());

Possibly, although I thought there was a larger issue discussed in #1010 where the character set of the cached data wasn't being stored.

Sc0rpic0m:
Alright I adjusted it for now.

justinedelson:
merged with some formatting changes.

@Sc0rpic0m do you need a 6.0 compatible release for this? I'm hoping to release 3.13.0 soon with the change you submitted in #1191

Sc0rpic0m:
If the new release of 6.0 in the distant future would include this, it would be good for my current client so they can continue upgrading, but for now I have a custom version installed which includes this fix. Cheers!

justinedelson:
@Sc0rpic0m thanks. this will be in the next 2.x version, but if you don't need one right now, I'm going to focus on releasing the 3.13.0 version.

PR #1195, opened 2017-11-30
Catch IOExceptions during cleanup of test

badvision:


justinedelson:
just curious - is this the only case of this? I feel like this might be a more general problem.

PR #1196, opened 2017-12-01
Feature/issue 1193 property merge improvements

badvision:
This adds support for:

- Asset (TouchUI) metadata editor
- Bulk updates in metadata editor
- "merge-all-tags" for merging all detected tag values into a single attribute


badvision:
Recommend that you squash the commit to de-clutter the revision history.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PropertyMergePostProcessor.java
```diff
@@ -21,183 +21,257 @@
...
+
     private static final Logger log = LoggerFactory.getLogger(PropertyMergePostProcessor.class);
 
     private static final String AT_SUFFIX = "@PropertyMerge";
     private static final String ALLOW_DUPLICATES_SUFFIX = AT_SUFFIX + ".AllowDuplicates";
     private static final String TYPE_HINT_SUFFIX = AT_SUFFIX + ".TypeHint";
     private static final String IGNORE_PREFIX = ":";
+    private static final String ALL_TAGS = "merge-all-tags";
+    private static final String VALID_JCR_NAME = "[^:/\\[\\]\\|\\s*]+";
+    private static final Pattern VALID_TAG = Pattern.compile("^" + VALID_JCR_NAME + ":" + VALID_JCR_NAME + "(/" + VALID_JCR_NAME + ")*$");
```
davidjgonzalez: Selecting ONLY a tag name space is a valid tag value ... ex. 

`"^" + VALID_JCR_NAME + ":$"` is valid as well.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PropertyMergePostProcessor.java
```diff
@@ -21,183 +21,257 @@
...
+            return false;
+        } else {
+            return Stream.of(params).allMatch(param -> looksLikeTag(param.getString()));
+        }
+    }
 
-            final String typeHint =
-                    typeHintParam != null ? typeHintParam.getString() : String.class.getSimpleName();
+    protected static boolean looksLikeTag(String value) {
+        return VALID_TAG.asPredicate().test(value);
```
davidjgonzalez: I wonder if this should be a bit more stringent and use `tagManager.resolve(..)` to *make sure* its a tag, vs that it just "looks like" a tag.

bundle/src/test/java/com/adobe/acs/commons/wcm/impl/PropertyMergePostProcessorTest.java
```diff
@@ -189,6 +200,75 @@ public void testMerge_Duplicates_Calendar() throws Exception {
         Assert.assertArrayEquals(new Calendar[]{ january, september, july, september },
                 properties.get("dates", Calendar[].class));
     }
+    
+    @Test
+    public void testTagDetection() {
+        Assert.assertTrue("Valid root tag detection", PropertyMergePostProcessor.looksLikeTag("Some_Root:Tag"));
```
davidjgonzalez: `Assert.assertTrue("Valid tag namspace detection", PropertyMergePostProcessor.looksLikeTag("Some_Root:"));` Should pass as well.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PropertyMergePostProcessor.java
```diff
@@ -21,183 +21,257 @@
...
 @Service
 public class PropertyMergePostProcessor implements SlingPostProcessor {
+
     private static final Logger log = LoggerFactory.getLogger(PropertyMergePostProcessor.class);
 
     private static final String AT_SUFFIX = "@PropertyMerge";
     private static final String ALLOW_DUPLICATES_SUFFIX = AT_SUFFIX + ".AllowDuplicates";
     private static final String TYPE_HINT_SUFFIX = AT_SUFFIX + ".TypeHint";
     private static final String IGNORE_PREFIX = ":";
+    private static final String ALL_TAGS = "merge-all-tags";
```
davidjgonzalez: Something about this naming doesnt pop out to me as it being a special operation.. which is it (vs a property name; even thought property name's SHOULD be camelCase). If we wind up having more special operations like this, it might be nice to have a clean convention .. maybe something like:

`Operation.MergeAllTags` 

.. WDYT?

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PropertyMergePostProcessor.java
```diff
@@ -21,183 +21,257 @@
...
-            // Get the source value as type T
-            final T[] tmp = properties.get(source, emptyArray);
-
-            // If the value is not null, add to collectedValues
-            if (tmp != null) {
-                collectedValues.addAll(Arrays.asList(tmp));
+            Resource sourceProperties = resource;
+            String sourceParam = source;
+            if (source.contains("/")) {
+                sourceProperties = resource.getResourceResolver().getResource(resource, source.substring(0, source.lastIndexOf('/')));
```
davidjgonzalez: `StringUtils.substringAfterLast("source", "/")` is a bit easier to read.

also forgiving if there is not "/" at all.. just returns the String. 

https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#substringAfterLast-java.lang.String-java.lang.String-

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PropertyMergePostProcessor.java
```diff
@@ -21,183 +21,257 @@
...
+
+            T[] tmp = sourceProperties.adaptTo(ModifiableValueMap.class).get(sourceParam, emptyArray);
+            collectedValues.addAll(Arrays.asList(tmp));
         }
 
-        final T[] currentValues = properties.get(destination, emptyArray);
+        Resource targetResource;
+        if (destination.contains("/")) {
+            targetProperty = destination.substring(destination.lastIndexOf('/') + 1);
+            targetResource = resource.getResourceResolver().getResource(resource, destination.substring(0, destination.lastIndexOf('/')));
```
davidjgonzalez: `StringUtils.susbtringBeforeLast(destination, "/");` a little easier to read.

also forgiving if there is not "/" at all.. just returns the String. 

https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#substringBeforeLast-java.lang.String-java.lang.String-

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PropertyMergePostProcessor.java
badvision: I thought about that -- but I also thought about how slow TagManager is, too.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PropertyMergePostProcessor.java
badvision: Oh nice!  I didn't know about that method.  Thanks!

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PropertyMergePostProcessor.java
badvision: I think that anything long becomes hard to enter when working with the asset schema editor UI.  Though it is more coder-friendly, it becomes rather business user hostile and harder to remember.  I use hypens because I ultimately do a case-insensitive match on it to make it more user friendly.  There is a slim-to-nil chance of someone using this as a property name.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PropertyMergePostProcessor.java
davidjgonzalez: The TagManager.resolve(..) code does not look slow. Why do you think its slow, and prohibitively slow for an activity that is bounded by human interaction?

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PropertyMergePostProcessor.java
davidjgonzalez: If youre dead set on it, might as well remove "merge" and just have "all-tags", it's even less for the schema editor user to remember; "merge" is a bit of an implementation idiom. (couldve just as easily been "combine" or another synonym).

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PropertyMergePostProcessor.java
badvision: I have my reservations about TagManager in general, but more to the point this has to survive bulk editing too, so any additional work can become multiplicative.  If we use TagManager, we should also consider a small memory-limited cache to avoid looking up the same tags over and over.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PropertyMergePostProcessor.java
badvision: Well all-tags is likely to be a name for a property.  What if we used a non-jcr friendly prefix like you did for :?  Like :@all-tags or something?

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PropertyMergePostProcessor.java
davidjgonzalez: @badvision have you looked at the resolve code? it seems very simple (string checks).. or am i missing something? I dont think that this check will make or break anything in the context of writes; i'd be far more concerned around ~everything else (oak write overheads, any property indexes effected, WF's via launchers getting triggered). 

Do you have experience w `tm.resolve(..)` specifically where its prohibitively slow?  You seem very concerned about :) .. i feel like perhaps im missing something.

badvision: If you've moved or renamed tags, it's not that simple.  The whole implementation is a mess I would rather avoid.  But if this is what you want to accept the PR, then consider it added.  I have reservations about it, but pre-mature optimizations are the root of all evil so I'll shelve the caching idea for now.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PropertyMergePostProcessor.java
davidjgonzalez: surely a property `all-tags` would be unusual since thats what `cq:tags` would (should) be for? In a way the raison d'etre for this feature is to aggregate tags into cq:tags, since cq:tags is the only property that the rest of AEM knows about, but managing multiple namespaces of tags in a single property is bad UX.

`:@all-tags` is good with me; there is a clear indication that this is "special" about this vs someone being lazy when coming up a property name. 

Ultimately someone is gonna have to look this up one way or another..  seems simple, and good to me.

PR #1199, opened 2017-12-07
implementing dynamic loader for TouchUI JavaScript. fixes #989

### LINKED ISSUE #989: /editor.html requests parsys specific designs rather than the template design resulting in many 404s when parsys are nested
In AEM 6.2 with ACS Commons 3.8.2 installed, if a component with a dynamic path (such as one in a parsys) contains a parsys, /editor.html makes an ajax call for a design for that specific parsys which results in a 404.

These requests are being made by /apps/acs-commons/touchui-widgets/source/touchui-configure-parsys-placeholder.js Line 130 - https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/source/touchui-configure-parsys-placeholder.js#L130

How to reproduce?
To prep, we'll need to create a component with a parsys in it. We can do this by modifying a geometrixx component like so:
1) Go to /crx/de/index.jsp#/apps/geometrixx-gov/components/thankyoumessage/thankyoumessage.jsp
2) Add the following to the end of the file
<cq:include path="anotherparsys" resourceType="foundation/components/parsys"/>
3) Click save all

Now we'll enable the component in design mode
1) Go to /editor.html/content/geometrixx-gov/en/thankyou.html
2) Switch to design mode from the dropdown in the upper right next to "Edit"
3) Click on the lowest box marked "Drag components here"
4) Click the "Parent" option in the toolbar that appears
5) Click the wrench in the toolbar that appears
6) Scroll down to Geometrixx Gov and expand it
7) Select the check box next to Thank You Message inside the group
8) Click OK

Next we'll nest a few of these new components
1) Switch back to edit mode from the dropdown in the upper right next to "Design"
2) Switch to the components list in the side panel
3) Change the side panel dropdown to "Geometrixx Gov"
4) Drag the Thank You Message component into the top "Drag components here box" immediately below the line that says "Please specify your application tracking number."
5) Repeat step 4 twice, using the highest "Drag components here box" each time.

Open dev tools and refresh the page. Note the 404s in the request log such as
/etc/designs/geometrixx-gov/jcr:content/contentpage/par/thankyoumessage/anotherparsys.2.json

Proposed Fix
Retrieve the design once at the template level in the same way that AEM's native code does with /etc/designs/express/_jcr_content.0.json
OR
Provide a way to disable this feature.

Other Benefits to fixing:
Reduce the number of requests made.
Eliminate 404s for parsys prior to a design being selected.

justinedelson:
This is essentially the same technique used for the existing Classic UI libraries. As with that, the existing two libraries (limit and placeholders) are enabled by default for backwards compatibility reasons.


modeckimellett:
Thanks for working in a fix for #989 I really appreciate it!

PR #1200, opened 2017-12-08
Feature/httpcache jcr cache 6.0

Sc0rpic0m:
-Same as before but now on correct branch

-Includes implementation of review comments

-Still need to write additional unit tests


content/src/main/content/jcr_root/etc/acs-commons/httpcache/.content.xml
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
davidjgonzalez: Should this be `/var/acs-commons/httpcache` ?

/etc is falling out of favor, and /var is the spot for transient, system-managed content.

content/src/main/content/jcr_root/etc/acs-commons/httpcache/root/.content.xml
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
davidjgonzalez: Same as above..

/etc is falling out of favor, and /var is the spot for transient, system-managed content.

bundle/src/main/java/com/adobe/acs/commons/httpcache/store/jcr/impl/visitor/InvalidateByCacheConfigVisitor.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.apache.sling.commons.classloader.DynamicClassLoaderManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.adobe.acs.commons.httpcache.config.HttpCacheConfig;
+import com.adobe.acs.commons.httpcache.keys.CacheKey;
+import com.adobe.acs.commons.httpcache.store.jcr.impl.handler.EntryNodeToCacheKeyHandler;
+
+public class InvalidateByCacheConfigVisitor extends AbstractNodeVisitor
```
davidjgonzalez: Can't think of a clever way around this, but IIUC this will walk the entire cache looking for entries to invalidate, correct? 
HttpCache config's can be "dynamic" so i dont think its even possible to use some sort of hashing / direct lookup unfortunately..

Wonder if there should be a WARN in the logs if the traversal takes over N seconds, since invalidates are less controlled than maintenance walks; on bulk activations this could start hitting this rather hard if the size of the jcr cache space is unbounded. InMem cache gets around this by simply limiting the size of the cache.

Anyhow - dont think this can really be improved, just wanted to mention it in case it sparks any ideas.

Sc0rpic0m: Well, the intend of the JCR storage as I see it is to store entries that contain alot of data, so to preserve memory.. and that despite server restarts the cache is preserved.

So in that regard, we could limit the amount of entries that are stored? On service restart we could (recount) the amount of entries, and then use atomiclong to keep track of the count. Only problem I see then is, which entries to evict without heavy io usage... If you want to evict the ones based on the FIFO principle you'd have to do a shit ton of IO, that will never work I guess. We could throw out random entries instead or something.. not sure if you have any idea for this?

Sc0rpic0m: Damn :( been thinking of this. But I see no way to efficiently know which entries to evict. We would have to storage the hitcount of these entries on the node, like we have in the guava cache. That would result in alot of write operations, one for each view.

In theory then you can store the total hit count off all entries and get an average, and then remove the ones that are below a threshold. 
Even then, you would still have to loop the entire cache in the scheduled job again next to purgeExpiredEntries(), until you have evicted a certain amount of entries.

Should we do this in a eventual later release?

davidjgonzalez: @Sc0rpic0m agree this can be a future enhancement; I think if we drop a WARN in with the time it takes to walk and invalidate/evict that should be fine -- say, if the time is > 1 second? 2?

 since we dont have a great solution today, no sense in prematurely optimizing esp if we dont have a optimization we love :)

Sc0rpic0m: Alright! Ready then to be merged in. I'll see if I can add some extra unit tests still but you can accept those in later.

Sc0rpic0m:
Alright, ready to be accepted, unless you have some remarks regarding the final commits of course

davidjgonzalez:
@Sc0rpic0m merged! thanks!

FWIW - I think his could be merged into the main-line with just a service user (vs admin RR)

PR #1205, opened 2017-12-12
fix unclosed stream. fixes #1204

justinedelson:

PR #1209, opened 2017-12-13
Calculate the MD5 from the minified clientlib rendition in case

### LINKED ISSUE #1204: Unclosed stream in VersionedClientlibsTransformerFactory
Reported in #1203 by @kwin

### LINKED ISSUE #1208: Versioned Client Libs: Requesting mixed minified/unminified client libraries leads to incomplete validation
## Required Information

* AEM 6.3 Version SP1
* ACS AEM Commons Version: 3:13
* Reproducible on Latest? yes

### Expected Behavior

When modifying client library B) from category b* to b
and embedding b from client library A) then A should be invalidated (i.e. the MD5 should be newly generated)

### Actual Behavior

AEM (even in the most recent AEM 6.3 SP1) only maintains 1 flag per clientlibrary which defines whether a clientlibrary is dirty (i.e. needs to be rebuilt or not). If the clientlibrary is first being requested in the minified version, the dirty flag is being reset. If some other code is requesting now the clientlibrary in an unminified version, the clientlibrary is not longer being flagged "dirty" and therefore the unminified version is not correctly regenerated. The same vice-versa (i.e. when first requesting the unminified version only that one is correctly invalidated).

Requesting the unminified client library happens internally by AEM Commons in https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/b1ef63a5a6ae9feaed00c37c876512963452a7aa/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/VersionedClientlibsTransformerFactory.java#L288.

The call to `htmlLibrary.getInputStream()` always requests the unminified version (which in turn will lead to setting the dirty flag back to `false`). Any subsequent requests to the minified version will lead to delivering the stale minified clientlib.

### Steps to Reproduce

1. Create client library A) with category `a` and embed `b`
1. Include this client library in a simple page
1. Enable the minification on the HTML Client Library
1. Create client library B) with category`b*`
1. Now request page from step2
-> You see that the clientlibrary A) is being requested with an MD5 and the response does not contain content from B)
1. Now modify the client library B) in a way that the category is now `b` instead of `b*`
1. Now request page from step2 again
-> Although the MD5 has changed, the actual content of the client library has not. You need to manually remove the cached clientlibrary from `/var/clientlibs/...` to actually see the embedded clientlib's content as well.

Although one might argue that having only one dirty flag for potentially two different renditions of the same clientlibrary is a bug in AEM, one can easily workaround that bug in ACS Commons. For that just request the client library in in https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/b1ef63a5a6ae9feaed00c37c876512963452a7aa/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/VersionedClientlibsTransformerFactory.java#L288 with parameter `true` or `false` depending on whether minification is enabled or disabled!

I will come up with a PR for that.

kwin:
minification is enabled.

Fixes #1204
Fixes #1208


kwin:
This overwrites #1205

justinedelson:
Thanks! Change looks good to me. Can you take a look at the failing unit tests?

kwin:
I fixed the unit test.

PR #1218, opened 2017-12-19
AEM Report Builder

klcodanr:
Hey @davidjgonzalez & @justinedelson --

So this one came out of a current project and an assessment. The one of the big asks was to allow for reports in AEM. For example, if I wanted to know where all of the references to a path existed, or know where pages are within a process. 

This is to create such a faculty. It's an extremely configurable reporting mechanism. The only thing really built in is the concept of:

- *config* -- the configuration for the report executor. By default a query based report executor is provided
- *parameters* -- AEM fields for displaying parameters to configure the report executor
- *columns* -- called for every row in the results to display the result cell to the user

By default, there are a number of useful examples provided and I have a couple more in the works. 

To use it, you'd start at the report landing page:

Reporting Landing Page

From there you can create, edit and delete reports, you could click on the pencil to edit a report:

Editing a Report

**Note --** the query is evaluated by Handlebars, so you can perform basic logic.

Finally, if you click on the report title, you can run the actual report:

Executing a Report

As you configure the parameters, it'll update the hash so that you can share the URL with others to share the results.

The details show the query executed and how it's being interpreted by the query engine. You can page through the results or get all the results. I have run into an issue if there are > 300 or so results where this will cause Sling to kill the request, so I added an error to that effect.

You can also download a CSV representation of the report results which is rendered by Sling Models associated with the columns.


klcodanr:
@justinedelson @davidjgonzalez -- in my opinion, the remaining two @codeclimate major flags are by design. The objective of the ResultsPage class is to be a neutral container of results, no matter what type. The only provided class uses Sling Resources, but there's no reason you couldn't have reports which operate on a completely different type of results.

justinedelson:
@klcodanr I think the issue with the `ReportExecutor` class which codeclimate/sonar is highlighting is that the caller has no awareness of what the result type parameter would be. So from the caller's perspective, it makes no difference whether `ResultPage` has a type parameter or not; the calling code always has to assume the the underlying list is `List<Object>` (as seen elsewhere in the pull request).

I suppose it might make sense to have something like

    ResultsPage<T> getAllResults(Class<T> resultType) throws ReportException;

Which then would give the caller awareness of what the result type was going to be.

msiraj20:
When will this pull will merge to master

klcodanr:
Not sure @msiraj20 -- Adobe just returned from their winter shutdown so I'd assume that they're digging out inboxes. 

@justinedelson -- as discussed, I added a second method with the parameter to enable specifying a type. It's still throwing an exception because I'm not using that class, just casting the results into the correct type, since the type is a parameter for the returning collection.

justinedelson:
sorry for the delay... @klcodanr I'm not sure your latest changes are really better from a generics standpoint. Let me mull it over for a day or so and try to come back with some changes.

justinedelson:
@klcodanr Please take a look at https://github.com/klcodanr/acs-aem-commons/pull/1 which effectively removes the parameterized types. LMK either here or there what you think.

justinedelson:
@klcodanr one thing I'm noticing is that when I create a new report, the Components tab in the sidebar is empty. Do we need to do something special with the design settings to make that work?

klcodanr:
@justinedelson thanks for the catch! I forgot to set the `cq:designPath` when creating the new pages. Fixed in dcd845a5f962dbdd6ccf12bc8a65431635506f94

klcodanr:
Hm, it looks like the Travis CI failure was a timeout when trying to download dependencies, not a "real" problem.

justinedelson:
thanks! merging.

PR #1219, opened 2017-12-19
Removing PowerMock Tests

klcodanr:
Removing a number of PowerMock tests where they seem to have no effect. PowerMock tests are not reported correctly by Jacoco:

https://github.com/jacoco/jacoco/issues/51


justinedelson:
Thanks for starting this @klcodanr. I'm going to merge this offline with #1220.

Issue #943, opened 2017-03-28
Migrate from org.apache.commons.json to gson


justinedelson:
The org.apache.commons.json is in the process of being deprecated and may eventually be removed from AEM. I would like to get ahead of this problem and start removing org.apache.commons.json from ACS AEM Commons. Although the replacement used by Sling will be Apache Johnzon, that library is not yet ready for use (and will not be in AEM 6.3). As such, I think we should use gson which is available in AEM 6.2 and 6.3. It is *not* available in 6.0, so this work would only apply to the 6.2 branch.

ireasor:
@justinedelson is there anything else to do here?  I ran a search through the code for org.apache.commons.json and came up empty handed.

badvision:
Is this issue still pending @justinedelson, @ireasor ?

PR #1224, opened 2018-01-08
Feature/943 gson refactor

ireasor:
Refactored several components to use GSON per #943.
Note that there are some old commits from April being listed here.  I'm not sure why that is, but it doesn't look like there is actually any new changes being merged?


justinedelson:
Thanks @ireasor !

PR #1225, opened 2018-01-10
Workflow delegation: allow configuration via /conf

bcsott:
Extension to the workflow delegation process step. It allows to configure the delegation workflows in /conf
This has some benefits:
- No need to modify /content
- ConfMgr inheritance can be used
- Clean separation of content and configuration

Documentation was changed in adobe-consulting-services.github.io.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/WorkflowDelegationStep.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/WorkflowDelegationStep.java
@@ -144,17 +144,17 @@ public class WorkflowDelegationStep implements WorkflowProcess {
     private WorkflowModel getDelegateWorkflowModel(WorkflowSession workflowSession, String confWorkflowModelId,
         String contentWorkflowModelId, String defaultWorkflowModelId) throws WorkflowException {
       WorkflowModel workflowModel = null;
-      final String DEBUGTEXT = "Using configured delegate Workflow Model [ {} ] found in {}";
-      final String EXCEPTIONTEXT = "Could not find configured Workflow Model [ %s ] in %s";
+      final String debugText = "Using configured delegate Workflow Model [ {} ] found in {}";
+      final String exceptionText = "Could not find configured Workflow Model [ %s ] in %s";
 
       // Check /conf
       if (StringUtils.isNotBlank(confWorkflowModelId)) {
         workflowModel = getWorkflowModel(workflowSession, confWorkflowModelId);
 
         if (workflowModel != null) {
-          log.debug(DEBUGTEXT, workflowModel.getId(), "/conf");
+          log.debug(debugText, workflowModel.getId(), "/conf");
         } else {
-          throw new WorkflowException(String.format(EXCEPTIONTEXT, confWorkflowModelId, "/conf"));
+          throw new WorkflowException(String.format(exceptionText, confWorkflowModelId, "/conf"));
         }
         // Check Content
       } else if (StringUtils.isNotBlank(contentWorkflowModelId)) {
@@ -162,9 +162,9 @@ public class WorkflowDelegationStep implements WorkflowProcess {
         workflowModel = getWorkflowModel(workflowSession, contentWorkflowModelId);
 
         if (workflowModel != null) {
-          log.debug(DEBUGTEXT, workflowModel.getId(), "/content");
+          log.debug(debugText, workflowModel.getId(), "/content");
         } else {
-          throw new WorkflowException(String.format(EXCEPTIONTEXT, confWorkflowModelId, "/content"));
+          throw new WorkflowException(String.format(exceptionText, confWorkflowModelId, "/content"));
         }
 
         // Check default
@@ -172,10 +172,10 @@ public class WorkflowDelegationStep implements WorkflowProcess {
         workflowModel = getWorkflowModel(workflowSession, defaultWorkflowModelId);
 
         if (workflowModel != null) {
-          log.debug(DEBUGTEXT, workflowModel.getId(), "parameter defaultworkflowmodelid");
+          log.debug(debugText, workflowModel.getId(), "parameter defaultworkflowmodelid");
         } else {
           throw new WorkflowException(
-              String.format(EXCEPTIONTEXT, confWorkflowModelId, "parameter defaultworkflowmodelid"));
+              String.format(exceptionText, confWorkflowModelId, "parameter defaultworkflowmodelid"));
         }
       }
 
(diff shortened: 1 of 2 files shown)


bcsott:
I created some democontent which shows how it works and can be used for testing here: https://github.com/bcsott/acs-aem-commons/tree/develop

bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/WorkflowDelegationStep.java
```diff
@@ -11,11 +11,14 @@
 import org.apache.jackrabbit.JcrConstants;
 import org.apache.sling.api.resource.Resource;
 import org.apache.sling.api.resource.ResourceResolver;
+import org.apache.sling.api.resource.ValueMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.adobe.acs.commons.util.ParameterUtil;
 import com.adobe.acs.commons.util.WorkflowHelper;
+import com.adobe.granite.confmgr.Conf;
```
davidjgonzalez: The Granite Conf APIs are deprecated in AEM 6.3 [1] in favor of the Sling CAConfig APIs.

It would be good to check if CAConfig is available in AEM 6.2 and move to that if it is, else were introducing new code to Commons that uses deprecated APIs with 6.4's release right around the corner (which means we may/could drop 6.2 support in the main line of ACS Commons when its released)

https://helpx.adobe.com/experience-manager/6-3/sites/developing/using/reference-materials/javadoc/com/adobe/granite/confmgr/ConfMgr.html

bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/WorkflowDelegationStep.java
```diff
@@ -78,125 +103,217 @@
...
-
-        String propertyName = args.get(WORKFLOW_PROPERTY_NAME);
-        String defaultWorkflowModelId = args.get(DEFAULT_WORKFLOW_MODEL);
-        boolean terminateOnDelegation = Boolean.parseBoolean(StringUtils.lowerCase(args.get(TERMINATE_ON_DELEGATION)));
-
+    public void execute(WorkItem workItem, WorkflowSession workflowSession, MetaDataMap metadata) throws WorkflowException {
+
+      // Get Configuration
+      Map<String, String> args = getProcessArgsMap(metadata);
+      String confPropertyName = args.get(CONF_WORKFLOW_PROPERTY_NAME);
```
davidjgonzalez: Any reason we add this configuration param? Arent the Granite Conf's expecting the property to be `cq:conf` and if we can move to Sling CAConf `sling:configRef`.. when wouldn't it be the default?

Is this so if you are using Conf's but dont want it to drive WF delegation, that you can leave it out of process args which will "force" it to look at content property?

bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/WorkflowDelegationStep.java
```diff
@@ -78,125 +103,217 @@
...
+    /**
+     * Does fallback from 1. conf property 2. content property 3. default property
+     * 
+     * @throws WorkflowException
+     *           if found model id can not be located
+     */
+    private WorkflowModel getDelegateWorkflowModel(WorkflowSession workflowSession, String confWorkflowModelId,
+        String contentWorkflowModelId, String defaultWorkflowModelId) throws WorkflowException {
+      WorkflowModel workflowModel = null;
+      final String debugText = "Using configured delegate Workflow Model [ {} ] found in {}";
```
davidjgonzalez: static final String DEBUG_TEXT

bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/WorkflowDelegationStep.java
```diff
@@ -78,125 +103,217 @@
...
+      final String exceptionText = "Could not find configured Workflow Model [ %s ] in %s";
```
davidjgonzalez: static final String EXCEPTION_TEXT

bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/WorkflowDelegationStep.java
```diff
@@ -78,125 +103,217 @@
...
+
+      // Check /conf
+      if (StringUtils.isNotBlank(confWorkflowModelId)) {
```
davidjgonzalez: 151 - 160 and 162 - 168 look like ~dupe code; can this be a method?

bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/WorkflowDelegationStep.java
```diff
@@ -78,125 +103,217 @@
...
+    }
 
-        final Resource pageOrAssetResource = workflowHelper.getPageOrAssetResource(resolver, path);
-        if (pageOrAssetResource == null) {
-            log.warn("Could not resolve [ {} ] to an Asset or Page. Skipping delegation.", path);
-            return;
+    /**
+     * Ensures that workflow model id points to $model$/jcr:content/model
+     */
+    private WorkflowModel getWorkflowModel(WorkflowSession workflowSession, String workflowModelId) {
```
davidjgonzalez: not 100% necessary, but this seems like a good method to have in https://adobe-consulting-services.github.io/acs-aem-commons/apidocs/com/adobe/acs/commons/util/WorkflowHelper.html

Swear ive written ~ the same code before in this code-base and its a pretty common use-case.

bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/WorkflowDelegationStep.java
```diff
@@ -78,125 +103,217 @@
...
+      } else if (StringUtils.isNotBlank(defaultWorkflowModelId)) {
+        workflowModel = getWorkflowModel(workflowSession, defaultWorkflowModelId);
 
-        // Get the resource resolver
-
-        resolver = workflowHelper.getResourceResolver(workflowSession);
-        if (resolver == null) {
-            throw new WorkflowException("Could not adapt the WorkflowSession to a ResourceResolver. Something has gone terribly wrong!");
+        if (workflowModel != null) {
+          log.debug(debugText, workflowModel.getId(), "parameter defaultworkflowmodelid");
```
davidjgonzalez: I dont think `defaultworkflowmodelid` this is the actual parameter name?

bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/WorkflowDelegationStep.java
```diff
@@ -78,125 +103,217 @@
...
+        } else {
+          throw new WorkflowException(
+              String.format(exceptionText, confWorkflowModelId, "parameter defaultworkflowmodelid"));
```
davidjgonzalez: I dont think `defaultworkflowmodelid` this is the actual parameter name?  Better to use the Constant DEFAULT_WORKFLOW_MODEL.

bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/WorkflowDelegationStep.java
davidjgonzalez: This appears fine actually - ive heard cq:conf is supported.

PR #1228, opened 2018-01-17
added a config option to generate extensionless urls in sitemap

mayankarora03:
Added a config option to generate extensionless urls in sitemap. By default, URLs will have .html extension but if extensionless.urls property is set to 'true', there will be none.


justinedelson:
the drop in coverage is expected since this class (unfortunately) doesn't have an unit tests, but the codeclimate issue should be fixed before merging.

mayankarora03:
Fixed codeclimate issue and added entry in changelog

justinedelson:
Merged. Thanks. Can you also create a pull request to modify the documentation?

PR #1229, opened 2018-01-20
Nesreb/feature/sitemap trailing slash for extensionless urls

nesreb:
Mayank Arora added a nice feature to optionally exclude ".html" extensions from urls in sitemap.xml.  This adds a complementary option for including or excluding trailing slashes in those extensionless urls in sitemap.xml.  Thanks, Nes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 
 ## [Unreleased]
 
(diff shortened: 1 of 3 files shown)


justinedelson:
I'm going to merge this but need to set up the test infrastructure first since this class is not yet tested.

justinedelson:
Thanks!

nesreb:
Thanks @justinedelson!

PR #1234, opened 2018-01-26
additional not null checks for touch ui parsys operation

szymonkrzysztyniak:
Additional not null checks for touch ui parsys operation. This change was done in our project code as we belive this acs commons bug.


justinedelson:
Merged with some minor formatting changes. Thanks!

PR #1237, opened 2018-01-30
Adding a report column for finding references to a resource

klcodanr:


justinedelson:
@klcodanr I noticed that the redirectmaps.models package was missing a package-info.java file. I fixed that in master, but you'll probably want to merge that into your pull request branch to get rid of the build warning.

justinedelson:
according to https://stackoverflow.com/questions/19589260/force-travis-to-build-a-specific-pull-request-or-commit-sha-that-it-missed, to trigger the build again, I'm closing this and reopening it. fingers crossed.

bundle/src/main/java/com/adobe/acs/commons/redirectmaps/package-info.java
```diff
@@ -17,5 +17,5 @@
  * limitations under the License.
  * #L%
  */
-@aQute.bnd.annotation.Version("3.1.0")
+@aQute.bnd.annotation.Version("3.14.4")
```
justinedelson: I don't think this change is necessary. I only set the version to 3.14.4 on the models package because it wasn't set before and thus was getting the bundle version by default so I was just locking it in as 3.14.4. In this case, we already set a version, so there's no need to change.

klcodanr: Sure @justinedelson -- I thought I saw baseline complaining previously, but it worked fine this time so rolling it back to 3.1.0

PR #1242, opened 2018-01-31
Utility functions for fetching properties from a Resource

HitmanInWis:
Handful of fully tested utility functions for grabbing properties off of a Resource without having to change to adapt to a `Node` in the calling code.  Of particular use is the `getPropertyStrings()` function which automatically handles the case where a value is a single value in the JCR instead of an array and `getPropertyReference()` which grabs a resource reference.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/util/ResourceUtil.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/util/ResourceUtil.java
@@ -22,11 +22,7 @@ package com.adobe.acs.commons.util;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.sling.api.resource.Resource;
 
-import javax.jcr.Node;
-import javax.jcr.PathNotFoundException;
-import javax.jcr.Property;
-import javax.jcr.RepositoryException;
-import javax.jcr.Value;
+import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.List;
@@ -36,30 +32,6 @@ import java.util.List;
  */
 public abstract class ResourceUtil {
     /**
-     * Private ctor to hide the defualt public ctor.
-     */
-    private ResourceUtil() {
-    }
-
-    /**
-     * Get a single-value property from a resource.
-     *
-     * If the property does not exist, this function will return null instead
-     * of throwing a RepositoryException.
-     *
-     * @param resource The resource from which to get the property.
-     * @param namePattern Property name.
-     * @return Property value.
-     */
-    public static Property getProperty(Resource resource, String namePattern) throws RepositoryException {
-        try {
-            return resource.adaptTo(Node.class).getProperty(namePattern);
-        } catch (PathNotFoundException p) {
-            return null;
-        }
-    }
-
-    /**
      * Convenience method for getting a single-value boolean property from
      * a resource.
      *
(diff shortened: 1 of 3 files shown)


justinedelson:
thanks!

Rather than adding individual comments, let me ask a more general question taking one of the cases as an example. If I flatten out the code, the method `getPropertyBoolean()` essentially does this:

    Property prop = resource.adaptTo(Node.class).getProperty(namePattern);
    return prop != null && prop.getBoolean();

Why not just do

    resource.getValueMap().get(namePattern, false);

This is both less code and works across JCR and non-JCR `Resource` implementations. Other than `getProperty()` (which I suspect could be dropped), is there a reason to rely on the JCR API for these convenience methods?

HitmanInWis:
Thanks for the feedback Justin - very good point on the unnecessary dependency on the JCR API.  I guess that was just kinda a blind spot for me, trying to grab properties from the resource directly rather than chaining down into getValueMap(), which appears to do the same automatic type transitions that the specialized functions on `Property` do (e.g. `Property.getLong()` converts a String to a Long, but so does `ValueMap.get("prop", Long.class)` (which my test suite verifies).

That all being the case, I've checked in the updated code (which has an awesome benefit of no longer throwing `RepositoryException`).  However, I now question the value of the util functions other than `getPropertyReference()` and `getPropertyStrings()` since the rest are one-liners.  Thoughts?

bundle/src/main/java/com/adobe/acs/commons/util/ResourceUtil.java
```diff
@@ -132,30 +120,14 @@ public static String getPropertyString(Resource resource, String namePattern) th
...
-    private static Value[] getPropertyValues(Resource resource, String namePattern) throws RepositoryException {
-        Property prop = getProperty(resource, namePattern);
-        if (prop != null) {
-            if (prop.isMultiple()) {
-                return prop.getValues();
-            } else {
-                return (new Value[]{prop.getValue()});
+    public static List<String> getPropertyStrings(Resource resource, String namePattern) {
+        String[] vals = resource.getValueMap().get(namePattern, String[].class);
+        List<String> valsList = new ArrayList<>();
```
justinedelson: what about just using Arrays.asList() here? No reason (at least none that I can think of) for the resulting list to be mutable.

HitmanInWis: ya...I always shy away from making lists immutable and blindsiding someone with an error if they have a reason to try to add to the list, since the `List` interface really gives no indication of whether a list can be modified or not.  However, since all other data types returned from properties are immutable, I can see going that way as well.

justinedelson:
@HitmanInWis I was wondering about that :) I think there is a lot of value in the `getPropertyReference()` method specifically and it would be weird to just have the class for one or two methods, so I'm inclined to keep it. 

The other thing I was thinking about, but wanted to do a scan of some existing code bases is if there was value in having a method which returned a `Date` since I think `ValueMap` (and definitely JCR) only ever return a `Calendar`. I feel like I've written the code to get a `Calendar`, and then call `Calendar.getTime()` (which always seems like a weird thing to do, but the `Calendar` class is weird in general) a whole bunch of times and it would be nice to have a one-liner to do that.

HitmanInWis:
I added `getPropertyDate()` now with a return of `java.util.Date` and it seems to work straight from the `ValueMap` - that is unless the mocks behind the scenes do not reflect true reality of whats available in AEM...

justinedelson:
Merged. Thanks!

PR #1244, opened 2018-01-31
Feature/ensure group

shsteimer:
Extends the Ensure Service User functionality to support Ensure Group.  Refactored some of the service user logic into shared classes to keep things clean.  Updated the Manager class and JMX operations to operate on both ensure service user and ensure group configs.

For the most part, this is just a natural extension of the existing behavior, allowing you to model needed groups, without having to move around group package and set membership on each instance individually.  I think it is especially handy because it supports setting memberOf, since membership is stored on the target group.  A typical config would look something like this.

```
operation="add"
aces=["type\=allow;privileges\=jcr:all;path\=/content/mycorp"]
member-of=["content-authors","tag-administrators"]
principalName="mycorp-authors"
ensure-immediately=B"true"
```


bundle/src/main/java/com/adobe/acs/commons/users/impl/EnsureAce.java
```diff
@@ -0,0 +1,281 @@
...
+     *            the resource resolver to perform the user management
+     * @param authorizable
+     *            the Authorizable (User or Group) the ServiceUser represents
+     * @param serviceUser
+     *            the Service User return # of ace entries that could not be processed
+     * @throws RepositoryException
+     */
+    @SuppressWarnings("squid:S3776")
+    public int ensureAces(ResourceResolver resourceResolver, Authorizable authorizable,
+            AbstractAuthorizable serviceUser) throws RepositoryException {
```
davidjgonzalez: WDYT about renaming this variable from serviceUser -> authorizable (and its cousins in the methods) ? When i read through it the first time, it made it sound like this was explicitly for serviceUsers because all the vars include "serviceUser".

bundle/src/main/java/com/adobe/acs/commons/users/impl/AbstractAuthorizable.java
```diff
@@ -0,0 +1,105 @@
+/*
```
davidjgonzalez: copyright across several files are malformatted.

bundle/src/main/java/com/adobe/acs/commons/users/impl/EnsureAce.java
```diff
@@ -0,0 +1,281 @@
...
+        final Map<String, String> params = new HashMap<String, String>();
+
+        params.put("type", AccessControlConstants.NT_REP_ACE);
+        params.put("property", AccessControlConstants.REP_PRINCIPAL_NAME);
+        params.put("property.value", principalName);
+        params.put("p.limit", "-1");
+
+        final Query query =
+                queryBuilder.createQuery(PredicateGroup.create(params), resourceResolver.adaptTo(Session.class));
+        final Iterator<Resource> resources = query.getResult().getResources();
```
davidjgonzalez: QueryBuilder has a RR resource leak - it'd better to write this as:

```
query.getResult().getPaths().steam().forEach(path -> { Resource resource = resourceResolver.getResource(path); ... }
```

bundle/src/main/java/com/adobe/acs/commons/users/impl/EnsureAuthorizableManager.java
```diff
@@ -5,12 +5,12 @@
 import com.adobe.granite.jmx.annotation.Name;
 
 @ProviderType
-@Description("ACS AEM Commons - Ensure Service User MBean")
-public interface EnsureServiceUserManager {
+@Description("ACS AEM Commons - Ensure Authorizable MBean")
+public interface EnsureAuthorizableManager {
 
-    @Description("Execute all Ensure Service User configurations")
+    @Description("Execute all Ensure Service User & Ensure Group configurations")
```
davidjgonzalez: Curious as why one description uses `&` and the other 'and' :) .. my OCD coming out perhaps.

bundle/src/main/java/com/adobe/acs/commons/users/impl/EnsureAce.java
SSTEI43: I don't believe (I could be wrong) I changed that code, just copied it to a new class to make it easier to re-use. either way...

I don't see a `getPaths()` method on the result object, so I'm using `query.getResult().getHits().forEach(hit->{})` hope that works.

Would be curious to know more about the Resource Leak in QB, Is that documented as a known issue anywhere?  As far I can tell the code doesn't "look" like it should cause a leak, unless I'm missing something.

bundle/src/main/java/com/adobe/acs/commons/users/impl/AbstractAuthorizable.java
SSTEI43: I'll check this.  I tend to habitually reformat my code as I work, so possible my formatter munged them up.

bundle/src/main/java/com/adobe/acs/commons/users/impl/EnsureAce.java
davidjgonzalez: yup - was going off memory on this one; youre right - there is not `getPaths()` ... what you suggested (`query.getResult().getHits().forEach(hit->{rr.getResource(hit.getPath()})`) is what I meant! 

The QB issue is logged as an issue via Adobe Support; The issue is a JCR Session object is passed into QB to provided security context, and that JCR Session is then used to open a new RR, and QB never closes this RR ... in large part because its very hard to tell when its safe to close it .. since when its closed since objects of the ilk `query.getResult.getResources()` will become invalid (since their backing RR has been closed).

Not engaging any of the "Sling" methods on QB's results happily avoids the opening on the RR, so its safe (even if its only "safe" by way of a QB implementation detail).

bundle/src/main/java/com/adobe/acs/commons/users/impl/EnsureAce.java
```diff
@@ -240,21 +249,23 @@ public void removeAces(ResourceResolver resourceResolver, Authorizable authoriza
...
-            Resource contentResource = resources.next().getParent().getParent();
-
-            if (!paths.contains(contentResource.getPath())) {
-                for (AccessControlPolicy policy : accessControlManager.getPolicies(contentResource.getPath())) {
-                    if (policy instanceof JackrabbitAccessControlList) {
-                        acls.add((JackrabbitAccessControlList) policy);
-                        break;
+        query.getResult().getHits().forEach(hit -> {
+            try {
+                Resource aceResource = hit.getResource();
```
davidjgonzalez: Oops... so this is going to be problematic.. we want this to me;:

```
query.getResult().getHits().forEach(hit -> {
   try {  
           Resource aceResource = resourceResolver.getResource(hit.getPath());
   ...
```

The problem is whenever we call a "Sling" (getResource or getProperties) method on the QB results, its creates the leaking RR, so we want to make sure we never do that, and instead use OUR RRs which we ensure is closed properly.

shsteimer: ah, got it.  that makes sense...will fix it later tonight.

badvision:
Can you update your branch to resolve the reported conflicts?

shsteimer:
for you @badvision, anything.

shsteimer:
BTW, I have a half-baked branch of the docs pages that documents this feature that I still need to finish off and submit.  I'm hoping to have time for that next week, but we'll see if that actually happens.

davidjgonzalez:
@shsteimer PR'd some minor updates to YOUR branch that should get this one through to merge; so if you can review and merge when you have a chance will be good!

Issue #325, opened 2014-09-23
Code Droplets


davidjgonzalez:
The Problem
Lately I've needed to run "mini-scripts" against multiple envs to post-process content. These are 1 time executions that involve small code bases; currently the simplest way to execute is via ACS Tools > AEM Fiddle, but this is less controlled and requires the installation of the ACS Tools package (potentially on higher level envs).

Idea
Create a CQ Package installation hook that inspects the package for a signature (TBD), and executes a script in a well-known location in said package. Progress/status could be marked on a /var node somewhere to prevent "re-running" or collecting information post-execution.

Actual impl details TBD.

Ideally all the data would be clean and not require this, however IRL these needs arise, and it is seldom practical to make the changes centrally on Author and replicate them out (too much content, too much mixed content; replicated but modified)

This problem grows as the content volume and no of publishers grow.

davidjgonzalez:
@chetanmeh interested if you have thoughts on a simple/clean way to do this.

badvision:
This is kind of the general idea behind MCP's blockly-based scripting feature, which is slated for v2 of that featureset.  Would that fill the need?

PR #1245, opened 2018-02-01
New Feature: On-Deploy Scripts Framework

HitmanInWis:
This PR is nearly, but not 100% complete.  Putting it out there to:
- Ensure this is a desirable feature to pull into ACS AEM Commons (we use this all the time at HS2 Solutions)
- Get early feedback on cleanups

There are other tools out there that allow you to run arbitrary code scripts on your AEM server, but the purpose of this tool is to automate these scripts so that it is fully tested and repeatable via CI deployment to pre-production environments.  See the "Why use this tool?" section on https://github.com/HS2-SOLUTIONS/hs2-aem-commons/tree/master/on-deploy-scripts-framework (that project repo will be obsolete after this PR).

There's just a few things I need to finish up before this PR is fully ready:
- There's a race condition where OSGi configs deploy prior to the bundle that contains the on-deploy script classes - need to determine a solution...
- May want to add unit tests to OnDeployExecutorImpl
- Curious to get feedback on ACLs for the service user...as of now I've granted read access to the full JCR, and write access to /content and /etc


justinedelson:
@HitmanInWis I agree that this looks like pretty commonly needed functionality and is in many ways a generalization of things like `EnsureServiceUser` & `EnsureOakIndex`.

A general, but yet specific, comment on the implementation. Is there a purpose of loading the script classes from the ClassLoader vs. having these be proper OSGi services? The problem with the approach you are using here is that it requires that those script classes be part of the API (so they are exported and visible from the dynamic classloader). This seems like an odd requirement to me.

What I would suggest instead is just to have the scripts be proper OSGi services (implementing the `OnDeployScript` interface) and then have an immediate component which "runs" those scripts when they are bound. But maybe I'm missing something.

Another question I have is with respect to the service user. Are the permissions sufficient? Obviously we can always change this later. What would be awesome (although I don't know how achievable this is) is allow script creators to set up their own users if they need different permissions, declare that user as a service property, and then have the service user impersonate that user. Not necessarily something to deal with at first, just an idea.

Finally, I was a little tripped up by the term "script". I usually think of scripts as being, you know, scripts :), i.e. written in ECMAScript or Groovy (bless you if you write a migration script in JSP). Not a big deal, but "process" might be a better term. If I'm the only one confused by this.

HitmanInWis:
Hi Justin, thanks for the immediate feedback.

Do you have an example you can point to on how the Executor would wait for the individual services to be bound before executing them?  I think its important to ensure the order of execution of the scripts, which is why I took this simple approach, but I'm definitely open to modifying.

Regarding the service user, we can really give any permissions we want - all we're really doing by limiting the permissions is trying to stop people from doing bad things accidentally.  I figured `write` access to /content and /etc is a good starting point, since implementors that want increased permissions can do so by deploying additional ACL nodes.  Im definitely open to recommendation here...

As for "Script" I use the term b/c it feels like more of a "one time" connotation versus a "Process" which may be used over and over.  However, let's see if @davidjgonzalez or others have an opinion on naming.

justinedelson:
bq. Do you have an example you can point to on how the Executor would wait for the individual services to be bound before executing them? 

Hmmm. Good question. https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/763d2968169be1d86937bc1f4354cf2f8e201866/bundle/src/main/java/com/adobe/acs/commons/users/impl/EnsureServiceUserManagerImpl.java#L71 kind of does this, although it just stores the services in a map and then executes them later. I'll poke around and see if there is a better example.

bq.  I think its important to ensure the order of execution of the scripts, which is why I took this simple approach, but I'm definitely open to modifying.

In theory, I believe the bind method should be called in order based on the service ranking, but this should be verified. But ordering in general is a very good point. I was thinking of the scripts as largely independent from each other, but it sounds like you have seen use cases where the ordering is important. Let me think about this a bit more.

bq. Regarding the service user, we can really give any permissions we want.

Yep. And I think what you've done here is a good first step. We can always iterate.

bq. As for "Script" I use the term b/c it feels like more of a "one time" connotation versus a "Process" which may be used over and over. 

Fair enough. I'm terrible at naming things, just giving you my first reaction :) But this comment raises an interesting point -- there doesn't seem to be any checking to ensure a script doesn't run multiple times. Is that correct? If you look at LiquidBase, for example, it keeps track of the changesets run and only runs the "newer" ones. I'm not sure this is necessary here (and in AEM in general), but we do need to document it.

HitmanInWis:
> I was thinking of the scripts as largely independent from each other, but it sounds like you have seen use cases where the ordering is important.

Yep.  Simple example can be where Script 1 renames a property from "a" -> "b" and Script 2 renames the property from "b" -> "c".  If Script 2 runs before Script 1, Script 2 will miss any property "b" created by Script 1.

> there doesn't seem to be any checking to ensure a script doesn't run multiple times. Is that correct?

Once a script has run successfully, it stores a node under `/var/acs-commons/on-deploy-scripts-status` with a status of `success`.  Next time the executor runs, it skips past any jobs that have a status of `success`.  As such, scripts do run only one time.

There are docs on this on the `OnDeployExecutorImpl` class, and would also be documented on the web page I create for the feature once we finalize it.

justinedelson:
bq. Once a script has run successfully, it stores a node under /var/acs-commons/on-deploy-scripts-status with a status of success. Next time the executor runs, it skips past any jobs that have a status of success. As such, scripts do run only one time.

Oh cool. I missed that part. Need to look more closely at the code :)

davidjgonzalez:
FWIW when i saw this the PR name i thought scripts (JSP, groovy, etc.) as well.

Is the assumption that bundles of "scripts/processes" would be deployed separately from the main application package, so they dont have to be tied to an application release? Thought arguable a bit more dangerous, i always liked the notion of actually using scripts since they dont require all the trappings of a maven project, and feel more ad-hoc, rather part of a formal project (but maybe thats for the best.)

This in alot of ways is similar to https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/325 (so i very excited about this 👍 )

still have to make my way through all the docs/samples/code! but looks cool

HitmanInWis:
I think they can really be deployed however you want.  We've generally been doing it as part of an application package, because the reason for the script is often because of a change in the application.  For example, let's say we refactor a component to a new folder structure.  When that change goes live, we want all content using that sling:resourceType to be updated to the new path.  The deployment of the application and the content update (via on-deploy script) need to happen at the same time.  Now that isnt to say you couldnt deploy ad-hoc packages to run scripts as well, deploying along with the application is just how we've been using it to keep things simple.

Good memory on referencing #325 :).  Yes, that kinda of functionality was the inspiration behind this.  Coming from a background of working on systems backed by a SQL database, a code release was almost always accompanied by a "SQL release script" to update tables and such.  This implementation was my way of enabling that pattern in AEM.

HitmanInWis:
What do you guys think about renaming "Scripts" to "Jobs".  Does that make it more clear?  Or does "Job" also have a connotation of a repeating background task?  Maybe "Tasks"?

HitmanInWis:
So I think the remaining items on this ticket are:
- Finalize naming (OnDeployScript, OnDeployProcess, OnDeployTask, OnDeployJob)
- Clean up final code climate issues and add further unit testing
- Resolve a race condition where OSGi configs deploy with script classes prior to classes being available.

**Re: naming** - I like either OnDeployScript or OnDeployTask, as I feel Process and Job are things that run more than once.  Happy to consider other names.

**Re: code climate/unit testing** - thats just something I need to finish up

**Re: race condition** - This is one I could use a hand with...  Note that I havent had to solve this issue yet in HS2's implementation because in our impl we simply hard-coded OnDeployScript classes into a list directly in the bundle code, so there is no possibility of the race condition.

Here is the current situation when configuring via OSGi: Deploy a client package that contains the OSGi config for OnDeployExecutorImpl, as well as the bundle that contains the script class.

In this situation, the OSGi config deploys to the JCR first, triggering OnDeployExecutorImpl to run.  When OnDeployExecutorImpl tries to gather the script classes, it cannot find the new script class because the included bundle has not yet completed deployment.  OnDeployExecutorImpl throws an error and exits.  The script is not executed, and will only execute in one of the following scenarios:
- AEM Server restart
- Reinstall of ACS Commons package (triggers restart of OnDeployExecutorImpl)
- Reinstall of client package (reinstalls the OSGi config for OnDeployExecutorImpl, triggering restart)

I could go back to an approach to how we do it internally right now - simply omit OSGi configuration, hard-coding the script instances into an OnDeployScriptProvider that would be picked up by OnDeployExecutorImpl.  By eliminating the OSGi configuration part of the equation, there would no longer be an race conditions because the list of scripts is directly part of the bundle that contains the scripts.

Thoughts?

bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutorImpl.java
```diff
@@ -0,0 +1,273 @@
...
+    @Reference
+    private QueryBuilder queryBuilder;
+    @Reference
+    private ResourceResolverFactory resourceResolverFactory;
+    @Reference
+    private DynamicClassLoaderManager dynamicClassLoaderManager;
+
+    private List<OnDeployScript> scripts = null;
+
+    // TODO: These were needed for injection to work in tests
```
HitmanInWis: @justinedelson @davidjgonzalez My sling-mock-fu is weak, and for whatever reason I was required to have these `bind` functions in order for the `@Reference` fields above to be injected correctly.

Is there a way around that, or can either of you provide insight as to why these are required and/or if its a good thing to keep these?

HitmanInWis:
Ok, I think I have the unit testing fairly complete.  Now to figure out those last two items (for which I'll wait on feedback first).

bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutorImpl.java
```diff
@@ -0,0 +1,239 @@
...
+    private QueryBuilder queryBuilder;
+    @Reference
+    private ResourceResolverFactory resourceResolverFactory;
+    @Reference
+    private DynamicClassLoaderManager dynamicClassLoaderManager;
+    @Reference(name = "scriptProvider", referenceInterface = OnDeployScriptProvider.class, cardinality = ReferenceCardinality.MANDATORY_MULTIPLE, policy = ReferencePolicy.DYNAMIC)
+    private List<OnDeployScriptProvider> scriptProviders;
+
+    // TODO: These were needed for injection to work in tests
+    protected void bindQueryBuilder(QueryBuilder queryBuilder) {
```
HitmanInWis: These `bindQueryBuilder`, `bindResourceResolverFactory`, and `bindDynamicClassLoaderManager` functions were needed for unit tests to work, but are not needed in true runtime.  Anyone have a clue if there's just something I need to change in the sling-mock tests to make these `bind` functions unnecessary?

HitmanInWis:
So I resolved the race condition issue by adding an `OnDeployScriptProvider` interface for which services can simply feed in real object instances of the script classes.  I see no downside to this compared to OSGi configs b/c the standard use case is the script classes are custom code deployed with the same bundle that wants to run those scripts.

**This PR is ready for code review**

bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutorImpl.java
davidjgonzalez: You couldnt use Mockito's @Inject?

bundle/src/main/java/com/adobe/acs/commons/ondeploy/scripts/OnDeployScriptBase.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    /**
+     * Update the sling:resourceType of a node.
+     *
+     * @param node         The node to update.
+     * @param resourceType The new sling:resourceType to be used.
+     */
+    protected void updateResourceType(Node node, String resourceType) throws RepositoryException {
+        String currentResourceType = node.getProperty(SLING_RESOURCE_TYPE).getString();
```
davidjgonzalez: Should this do a hasProperty() check if someone else calls this (thats not searchAndUpdateResourceType()) ?

bundle/src/main/java/com/adobe/acs/commons/ondeploy/scripts/OnDeployScript.java
```diff
@@ -0,0 +1,36 @@
...
+ * An on-deploy script that can be run via the OnDeployExecutor.
+ */
+public interface OnDeployScript {
+    /**
+     * Execute the script, passing in a resourceResolver and queryBuilder instance.
+     *
+     * @param resourceResolver Resource resolver.
+     * @param queryBuilder Query builder.
+     */
+    void execute(ResourceResolver resourceResolver, QueryBuilder queryBuilder);
```
davidjgonzalez: Is there a reason why QueryBuilder is explicitly passed in? You could get in the Base via `resourceResolver.adaptTo(QueryBuilder.class)` . Passing this is feels a big awkward IMO since not all scripts will query.

bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutorImpl.java
HitmanInWis: maybe, but that kinda subverts the usage of sling mock, no?  sling mock is successfully seeing the `@Reference` attributes but when it then tries to set them via call to `bind{name}` whereas the real osgi runtime does not require a `bind{name}` function.   At least that's my understanding.  Maybe one of these build plugins we use is actually adding the `bind{name}` functions to the classes automatically at build time?

bundle/src/main/java/com/adobe/acs/commons/ondeploy/scripts/OnDeployScript.java
HitmanInWis: how about "because we always did it that way"?

Seriously, this is likely just a vestige of my youth when I didnt realize a `ResourceResolver` could be adapted to a `QueryBuilder`.  Good call on that, I'll add it to my cleanups.

bundle/src/main/java/com/adobe/acs/commons/ondeploy/scripts/OnDeployScriptBase.java
HitmanInWis: yes that's a good call, it really should.

HitmanInWis:
Per review I've got just a few minor code cleanups to make.  I imagine Justin will want to comb the code a bit as well.

Do we still want to consider a rename for this?

- OnDeployScript (current)
- OnDeployTask (my next preference)
- OnDeployJob
- OnDeployProcess

bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutorImpl.java
HitmanInWis: heh...I think I answered my own question...

I took a look at `com.adobe.acs.commons.oak.impl.EnsurePropertyIndex` in the code - no methods for `bindCapabilityHelper` or `bindRepository` to match up with

```
    @Reference
    private AemCapabilityHelper capabilityHelper;

    @Reference
    private SlingRepository repository;
```

However, if I look at the class in the compiled jar, this is what I find:

```
  protected void bindCapabilityHelper(AemCapabilityHelper paramAemCapabilityHelper)
  {
    this.capabilityHelper = paramAemCapabilityHelper;
  }
  
  protected void unbindCapabilityHelper(AemCapabilityHelper paramAemCapabilityHelper)
  {
    if (this.capabilityHelper == paramAemCapabilityHelper) {
      this.capabilityHelper = null;
    }
  }
  
  protected void bindRepository(SlingRepository paramSlingRepository)
  {
    this.repository = paramSlingRepository;
  }
  
  protected void unbindRepository(SlingRepository paramSlingRepository)
  {
    if (this.repository == paramSlingRepository) {
      this.repository = null;
    }
  }
```

So these bind functions must need to be explicit for sling mock if running tests in intellij, because intellij is not running whatever maven plugin is automatically adding these functions.

I guess I'll keep the functions then, since the compile process adds them otherwise anyway.

bundle/src/main/java/com/adobe/acs/commons/ondeploy/scripts/OnDeployScript.java
davidjgonzalez: ahh - to be young again :)

HitmanInWis:
Code review feedback has been addressed.

bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutorImpl.java
```diff
@@ -0,0 +1,233 @@
...
+        logger.info("Executing on-deploy scripts from scriptProvider: {}", scriptProvider.getClass().getName());
+        List<OnDeployScript> scripts = scriptProvider.getScripts();
+        if (scripts.size() == 0) {
+            logger.debug("No on-deploy scripts found.");
+            return;
+        }
+        ResourceResolver resourceResolver = null;
+        Session session = null;
+        try {
+            resourceResolver = logIn();
```
royteeuwen: why not use try with resources? This is available since AEM 6.2

bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutorImpl.java
```diff
@@ -0,0 +1,233 @@
...
+    private static final String SERVICE_NAME = "on-deploy-scripts";
+
+    private final Logger logger = LoggerFactory.getLogger(this.getClass());
+
+    @Reference
+    private ResourceResolverFactory resourceResolverFactory;
+    @Reference
+    private DynamicClassLoaderManager dynamicClassLoaderManager;
+    @Reference(name = "scriptProvider", referenceInterface = OnDeployScriptProvider.class, cardinality = ReferenceCardinality.MANDATORY_MULTIPLE, policy = ReferencePolicy.DYNAMIC)
+    private List<OnDeployScriptProvider> scriptProviders;
```
royteeuwen: I don't see an implementation of this interface in the PR?

HitmanInWis: Correct.  This is a framework that allows other bundles to deploy scripts that will be run by this service.  ACS Commons would likely never include scripts itself.

royteeuwen: Ah, I thought the DeployScripts had to be implemented, it wasn't clear that also the provider had to be implemented :)

HitmanInWis: Yes this allows different bundles to each provide scripts.  The reason the scripts are provided from coded OnDeployScriptProvider implementations rather than simply an OSGi config is to prevent a race condition with OSGi configs deploying separate from the script classes.

The provider class is pretty simple:

```
/**
 * Provider for the OnDeployExecutor service.
 *
 * Specifies a list of OnDeployScript instances to run.
 */
@Component(immediate = true)
@Service
@Properties({
        @Property(name = "service.description", value = "Developer service that identifies code scripts to run on deployment via the OnDeployExecutor service")
})
public class OnDeployScriptProviderImpl implements OnDeployScriptProvider {
    /**
     * Get the list of scripts to run.
     *
     * Ideally, all scripts will remain in the list indefinitely, so that a new AEM
     * server can run all scripts from the first to the last to be
     * completely up to date w/no manual intervention.  As a reminder, scripts will
     * run only once, so it is safe to preserve the entire list of scripts.
     *
     * @return List of OnDeployScript instances.
     */
    @Override
    public List<OnDeployScript> getScripts() {
        return Arrays.asList(
                new TestScript(),
                new TestScript2()
        );
    }
}
```

HitmanInWis:
@davidjgonzalez I believe I have your requested changes addressed.  Anything more on this?  Or is it just a matter of needing to find more time to do a deeper review of the feature?

davidjgonzalez:
@justinedelson did you have other thoughts on this? LGTM

davidjgonzalez:
@justinedelson this good to merge from you’re POV? Lgtm

HitmanInWis:
Once merged we’ll be sure to put together a documentation page for the site

HitmanInWis:
@davidjgonzalez can you mark your review as "approved" since the issues have been addressed?

justinedelson:
Did some minor cleanup before merging:

* Remove unused OnDeployExecutor service interface
* Remove JCR usage from OnDeployExecutorImpl
* Change node type of /var/acs-commons/on-deploy-scripts-status to sling:Folder
* Ensure OnDeployScriptBase is designed for safe extension

justinedelson:
and thanks @HitmanInWis for your patience!

HitmanInWis:
Thanks for the cleanups and the merge @justinedelson !

Quick question on the change of /var/acs-commons-on-deploy-scripts-status from sling:OrderedFolder to sling:Folder - I had it as an OrderedFolder so that status nodes would be preserved in the order they were created, keeping them for the most part chronological when viewing in the JCR.  What was the reasoning behind the change?

justinedelson:
@HitmanInWis essentially it was to improve performance as adding nodes to an ordered parent gets slower over time. Since this is a background operation, I'm not opposed to switching it back if there is  a good reason.

justinedelson:
switched in 113390a7

HitmanInWis:
Thanks @justinedelson.

One final thing - in your updates to remove dependency on JCR I think `trackScriptStart()` is broken.

Line 203 of OnDeployExecutorImpl needs to be updated from:
`properties.put(SCRIPT_DATE_END, null);`
to
`properties.remove(SCRIPT_DATE_END);`

`ModifiableValueMap` doesn't take too kindly to putting `null` values.

HitmanInWis:
Ok, perfect.  Final testing of `master` branch on AEM 6.2 and 6.3 looks good.  This feature is locked and loaded!

bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutorImpl.java
yuvraj1610: I am getting unsatisfied reference error using it with 6.3 :+1: 
Unsatisfied
Service Name: com.adobe.acs.commons.ondeploy.OnDeployScriptProvider
Cardinality: 1..n
Policy: dynamic
Policy Option: reluctant
No Services bound

HitmanInWis: @yuvraj1610 what does your implementation class for `OnDeployScriptProvider` look like?

Issue #1250, opened 2018-02-12
Redirect Map upload reveals fails with "multipart boundary" error


kcmcgrath:
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: *6.3 Base Quickstart*
* [x] ACS AEM Commons Version: *3.14.6*
* [x] Reproducible on Latest? *yes*

### Expected Behavior

Uploading a simple .map file in to the redirect-map-manager should succeed

### Actual Behavior

On uploading a map, the UI appears to take the upload, but then nothing happens after page refresh. System error logs reveal: 
```
*ERROR* [qtp1517652272-166] org.apache.sling.engine.impl.parameters.ParameterSupport parseMultiPartPost: Error parsing request org.apache.commons.fileupload.FileUploadException: the request was rejected because no multipart boundary was found
```

### Steps to Reproduce

 - From base 6.3 quickstart, install ACS Commons 3.14.6
 - Navigate to http://localhost:4502/miscadmin#/etc/acs-commons/redirect-maps and create a new page
 - Create a `test.map` file with the following content: 

```
/test www.test.com/new
/testmore www.test2.com/more
```
 - Upload `test.map` file (Attempted in both Chrome 63.0.3239.132 and Safari 11.0.2)
 - Observe error in error.log

justinedelson:
@klcodanr any ideas on this?

klcodanr:
@justinedelson - yes, it seems like a regression introduced by my previous javascript change. I'd set the POST Ajax call to:

```
    $.ajax({
                url: $form.attr('action'),
                data: new FormData($form[0]),
                cache: false,
                contentType: 'multipart/form-data',
                processData: false,
                type: 'POST',
                success: function(data){
                	location.reload(true);
                }
            });
```
Which doesn't set the multitype boundary. Oddly enough, when the Sling Post Servlet gets a file post without a Multitype Boundary rather than returning an error, it seems to just return 200, so I totally missed this when I smoke tested the change.

I just created a PR #1251 to resolve this. Sorry for the trouble over a 1 line issue!

PR #1251, opened 2018-02-12
Fixing an issue when uploading a new map file

klcodanr:
Fixing an issue when uploading a new map file as reported in #1250


justinedelson:
Thanks @klcodanr. I'm going to cut a 3.14.8 bug fix release with this in it shortly.

PR #1255, opened 2018-02-18
Feature/dispatcher flush rules trimming

royteeuwen:
Add trimming of dispatcher flush rules
Add crx junction to allow deployment to servers with a junction


davidjgonzalez:
@royteeuwen can you make an issue for the crx.junction and add that to the changelog? Would be good to link out to the documentation for it as its an esoteric configuration. We shouldn't hide this change in the dispatcher issue. 

Otherwise looks good.

royteeuwen:
Also solves #1256

PR #1247, opened 2018-02-08
Feature/dynamic select parameter component for report builder

DXSaurabh:
**_NOTE: @davidjgonzalez made some suggestions/changes via #1258 but cannot PR it to DXSaurabh so i just PR'd it to master. This and #1258 are ~ the same, and eventually one will need to supersede the other._**

Hi @justinedelson/ @dchang ,

I created the one new component select parameter for report builder component. This component is having the feature for populating the select parameter drop-down by JCR Query instead of static options.

dynamic-dropdown

Thanks

PR #1258, opened 2018-02-21
#1247 - Moved DataSource to Servlet (cannot PR into DXSaurabh's branch)

davidjgonzalez:
@DXSaurabh i was thinking something more like this.
For some reason i cannot PR into your repo which is what i wanted to do :T ... Not sure if there is some setting on your GitHub account that would let me PR into it?

Anyhow - I havent tested this (TBH, im not really sure how? havent kept up on report builder...) but the major things are:

* Made it bind to a resourceType and NOT a path (servlets bound to paths are an antipattern)
* I used the ACS Commons DataSourceBuilder API that let me clean things up significantly.
* I moved off parsing the JCR-SQL2 query to figure out what the SELECT X X X  Columns are and just added a mutlifield text field where an author can just put those in. This has the side effect of now various query languages (including my fave, querybuilder) can be used to define the queries.

Can you check this out and see if i understood what you were trying to do?

Also - would be good to get some light test coverage in as well.

+ @klcodanr


davidjgonzalez:
Closing as this was merged into 1247

Issue #1260, opened 2018-02-21
MCP status page breaks for large file uploads


badvision:
AEM Version: 6.3+
Commons version: 3.11+
Reproducible: Yes.

Currently there is a recursive death in the MCP servlet when the following conditions occur:
- User uploads a large spreadsheet, like say 300kb or so
- MCP task is started and can be seen working in the JMX beans
- MCP servlet fails during GSON serialization because it gets stuck in a recursive loop

Currently the best solution I've identified is to edit the ManagedProcess bean and make the requestInputs property transient so it is excluded from serialization.  The original intent of including this in the bean was to later surface this in the UI -- but unfortunately that feature idea will have to be done differently to avoid this problem.  For now, since it isn't made visible in the UI, there should be no impact in usability to exclude from the JSON until a better way is identified.

justinedelson:
fixed via #1261

PR #1261, opened 2018-02-21
Marked processInputs transient

badvision:
as per issue #1260


justinedelson:
@badvision can you update the changelog?

Issue #1272, opened 2018-03-04
Ensure Service User service is not restricted ACE by path.


drfits:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: AEM 6.3
* [ ] ACS AEM Commons Version: ACS 3.13.0
* [ ] Reproducible on Latest? yes

### Expected Behavior

When I'm set this list of ACEs aces="[type=allow;privileges=jcr:read\,rep:write;path=/content/drfits,type=allow;privileges=jcr:read\,rep:write;path=/content/dam,type=allow;privileges=jcr:read;path=/apps/drfits/templates,type=allow;privileges=jcr:read\,rep:write;path=/etc/tags,type=allow;privileges=jcr:read;path=/home]"/>

All of this ACE have to be added into the OAK repository.

### Actual Behavior

Not all ACEs applied.

### Steps to Reproduce

Root cause why it's happening: ACE do not have restriction by ACL path and first found ACE will be returned from ServiceUser.java class [at this point](https://github.com/Adobe-Consulting-Services/acs-aem-commons/compare/master...drfits:issue/ensure-service-user?expand=1#diff-01fbf1269757794474266940e78d8a10L94)

PR #1273, opened 2018-03-04
#1272 - Fixed an issue when service user ACE not restricted by path.

drfits:
Fix for #1272 issue

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/users/impl/ServiceUser.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/users/impl/ServiceUser.java
@@ -91,9 +91,9 @@ public final class ServiceUser {
         return aces;
     }
 
-    public Ace getAce(JackrabbitAccessControlEntry actual) throws RepositoryException {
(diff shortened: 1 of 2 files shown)


bundle/src/main/java/com/adobe/acs/commons/users/impl/ServiceUser.java
```diff
@@ -91,9 +91,9 @@ public String getPrincipalName() {
         return aces;
     }
 
-    public Ace getAce(JackrabbitAccessControlEntry actual) throws RepositoryException {
+    public Ace getAce(JackrabbitAccessControlEntry actual, String path) throws RepositoryException {
         for (Ace ace : getAces()) {
-            if (ace.isSameAs(actual)) {
+            if (StringUtils.equals(path, ace.getContentPath()) && ace.isSameAs(actual)) {
```
davidjgonzalez: @drfits food for thought; should isSameAs(..) include this path check implicitly... have to look refresh my brain on this code, but on the surface seems like it should

Issue #1274, opened 2018-03-06
MCP: When adding file upload to a process, I would like to capture the filename optionally


badvision:
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.3+
* [x] ACS AEM Commons Version: 3.10+
* [x] Reproducible on Latest? yes

### Expected Behavior

I would like there to be a way to capture the file name when uploading to an MCP task

### Actual Behavior

Currently only InputStream is captured, all other information is lost in translation.

justinedelson:
closed via #1275

PR #1275, opened 2018-03-06
Feature/1274 mcp file improvements

badvision:
Feature #1274 added - Uploading file to a process now supports InputStream and RequestParameter types


badvision:
@justinedelson The coveralls test finished (no change) -- not sure why Girhub isn't getting the completion trigger.

PR #1277, opened 2018-03-12
New Feature : Debug sling component/resource paths

kunalgaba:
Proposing a new feature which allows AEM devs to quickly identify the resource paths of AEM components in the HTML. 

**Problem:**
 This feature is useful when there is a need to customise AEM Touch UI and find out the resource paths of AEM Touch UI Granite components.  The Devs currently use non-standard techniques to find out the right paths to be overlaid for customising the Touch UI. These techniques are time consuming and does not provide an instant solution. 

**Solution**
With this feature the devs can just inspect the HTML of the selected element and find the CRX resource paths in the HTML comments. The solution uses a Sling component filter which filters all requests and injects the resource path in the HTML response of the components. 

- The sling component filter works only when it is enabled in OSGI configuration. By default it is disabled. 
- The filter only modifies the response of the requests whose URI contains ".html" extension. 
- The format of the HTML comment is given below- 
 

 Screenshots for more examples. 
Asset finder-
image

Lock button in Sites UI-
image

I am still working on this PR and need to do some cleanups.  I have opened this to get comments/feedback from others on the solution.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/wcm/debug/SlingComponentDebugFilter.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/wcm/debug/SlingComponentDebugFilter.java
@@ -1,0 +1,4 @@
+package com.adobe.acs.commons.wcm.debug;
+
+public class SlingComponentDebugFilter {
+}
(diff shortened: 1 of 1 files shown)


davidjgonzalez:
This looks pretty cool - but should this be in ACS Tools project instead? This seems like a developer tool vs. something you'd run on production.

kunalgaba:
@davidjgonzalez - Thanks. I agree it should be part of acs-aem-tools. I have created a PR in tools repo https://github.com/Adobe-Consulting-Services/acs-aem-tools/pull/204 and we can close this.

PR #1284, opened 2018-03-19
Add the shared and global properties resources to the bindings

rbotha78:
The shared resources are required when working with CoralUI 3 multi-field values in HTL.


badvision:
Please add test coverage for this new functionality (however small, I know...) -- a quick example is to write a until test that instantiates SharedComponentPropertiesBindingsValuesProvider and runs a few basic methods to ensure they are not returning null.  This should allow your PR to pass the automated code coverage check

rbotha78:
Thanks @badvision
I've added a test for SharedComponentPropertiesBindingsValuesProvider 👍

PR #1294, opened 2018-03-26
New Feature: AEM Remote Assets

HitmanInWis:
This is a PR for the AEM Remote Assets feature, to be presented at Adobe SUMMIT AEM Rockstar on Wednesday Mar 28.

The purpose of AEM Remote Assets is to allow an AEM server to be fully functional with assets from a remote AEM instance without sync'ing the entire set of DAM asset file binaries (the bulk of the disk space) up front.  Asset file binaries are sync'd into the server the first time they are requested by a user.  To the end user it appears all assets are present, but in actuality some assets may be getting fetched "just in time" from the source AEM server as the user accesses them.  Once an asset file has been pulled in from the source server the asset is now a standard AEM asset.  The bulk of rarely-used assets, however, remain un-synced.

I would consider this feature "beta" level.  It's pretty solid in our local testing, but I encourage anyone looking to use this feature to do so with caution in non-production environments until the feature is further battle-tested.  Though this feature ultimately supports use of a remote AEM Assets instance from an AEM Sites instance, which is a very important business use case, it is also extremely useful for developers working on local AEM instances of production websites, where pulling the entire DAM from production to the local server is simply not an option due to size.

I understand that unit testing will need to be added to this PR before it is accepted, but wanted to get feedback on the code before diving into that effort.

Other "nice to have" features that I can see for the future of AEM Remote Assets include:
- Function to purge remote asset binaries (to manually maintain a small footprint)
- Expire image binaries after awhile (to automatically maintain a small footprint)
- Protection mechanisms against fetching assets > specified size
- Protection mechanisms against fetching too many assets at once
- Option to delete assets that are no longer in the source system when sync'ing nodes
- When syncing nodes, keep any renditions that are smaller than the temporary binaries.
- Sync content fragments and experiences fully (so they can be indexed and since they are just content rather than large binaries)
- Determine a way to sync Binary properties (currently ignored b/c JSON fetch simply gives the binary property size rather than anything useable)


HitmanInWis:
Here is a video demo'ing this new feature: https://s3.amazonaws.com/HS2Presentations/AEMPublic/2018SUMMITRockstar-RemoteAssets.mp4

**KEY DISCLAIMERS:**
- Ensure all workflow launchers for Assets have "event-user-data:changedByWorkflowProcess" included in the "Exclude List" field, so the workflows are not triggered by sync processes
- This feature should be considered "beta" quality and thus recommended to be used in non-production environments until further battle tested

HitmanInWis:
@davidjgonzalez @justinedelson can we get an initial, light review of the code sometime this week?  I just want to make sure everything is directionally sound, so we can start building out the unit tests to make the code coverage test happy.  I just want to make sure we didnt overlook some obviously different way of doing something before we write a bunch of tests and then fundamentally refactor the code.

I've gotten communications that a lot of people are looking forward to this feature, so I want to try to help it along as fast as we reasonably can.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetDecorator.java
```diff
@@ -0,0 +1,175 @@
...
+ * referenced by the system and sync that asset from the remote server to
+ * make it now a "true" asset.
+ */
+@Component(
+        label = "ACS AEM Commons - Remote Assets - Asset Resource Decorator",
+        description = "Captures a request for a remote asset so that the binary can be sync'd to the current server making it a true local asset",
+        policy = ConfigurationPolicy.REQUIRE
+)
+@Service
+public class RemoteAssetDecorator implements ResourceDecorator {
```
davidjgonzalez: Are there situations where this doesnt work? IIRC the Assets APIs are generally lose w/ always using Sling Resource APIs and often drop straight to the JCR Node APIs?

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetDecorator.java
```diff
@@ -0,0 +1,175 @@
...
+     * @param resource Resource to check
+     * @return true if resource is remote, else false
+     */
+    private boolean accepts(final Resource resource) throws RepositoryException {
+        if (resource == null) {
+            return false;
+        }
+
+        ValueMap props = resource.getValueMap();
+        if (!props.get("jcr:primaryType", "").equals("dam:AssetContent")) {
```
davidjgonzalez: Use constants? JcrConstants and DamConstants? .. sort of feels weird w the order, since dam:AssetContent is as static string and thus impossible to ever return null.. where as .get() isa method and if the params changed it could.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetDecorator.java
```diff
@@ -0,0 +1,175 @@
...
+            return false;
+        }
+
+        if (!props.get("isRemoteAsset", false)) {
```
davidjgonzalez: Is there a constant for this?

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetDecorator.java
```diff
@@ -0,0 +1,175 @@
...
+            return false;
+        }
+
+        if (remoteResourcesSyncing.contains(resource.getPath())) {
+            return false;
+        }
+
+        Calendar lastFailure = props.get("remoteSyncFailed", (Calendar) null);
```
davidjgonzalez: Constant?

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsBinarySyncImpl.java
```diff
@@ -0,0 +1,209 @@
...
+     * @return Resource
+     */
+    @Override
+    public Resource syncAsset(Resource resource) {
+        try {
+            this.session.refresh(true);
+            Resource localRes = this.resourceResolver.getResource(resource.getPath());
+            Node node = localRes.adaptTo(Node.class);
+
+            Asset asset = localRes.getParent().adaptTo(Asset.class);
```
davidjgonzalez: Is it better to use DamUtil.resolveToAsset(..)  since this is a public API (not private).

does it need nullchecks?

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsBinarySyncImpl.java
```diff
@@ -0,0 +1,209 @@
...
+            URI pathUi = new URI(null, null, asset.getPath(), null);
```
davidjgonzalez: pathURI? .. typo?

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsBinarySyncImpl.java
```diff
@@ -0,0 +1,209 @@
...
+            Iterator<? extends Rendition> renditions = asset.listRenditions();
+            while (renditions.hasNext()) {
+                Rendition assetRendition = renditions.next();
+                if (StringUtils.isEmpty(assetRendition.getMimeType())) {
+                    continue;
+                }
+                String renditionName = assetRendition.getName();
+
+                URL url = new URL(String.format("%s%s", baseUrl, renditionName));
+                HttpURLConnection connection = (HttpURLConnection) url.openConnection();
```
davidjgonzalez: Use HttpClient?

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsBinarySyncImpl.java
```diff
@@ -0,0 +1,209 @@
...
+                connection.setRequestMethod("GET");
+                connection.setRequestProperty("Authorization", String.format("Basic %s", RemoteAssets.encodeForBasicAuth(this.remoteAssetsConfig)));
```
davidjgonzalez: Feels like it should WARN or something if the connection is NOT https... since the password would be compromisable.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsBinarySyncImpl.java
```diff
@@ -0,0 +1,209 @@
...
+
+                if (LOG.isDebugEnabled()) {
```
davidjgonzalez: is this if { .. } necessary? its usually reserved for when the debug call is expensive;

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsBinarySyncImpl.java
```diff
@@ -0,0 +1,209 @@
...
+                    LOG.debug("syncing from remote asset url {}", url);
+                }
+
+                setRenditionOnAsset(connection, assetRendition, asset, renditionName);
+            }
+
+            node.setProperty("isRemoteAsset", (Value)null);
```
davidjgonzalez: Why not ModifiableValueMap.. ? bit cleaner?

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsBinarySyncImpl.java
```diff
@@ -0,0 +1,209 @@
...
+     * @param asset Asset
+     * @param renditionName String
+     * @throws FileNotFoundException exception
+     */
+    private void setRenditionOnAsset(HttpURLConnection connection, Rendition assetRendition, Asset asset, String renditionName)
+            throws IOException {
+
+        try (InputStream inputStream = connection.getInputStream()) {
+            Map<String, Object> props = new HashMap<>();
+            props.put("rendition.mime", assetRendition.getMimeType());
```
davidjgonzalez: Use constant... https://helpx.adobe.com/experience-manager/6-3/sites/developing/using/reference-materials/javadoc/constant-values.html

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsBinarySyncImpl.java
```diff
@@ -0,0 +1,209 @@
...
+            asset.setRendition(renditionName, inputStream, props);
+        } catch (FileNotFoundException fne) {
+            if ("original".equals(renditionName)) {
```
davidjgonzalez: Use constant.. https://helpx.adobe.com/experience-manager/6-3/sites/developing/using/reference-materials/javadoc/constant-values.html

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsBinarySyncImpl.java
```diff
@@ -0,0 +1,209 @@
...
+    /**
+     * Sets a property on the resource if the asset sync failed.
+     * @param resource Resource
+     * @return Resource
+     */
+    private Resource flagAssetAsFailedSync(Resource resource) {
+        try {
+            Resource localRes = this.resourceResolver.getResource(resource.getPath());
+            Node node = localRes.adaptTo(Node.class);
+            node.setProperty("remoteSyncFailed", new DateValue(new GregorianCalendar()));
```
davidjgonzalez: ModifiableValueMap + constant?

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncImpl.java
```diff
@@ -0,0 +1,663 @@
...
+    @Reference
+    private ResourceResolverFactory resourceResolverFactory;
+
+    @Reference
+    private RemoteAssetsConfig remoteAssetsConfig;
+
+    @Reference
+    private DynamicClassLoaderManager dynamicClassLoaderManager;
+
+    private ResourceResolver resourceResolver;
```
davidjgonzalez: not thread safe

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncImpl.java
```diff
@@ -0,0 +1,663 @@
...
+    /**
+     * Get {@link JSONObject} from URL response.
+     * @param path String
+     * @return JSONObject
+     * @throws IOException exception
+     * @throws JSONException exception
+     */
+    private JSONObject getJsonFromUri(final String path) throws IOException, JSONException {
+        try {
+            URI pathUi = new URI(null, null, path, null);
```
davidjgonzalez: typo in pathUi (pathUri)

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncImpl.java
```diff
@@ -0,0 +1,663 @@
...
+            // we want to traverse the JCR one level at a time, hence the '1' selector.
+            URL url = new URL(this.remoteAssetsConfig.getServer() + pathUi.toString() + ".1.json");
+            HttpURLConnection connection = (HttpURLConnection) url.openConnection();
```
davidjgonzalez: HttpClient

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncImpl.java
```diff
@@ -0,0 +1,663 @@
...
+            connection.setRequestProperty("Authorization", String.format("Basic %s", RemoteAssets.encodeForBasicAuth(this.remoteAssetsConfig)));
```
davidjgonzalez: Same WARN if not https as may leak password

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncImpl.java
```diff
@@ -0,0 +1,663 @@
...
+                    sb.append(line);
+                }
+
+                String sbString = sb.toString();
+
+                if (StringUtils.startsWith(sbString, "{")) {
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug("JSON successfully fetched for URL '{}'.", url.toString());
+                    }
+                    return new JSONObject(sbString);
```
davidjgonzalez: We use GSON now instead of Sling JSON as its deprecated.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncImpl.java
```diff
@@ -0,0 +1,663 @@
...
+                String line;
+                StringBuilder sb = new StringBuilder();
+
+                while ((line = br.readLine()) != null) {
+                    sb.append(line);
+                }
+
+                String sbString = sb.toString();
+
+                if (StringUtils.startsWith(sbString, "{")) {
```
davidjgonzalez: shouldnt the creation of a JSONObject from a invalid JSON string throw and exception? This is a weird check that im sure happens internally in the JSON APIs anyhow

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncImpl.java
```diff
@@ -0,0 +1,663 @@
...
+     */
+    private JSONObject getJsonFromUri(final String path) throws IOException, JSONException {
+        try {
+            URI pathUi = new URI(null, null, path, null);
+            // we want to traverse the JCR one level at a time, hence the '1' selector.
+            URL url = new URL(this.remoteAssetsConfig.getServer() + pathUi.toString() + ".1.json");
+            HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+            connection.setRequestProperty("Authorization", String.format("Basic %s", RemoteAssets.encodeForBasicAuth(this.remoteAssetsConfig)));
+
+            try (BufferedReader br = new BufferedReader(new InputStreamReader(connection.getInputStream()))) {
```
davidjgonzalez: So this is a big change, but can you pull down XML and use Session.importXML(..) to import everything?
You could even include the binary data in the XML and not make the separate calls...

something like...
http://localhost:4502/content/dam/we-retail/en/activities/climbing/climber-gear-outdoor.jpg/jcr:content.xml

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncImpl.java
```diff
@@ -0,0 +1,663 @@
...
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+/**
+ * Service to sync node tree from supplied felix configuration. Implements {@link RemoteAssetsNodeSync}.
+ */
+@Component(
```
davidjgonzalez: I think the big potential room for improvement would be using Session.importXML and pulling the asset down AS XML, and letting the JCR APIs sort out all the data transformations. It would cut this method down to a handful of lines.

All that JSON / property setting is scary :)

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncImpl.java
```diff
@@ -0,0 +1,663 @@
...
+        immediate = true,
```
davidjgonzalez: Why is this immediate?

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncImpl.java
```diff
@@ -0,0 +1,663 @@
...
+     * @see RemoteAssetsNodeSync#syncAssetNodes()
+     */
+    @Override
+    public void syncAssetNodes() {
+        List<String> syncPaths = new ArrayList<>();
+        syncPaths.addAll(this.remoteAssetsConfig.getTagSyncPaths());
+        syncPaths.addAll(this.remoteAssetsConfig.getDamSyncPaths());
+
+        try {
+            for (String syncPath : syncPaths) {
```
davidjgonzalez: So this is what creates the "remove asset nodes"... is this ONLY getting assets that directly reside in the syncPath folder? or can this feature descend into the folder sub-tree?

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncImpl.java
```diff
@@ -0,0 +1,663 @@
...
+    /**
+     * Create or update nodes from remote JSON.
+     * @param json JSONObject
+     * @param parentNode Node
+     * @throws IOException exception
+     * @throws JSONException exception
+     * @throws RepositoryException exception
+     */
+    private void createOrUpdateNodes(final JSONObject json, final Node parentNode) throws IOException, JSONException, RepositoryException {
+        Iterator<String> keys = json.keys();
```
davidjgonzalez: id be interested if the XML could be imported for these top level nodes, and not deal w JSON.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsBinarySyncImpl.java
```diff
@@ -0,0 +1,209 @@
...
+            Resource localRes = this.resourceResolver.getResource(resource.getPath());
+            Node node = localRes.adaptTo(Node.class);
+
+            Asset asset = localRes.getParent().adaptTo(Asset.class);
+            URI pathUi = new URI(null, null, asset.getPath(), null);
+            String baseUrl = this.remoteAssetsConfig.getServer().concat(pathUi.toString()).concat("/_jcr_content/renditions/");
+
+            Iterator<? extends Rendition> renditions = asset.listRenditions();
+            while (renditions.hasNext()) {
+                Rendition assetRendition = renditions.next();
```
davidjgonzalez: Im thinking these HTTP calls could get passed off to fast action manager to queue up and manage the work; it wont run away from you like if you were spawning threads, and it doesnt have the overhead of sling jobs.

Making the renditions sync synchronous on the request feels like trouble waiting to happen.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetDecorator.java
HitmanInWis: It's a good question - LMK a scenario you want to test.  I have yet to run into a valid scenario where I would expect it to work and it didnt.  In the case of Asset APIs returning JSON it would probably be ideal that it *doesn't* work b/c all of the metadata and nodes are there for a correct JSON API return, and then if someone actually accesses the asset to view it then at that time it would sync.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetDecorator.java
```diff
@@ -0,0 +1,175 @@
...
+    @Override
+    public Resource decorate(final Resource resource) {
+        try {
+            if (!this.accepts(resource)) {
+                return resource;
+            }
+        } catch (Exception e) {
+            // Logging at debug level b/c if this happens it could represent a ton of logging
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Failed binary sync check for remote asset: {} - {}", resource.getPath(), e.getMessage());
```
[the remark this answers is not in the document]
HitmanInWis: I was purposely trying to reduce the amount of logging here just because it could end up being a crazy amount of logging if something fundamental was breaking.  That said, I'm open to changing this if you disagree.  Please confirm.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetDecorator.java
```diff
@@ -0,0 +1,175 @@
...
+            return resource;
+        }
+
+        Resource ret = resource;
+        try {
+            remoteResourcesSyncing.add(resource.getPath());
+            LOG.info("Sync'ing remote asset binaries: {}", resource.getPath());
+            ret = this.assetSync.syncAsset(resource);
+        } catch (Exception e) {
+            LOG.error("Failed to sync binaries for remote asset: {} - {}", resource.getPath(), e.getMessage());
```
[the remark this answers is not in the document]
HitmanInWis: Same mentality as above - trying to reduce a done of repeated log clutter if possible.  But I can see that this could potentially make debugging more tricky.  Thoughts?

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetDecorator.java
HitmanInWis: Are you asking for this to be `if (!DamCostants.NT_DAM_ASSETCONTENT.equals(props.get(JcrConstants.JCR_PRIMARYTYPE)) {`?

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetDecorator.java
HitmanInWis: Can be, should be, will be.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetDecorator.java
```diff
@@ -0,0 +1,175 @@
...
+        if (lastFailure != null && System.currentTimeMillis() < (lastFailure.getTimeInMillis() + (this.config.getRetryDelay() * 60000))) {
+            return false;
+        }
+
+        for (String syncPath : this.config.getDamSyncPaths()) {
+            if (!resource.getPath().startsWith(syncPath)) {
+                continue;
+            }
+
+            Session session = resource.getResourceResolver().adaptTo(Session.class);
```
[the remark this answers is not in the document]
HitmanInWis: Yes, good call - the only thing that needs to be in the loop is the path check.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsBinarySyncImpl.java
```diff
@@ -0,0 +1,209 @@
...
+    private ResourceResolver resourceResolver;
+    private Session session;
+
+    /**
+     * Method to run on activation.
+     * @throws RepositoryException exception
+     */
+    @Activate
+    protected void activate() throws RepositoryException {
+        this.resourceResolver = RemoteAssets.logIn(this.resourceResolverFactory);
```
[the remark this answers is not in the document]
HitmanInWis: sure thing, will fix

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsBinarySyncImpl.java
HitmanInWis: Didn't realize DamUtil existed.  I'm fine with using it here if you think that we should.

A null check probably isnt a terrible idea as well - it *should* never happen since this service should only be called on a dam:AssetContent node, but it doesnt hurt to handle the case if someone were to pass something crazy in or if theres a bug somewhere else.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsBinarySyncImpl.java
HitmanInWis: I think this is a good consideration for a future iteration, where we monitor how many of these HTTP calls have queued up and are waiting, delegating to asynchronous sync when the queue gets a certain length.

Synchronous pull is what makes this function truly "magical".  Without it, the first time you access an asset you'd see the temporary binary.  The user would need to refresh the page to get the real thing, stealing much of the appeal of this feature.

If we add (again in a later enhancement) a feature to make requests async after X number are currently active, then a user could configure this to 0 if they truly want ALL syncs to be async.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsBinarySyncImpl.java
HitmanInWis: we can do that - should we also make it so that HTTPS is required unless a user configures specifically to allow HTTP?  For internal network traffic within a VPN it might not be a problem to be HTTP.

Or do you have alternate suggestions on how we authenticate?

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsBinarySyncImpl.java
HitmanInWis: Correct, not needed

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsBinarySyncImpl.java
```diff
@@ -0,0 +1,209 @@
...
+
+    /**
+     * Sets a property on the resource if the asset sync failed.
+     * @param resource Resource
+     * @return Resource
+     */
+    private Resource flagAssetAsFailedSync(Resource resource) {
+        try {
+            Resource localRes = this.resourceResolver.getResource(resource.getPath());
+            Node node = localRes.adaptTo(Node.class);
```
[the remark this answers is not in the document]
HitmanInWis: Because some of this code was written before you and Justin helped me understand that on the On-Deploy Scripts PR :)

Will update.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncImpl.java
HitmanInWis: Ha ok...so you're recommending that we use JSON calls to traverse the tree, but then when we have a single asset call for the XML and sync the asset in that way?

Couple questions on that...
1) It is recommended that default XML renderer be turned off on production instances - I believe this would require the remote server to have XML renderer turned on - not sure if this is a big deal or not - thoughts?
2) We'd still need to do some contortions to add in the temporary binary files when creating the nodes.

I agree that this code is somewhat scary, just want to make sure there is a strong and good reason to change it before we do...

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncImpl.java
HitmanInWis: The service already depends on `RemoteAssetsConfig` so it seemed fine to let this start on its own w/o requiring another config.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncImpl.java
HitmanInWis: I assume you mean "remote" asset nodes.  Yes, this is what syncs the nodes in.  Code is recursive, thus traversing the entire node tree downward one folder at a time.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncImpl.java
HitmanInWis: Saw your comment below about including the binary data.  We actually dont want the binary data included in this request.  This request is getting all of the node structure *except for* the binary files - that makes this sync both fast (at least fast-er) and SMALL (this is one of the main tenets of Remote Assets).

If we switch to XML, and there's no way to exclude the jcr:data properties from the fetched content, I think this could dramatically increase the cost of this sync on both AEM servers involved.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncImpl.java
HitmanInWis: See thread above...

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncImpl.java
HitmanInWis: Let's discuss in the thread on XML vs. JSON above, so as to keep one contiguous discussion.

HitmanInWis:
Thanks @davidjgonzalez for the solid first review.

- I had wondered about the thread safety thing, so thanks for confirming that is an issue. We will resolve
- We could consider using the Asset API to traverse the DAM structure, but it wouldnt work for sync'ing tags, not sure if it can include ALL properties, and ultimately unsure if it provides any benefit over using the default sling JSON rendering capabilities.
- The XML vs. JSON question is interesting - I have responded with some thoughts within the PR and look forward to further discussion.

kaushalmall:
Hi @HitmanInWis a few items to consider after talking to a whole bunch of people re: this PR and to make it useful for the whole community. 

1. AEM Remote Assets name seems to be confusing people, can we rename it to "AEM Assets Sync" or something similar. 
2. When we document it, lets make sure we have all the use cases defined (similar to the one you put in your blog) or just link the blog in ACS Commons docs (if David and Justin allow that) 
3. Is it possible to add support for a shared datastore based architecture? This would basically make the sync faster. 
4. A lot of feedback is coming on how the assets are syncing on search. Can we look into only syncing the assets if they are used on a page (primary use case for a Sites instance). And maybe add a feature flag for a button on the search results page to sync assets on an explicit user action
5. Would it possible to use replication or SCD to sync assets? This would automatically work for a shared DS architecture w/o you having to check anything in code. 

WDYT?

HitmanInWis:
Hi Kaushal,

1) Yes we can rename to AEM Assets Sync, though it could make the feature more difficult for people to find if they are searching for it based off of blog posts and AEM Rock Star press.  I'm open to if if we think it's important, but I personally feel AEM Remote Assets is more reflective of the "on-demand" nature where we dont actually sync the entire set of assets.

2) Sure, no problem - I generally try to include use cases in the instructions (as well as how to set the feature up correctly) - see https://adobe-consulting-services.github.io/acs-aem-commons/features/on-deploy-scripts/index.html

3) A shared data store architecture could be added as a future option.  Happy to entertain PRs from the community to add further options.  I think we still want to support non-shared (current impl) since many applications are not using a shared data store.

4) I think adding options for which features trigger a sync is a future possibility.  One of the good things about where we instrumented AEM for this (resource decorator) is that it "just works" whenever you ask for a particular asset, which eliminates edge cases where it doesnt work.  Another alternative to avoiding sync on search (how can you tell if you want to use an asset if you can't see it?) is to instead automatically de-sync assets (after X minutes) that are not published or referenced.

5) Need more information on how replication would be used for this, as I dont know enough about the underlying subsystems.  One of the keys is that we need to omit the binary files in the initial sync of the node structure, substituting in the temporary assets.

I'd be curious to hear more about the shared data store scenario, and how important we feel it is to cater this tool to that use case.  In my mind if someone is already using a shared data store this feature already provides a pretty limited value.  There are other great ways to sync node structures (e.g. Grabbit) that seemingly already address that use case.

In my eyes the main benefits of AEM Remote Assets are:
- Automation of the sync (Grabbit or other methods can do this for shared data store scenario)
- Sync the entire node structure but w/o original binaries (not a concern for shared data store scenario)
- Sync only the binary files for assets that are used (not a concern for shared data store scenario)

If I am overlooking some key benefits please let me know.  I do think, however, that the best course of action on this feature would be to get what we have thus far in as-is (with fixes from PR) and then further iteration of more features and functionality can be discussed.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsBinarySyncImpl.java
HitmanInWis: sweet reference!

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsConfigImpl.java
```diff
@@ -0,0 +1,343 @@
...
+    /**
+     * @see RemoteAssetsConfig#closeResourceResolver(ResourceResolver)
+     */
+    @Override
+    public void closeResourceResolver(ResourceResolver resourceResolver) {
+        if (resourceResolver != null) {
+            try {
+                Session session = resourceResolver.adaptTo(Session.class);
+                if (session != null) {
+                    session.logout();
```
HitmanInWis: No idea if this `logout()` is necessary - my gut says it isnt but I dont know enough of the internals to be sure.  The reason I'm concerned in the first place is b/c the `getResourceResolver()` function uses the session to `setUserData()` on the workspace observation manager so as to prevent triggering workflows.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncImpl.java
```diff
@@ -0,0 +1,599 @@
...
+     * @return Resource
+     * @throws RepositoryException exception
+     */
+    private Resource getOrCreateNode(final ResourceResolver remoteAssetsResolver, final String nextPath, final String primaryType) throws RepositoryException {
+        Resource resource;
+
+        try {
+            resource = remoteAssetsResolver.getResource(nextPath);
+            if (resource == null) {
+                Node node = JcrUtil.createPath(nextPath, primaryType, remoteAssetsResolver.adaptTo(Session.class));
```
HitmanInWis: If theres a simple way to do this w/o Node I'm happy to update this line.

davidjgonzalez:
@HitmanInWis  @kaushalmall  - so i was playing around this past 2 nights... and it seems that the sync times are very high (i.e. very noticeable as a user of the system).

For the scheduled sync this is transparent since its an async job (though watching the logs it takes 2 mins or so to pull my 50 assets down from my remote AEM instance -- which is remote, it's not running on my local - im using a hosted Asset Share Commons author and syncing the ASC folder).

The more concerning part is the renditions (binary) sync .. when i search and get a pull page of results, it tends to hang for quite a while (pulling in 10 asset x 5 renditions = 50 requests, some large)... 

I also noticed that certain requests, especially when accessing AEM Assets folders, tend to hit many assets in a folder.. so browsing, to a folder that contains other folders which in turn contain assets, will invoke the thumbnail folder servlet which seems to walk many assets in that folder (even though i never clicked in).

I also noticed that occasionally the following...
1) Request A references DAM Asset X (perhaps a request for the assets cards servlet in folder view)
2) Request A blocks and downloads Asset X (along w a many other assets, making this a long request)
3) Request B is initiated for an asset itself.
4) Request B references DAM Asset X, but DAM Asset X is already in the list of being sync'd .. so it doesnt sync.
5) Asset details page looks "broken" (ie still shows the placeholder).
6) Request A completes after the renditions are downloaded.
7) Refresh the details page and everything looks great.

Am I doing something wrong here? I watched your video and you didn't have delays but it sounded like you were running both Src and Dest locally removing the network time.

HitmanInWis:
Hi David.  I think all of your concerns are imperfections in the system as it stands today.  However, I think we should look at it from a right perspective of "is it useful for some use cases, even if not all use cases yet?".

I think its valid to say that there will be a delay when the remote AEM server is truly remote.  However, that's going to largely depend on connection quality between AEM instances.  Two servers on the same network, or in the AWS cloud, may perform significantly better than what you're seeing (maybe not as fast as in my demo, but much closer).

Regardless of the delay, keep in mind that this is a one-time "tax" you are paying for auto-sync of just the assets you want.  For example, a developer working on a client website locally would get "taxed" the first time they go to a certain page on the site, but after that page loads the first time it will load w/o any delay in the future (the assets are now in the local DAM).  The same can be said about a true Assets -> Sites scenario - the first time a search is performed it will delay in retrieving binaries, but the second time the assets will already be there.

Down another thought path, the core tenet of this feature is pulling only the assets that are *actually* in use - hence why doing the "just in time" sync.  I think for assets used in a site we could add a future enhancement to pre-sync binaries, thus eliminating the delay for that particular use case.  For use cases like asset search and DAM traversal, however, we'd still be doing "just in time".

Lastly, good catch on the race condition where the 2nd browser sees temporary binaries that the first browser is currently sync'ing.  We could definitely have the 2nd request sleep while the first thread is pulling the asset - we already have logic in there that's preventing that 2nd request from fetching the same binary that the first request is pulling, so the change would be relatively trivial (sleep 100ms in a loop until the asset is sync'd).

HitmanInWis:
@davidjgonzalez another thing I thought of - please make sure all of your asset workflow launchers have "event-user-data:changedByWorkflowProcess" in the "Exclude List" so that the sync does not trigger workflows.  Unfortunately I couldn't find an easier way to avoid triggering workflows, so this will be part of configuration instructions.

HitmanInWis:
Fixed the race condition you found @davidjgonzalez, where Thread 2 does not wait for Thread 1's binary sync to complete.  With the new impl, both Thread 1 and Thread 2 will correctly see the real asset when the page finishes loading.

HitmanInWis:
@davidjgonzalez is there anything remaining on this PR other than writing unit tests?  I do plan on getting around to doing tests (especially since a couple companies have reached out to me directly about this feature) but just want to make sure that otherwise things are good.

HitmanInWis:
Ok.  I do want to get confirmation on whether or not this is a "go" before I invest time into a bunch of unit tests, so best we work through these questions now rather than later.

I can say the I've had two separate companies reach out to me specifically about the feature, and have heard from someone at Adobe they were getting all sorts of requests for it, so there's definitely a high level of interest in the feature.

Even though the feature is not yet currently perfect, I believe it provides a TON of developer value if nothing else, as it can be used to sync assets from a non-prod server to their local instance to have a full working local site.  Getting the feature in at its current state will then allow for future iterations to make it even better.

I'm less concerned about the JSON note transformation issue because a) it seems to be working well even if not 100% perfect and b) the server that is using remote assets is not the system of record, so a 99% perfect copy of meta data is probably going to work just fine nearly 100% of the time.

kaushalmall:
I had some time to play around as well. Not as much as @davidjgonzalez though. My opinion is to merge this, but document all the _drawbacks_ and implementation details that we have discussed here. That ways anyone who decides to use this feature knows what they are getting and if they see issues, hopefully there will be more PRs in the future.

badvision:
@HitmanInWis AEM 6.4 now stores tags in /content/cq:tags but using this path results in an error during sync: java.net.URISyntaxException: Illegal character in scheme name at index 0: 10.51.34.172:4502/content/cq:tags.1.json

badvision:
@HitmanInWis The config should clarify if URL scheme is required or not.

badvision:
Also, this doesn't synchronize even the smaller 140x140 thumbnail images?

badvision:
Also, remote assets should not be editable because that would lead to a lot of confusion.

HitmanInWis:
Just a FYI, we are looking to get someone internally assigned to banging out these tests in the (hopefully) near future.

PR #1295, opened 2018-03-27
fixing the recurssive call for S3 Asset upload.

catchvhebbar:
This is to enable to recursively upload the Assets from S3. Instead of calling ImportAssets, call was being made to CreateFolders. 

First the S3 would always give only 1K objects in the getObjectList, which was getting processed as part of the ImportAssets. Once this list was exhausted, the next set of list was requested and the same ImportAssets method was supposed to be called. Instead the call was being made to CreateFolder. The correction is been made.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 
 ### Fixed
 
+- #1291 - S3 Asset Ingestor stops after 1000 Assets attempting to be imported
 - #1286 - Error page handler now verifies parent resource is not a NonExistingResource
 - #1288 - Restrict the redirect map file upload to .txt file extension
 
--- a/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/S3AssetIngestor.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/S3AssetIngestor.java
@@ -153,7 +153,7 @@ public class S3AssetIngestor extends AssetIngestor {
             }
         });
         if (listing.isTruncated()) {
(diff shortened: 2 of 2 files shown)


badvision:
I'm working on a refactor branch for the asset ingest tooling and will find a way to incorporate these changes in there.  I'll probably close this pull request without merging into the main branch directly if that's ok with you.

catchvhebbar:
Hi, I have a production cutover in 3 days. It would be of great help if this fix can be moved in with a release. Not sure if this will be faster than your refactoring.

justinedelson:
@badvision given the scope of this change, I'd suggest merging it. Granted, I'm partly saying this because this was caused by a dumb oversight on my part.

badvision:
Fair enough.  I still need to merge this into my branch which is getting a PR once I recover from this week.

catchvhebbar:
@justinedelson / @badvision , can we know as to when the release is planned for this. We have to migrate assets sizing in TB's, it would be of great help if this comes in as a release.

justinedelson:
@catchvhebbar we should do a 3.15.0 release this week

badvision:
Ok yeah that is in my feature branch.  I didn't realize it was such a small patch.  I'll finish my feature branch today so we can get all these improvements into 3.15.0. 👍

justinedelson:
@catchvhebbar I ended up cherry picking some bug fixes and releasing a 3.14.12. You can find it here: https://github.com/Adobe-Consulting-Services/acs-aem-commons/releases/tag/acs-aem-commons-3.14.12 and it should be on repo.adobe.com in a day or two (at which point I'll update the website). Thanks!

catchvhebbar:
Thanks a ton @justinedelson for giving the release.
Appreciate it.

Issue #16, opened 2013-10-08
service.vendor and service.description services properties should be set automatically


justinedelson:
Would be best not to reproduce these all over the place.

PR #1302, opened 2018-04-02
Feature/1279 folder import tool

badvision:
- New MCP tool for importing nodes/metadata/folders from XLSX spreadsheet
- New MCP tool for importing assets listed in XLSX spreadsheet, can be URLs or local file paths
- Major refactoring for existing asset import utilities
- Added "Dry Run" and "Detailed Report" features to all asset ingest utilities


bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/FileOrRendition.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.TreeMap;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Abstraction of a file which might be an asset or a rendition of another asset
+ */
+public class FileOrRendition extends HashMap<String, String> implements HierarchialElement {
```
justinedelson: This isn't really critical since this class isn't API, but it would be better to *not* extend HashMap here. See Effective Java, Item #16

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/AssetIngestor.java
```diff
@@ -45,39 +46,63 @@
...
-    String jcrBasePath;
+    String jcrBasePath = "/content/dam";
     @FormField(
             name = "Ignore folders",
             description = "List of folder names to be ignored",
             hint = "tmp,.DS_STORE",
             options = {"default=tmp,ds_store,.ds_store,.thumbs,.appledouble"}
     )
-    String ignoreFolders;
+    String ignoreFolders = "tmp,ds_store,.ds_store,.thumbs,.appledouble";
```
justinedelson: Why is this (and the others like it) duplicated now? Is there a risk of getting out of sync?

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/Folder.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Override
+    public String getItemName() {
+        return name;
+    }
+
+    @Override
+    public AssetIngestor.Source getSource() {
+        throw new UnsupportedOperationException("Not supported yet."); //To change body of generated methods, choose Tools | Templates.
```
justinedelson: A bit of IDE-generated noise here.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/FileOrRendition.java
```diff
@@ -0,0 +1,158 @@
...
+        return getName();
+    }
+
+    @Override
+    public Source getSource() {
+        if (fileSource == null) {
+            FileOrRendition thizz = this;
+            fileSource = new Source() {
+                Long size = null;
+                URLConnection connection = null;
```
justinedelson: *important* please use HttpClient here, not URLConnection, so that the proxy configuration from AEM is leveraged. Also, there should be configurable timeouts set (both connect and socket).

justinedelson: Also, I suspect we need a close() method on the Source interface to ensure that the connection gets cleaned up when the source is not needed anymore.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/AssetIngestor.java
badvision: It's because there are two ways the process will be instantiated: either through reflection with the servlet which looks @ the annotations, or directly by either user code or via a unit test.  Direct instantiation won't address those annotation values, so it's better to assign them to get consistent behavior between unit test and runtime.  We can open a feature request to close that gap in the default init method if you'd like.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/FileOrRendition.java
badvision: Actually URLConnection is kinda desired in case it is not an HTTP URL.  I'll add a switch block in there to use HTTP Client for cases where it is an HTTP Url, and URLConnection otherwise (for file:// or ftp://)

justinedelson: That raises some more questions :) how would proxy support work for ftp? just document as a known issue.

What about timeouts for FTP?

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/AssetIngestor.java
justinedelson: ok. makes sense.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/UrlAssetImport.java
```diff
@@ -0,0 +1,312 @@
...
+    public static String CONTENT_BASE = "/content";
+    public static String UNKNOWN_TARGET_FOLDER = "/content/dam/unsorted";
+
+    private static final Logger LOG = LoggerFactory.getLogger(UrlAssetImport.class);
+    private HttpClient httpClient = null;
+
+    public UrlAssetImport(MimeTypeService mimeTypeService, HttpClientBuilderFactory httpFactory) {
+        super(mimeTypeService);
+        if (httpFactory != null) {
+            this.httpClient = httpFactory.newBuilder().build();
```
justinedelson: I think this is where the timeouts need to be set.

badvision: I can make it a configuration or something.  WDYT?

justinedelson: configuration with defaults. See https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/http/impl/HttpClientFactoryImpl.java#L85

badvision:
```RequestConfig requestConfig = RequestConfig.custom()
                .setConnectTimeout(connectTimeout)
                .setSocketTimeout(soTimeout)
                .build();```

Got it... yeah that's what I did and moved this to init() after the timeout variable has populated from the deserialization logic.

Issue #1303, opened 2018-04-02
JUnit tests for HTTP Cache fail under Windows


badvision:
```Running com.adobe.acs.commons.httpcache.store.jcr.impl.visitor.TotalCacheSizeVisitorTest
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.275 s <<< FAILURE! - in com.adobe.acs.commons.httpcache.store.jcr.impl.visitor.TotalCacheSizeVisitorTest
test(com.adobe.acs.commons.httpcache.store.jcr.impl.visitor.TotalCacheSizeVisitorTest)  Time elapsed: 0.993 s  <<< FAILURE!
java.lang.AssertionError: expected:<1950> but was:<2010>
    at com.adobe.acs.commons.httpcache.store.jcr.impl.visitor.TotalCacheSizeVisitorTest.test(TotalCacheSizeVisitorTest.java:32)
```

Need to update unit test to pass regardless of what platform they run on.

PR #1304, opened 2018-04-02
Added better handling of differently-sized line endings

badvision:
- Updated unit test to be more platform-agnostic


badvision:
This is for issue #1303

Issue #1276, opened 2018-03-07
Issue starting ACS Commons bulk workflow with smart tag workflow


badvision:
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.4 L21
* [x] ACS AEM Commons Version: 3.11+
* [x] Reproducible on Latest? yes

### Expected Behavior

I would really like to run bulk workflow for smart tags.

### Actual Behavior

It acts like it's starting, creates an action manager (using multi-threaded mode with Query Builder) and it just dies.  Found this error in the log:

```07.03.2018 23:29:27.623 *ERROR* [192.150.9.200 [1520465367403] POST /etc/acs-commons/bulk-workflow-manager/smart-tags-for-celium/_jcr_content.start.json HTTP/1.1] com.adobe.granite.workflow.core.jcr.WorkflowModelManager Failed to check for the existance of a overlaid workflow model with ID /var/workflow/models//var/workflow/models/dam/dam-autotag-assets, continuing with original model ID /etc/workflow/models//var/workflow/models/dam/dam-autotag-assets/jcr:content/model
javax.jcr.RepositoryException: Invalid name or path: /var/workflow/models//var/workflow/models/dam/dam-autotag-assets
	at org.apache.jackrabbit.oak.jcr.session.SessionContext.getOakPathOrThrow(SessionContext.java:375) ```

justinedelson:
is that the full stack trace?

davidjgonzalez:
+1 to adding the full stack trace - took a look at the synth code and not clear where that would be happening (thought there are static references to the pre-6.4 /etc/workflow/models that needs adjusting; but are not this issue)

badvision:
I'm pretty sure it's whatever converts that numeric value into a workflow path.  The UI doesn't have the workflow paths as selection values, so it's whatever happens on the other side of the POST of the setup part, I think.  Probably affects synthetic and plain workflow the same way because neither of them worked.

davidjgonzalez:
Fixed via  #1306

PR #1306, opened 2018-04-03
Feature/1276 workflow fixes

badvision:
This fixes the issue reported in #1276 as well as the user-event-data bug too.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/workflow/synthetic/impl/SyntheticWorkflowModelImpl.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/workflow/synthetic/impl/SyntheticWorkflowModelImpl.java
@@ -59,12 +59,17 @@ public class SyntheticWorkflowModelImpl implements SyntheticWorkflowModel {
             throw new WorkflowException(ex.getMessage(), ex);
         }
 
-        if (!StringUtils.endsWith(modelId, WORKFLOW_MODEL_PATH_SUFFIX)) {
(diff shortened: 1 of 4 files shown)


bundle/src/main/java/com/adobe/acs/commons/workflow/synthetic/impl/SyntheticWorkflowModelImpl.java
```diff
@@ -33,11 +33,18 @@
...
 import java.util.List;
 import java.util.Map;
+import javax.jcr.RepositoryException;
 
 public class SyntheticWorkflowModelImpl implements SyntheticWorkflowModel {
     private static final Logger log = LoggerFactory.getLogger(SyntheticWorkflowModelImpl.class);
 
-    private static final String WORKFLOW_MODEL_PATH_PREFIX = "/etc/workflow/models/";
+    private static final String[] WORKFLOW_MODEL_PATH_PREFIXES = new String[]{
+        "",
```
davidjgonzalez: This feels a bit random ...  which models do we need? 

In 6.4 - If they aren't in `/var/workflow/models` can they even be run? (Wouldnt that mean they aren't synch'd and there is nothing to operate against?)

If thats not the case, and an unsynched model is sufficient, we probably should first check:

1. /conf/global/settings/workflow/models
2. /apps/settings/workflow/models
3. /libs/settings/workflow/models

to follow CA Config look-up.

And then finally fallback to /etc/workflow/models (which per my first point on published models seems weird, since IIRC < 6.4 also synch'd them out somewhere).

PR #1317, opened 2018-04-04
#1265 - Set default Replicated At and Replicated By values when the p…

### LINKED ISSUE #1265: Package Replication Status Updater: Add workaround for bug in JcrPackageDefinition.getCreated() with maven-plugin generated packages
Due to https://issues.apache.org/jira/browse/JCRVLT-276 packages being created from the `filevault-package-maven-plugin` or its predecessor the `content-package-maven-plugin` never have a proper created date set. It would be useful to add a workaround to also parse the date format created by those plugins.

davidjgonzalez:
…arameterized values are null in ReplicationStatusManagerImpl to prevent NPEs

@kwin  can you take a quick look at this? It should resolve #1265 but you're a bit closer to the problem so would be great if you can double-check.


davidjgonzalez:
@kwin good idea - added more informative logging.

PR #1332, opened 2018-04-16
Feature/httpcache jcrstorage bugfixes

Sc0rpic0m:
I thought I tested this before. But now when I run the code on a fresh 6.3 instance, I get an AccessDenied exception. the jcr:all seems necessary on the system user for the HTTPCache JCR storage.


davidjgonzalez:
Looks like the issue is the rep:read/rep:write should be on /var/acs-commons/httpcache and not the /root .. fixing via new PR

Issue #1338, opened 2018-04-21
Asset ingestion for non-administrators


badvision:
Some managed services admins prefer not to add any content users to the administrators group.  In such cases, the asset ingestion should look for an optional bulk-content-loader (or similarly-named) group and check if a user is in that group, or in the administrators group.  This gives the customer some flexibility to manage their own ingestion without requiring administrator group membership.

PR #1339, opened 2018-04-22
#1338 - Added refactored process definition factory for auth checks, …

badvision:
…broadened allowed groups for asset ingestion

This is a small refactoring, but wanted to run the idea by you since it supposes group names which might not exist (but it's not going to fail or throw errors if those groups do not exist)


bundle/src/main/java/com/adobe/acs/commons/mcp/AdministratorsOnlyProcessDefinitionFactory.java
```diff
@@ -20,36 +20,17 @@
...
-                        try {
-                            return g.getID().equals("administrators");
-                        } catch (RepositoryException e) {
-                            return false;
-                        }
-                });
-        } catch (RepositoryException e) {
-            return false;
-        }
+    public String[] getAuthorizedGroups() {
```
justinedelson: this method should probably be final. Also it is missing the @Override annotation AFAICT

bundle/src/main/java/com/adobe/acs/commons/mcp/AuthorizedGroupProcessDefinitionFactory.java
```diff
@@ -0,0 +1,66 @@
...
+                                } catch (RepositoryException ex) {
+                                    return false;
+                                }
+                            });
+        } catch (RepositoryException e) {
+            return false;
+        }
+    }
+
+    public abstract String[] getAuthorizedGroups();
```
justinedelson: can be protected scope

bundle/src/main/java/com/adobe/acs/commons/mcp/AdministratorsOnlyProcessDefinitionFactory.java
badvision: It has the annotation but making it final makes sense.

bundle/src/main/java/com/adobe/acs/commons/mcp/AuthorizedGroupProcessDefinitionFactory.java
badvision: That is true in a few other places too.  Pushing an update for access modifiers.

bundle/src/main/java/com/adobe/acs/commons/mcp/AuthorizedGroupProcessDefinitionFactory.java
```diff
@@ -0,0 +1,66 @@
...
+ * groups.
+ *
+ * @param <P> Process definition generated by this factory
+ */
+@ConsumerType
+public abstract class AuthorizedGroupProcessDefinitionFactory<P extends ProcessDefinition> extends ProcessDefinitionFactory<P> {
+
+    @Override
+    @SuppressWarnings("squid:S1141")
+    public boolean isAllowed(User user) {
```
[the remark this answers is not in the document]
badvision: I refactored the existing administrator implementation to use this and also added additional groups allowed to use the asset ingest tools.

davidjgonzalez:
@badvision you planning on fixing JE's 2 comments?

badvision:
Yes sir, already done.

Issue #1278, opened 2018-03-12
Extract code into a method in EvolutionContext


etugarev:
https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/062b9001b7602093cf76b4e812db4b3e6b584f96/bundle/src/main/java/com/adobe/acs/commons/version/impl/EvolutionContextImpl.java#L79

I would suggest that we have a separate method in the context to get the current evolution instead of adding it to the versions list - the code which is comparing last and previous version based on status info won't be able to detect any changes since CurrentEvolutionImpl uses CurrentEvolutionEntryImpl and CurrentEvolutionImpl.getStatus() is always an empty string.

Example:

` if (StringUtils.isNotEmpty(evolutionEntry.getStatus())) {   
          evolutions.put(i, getEvolutionalChange(i, evolutionEntry, previousEvolution));
}`

This code won't be able to fill evolutions list.

etugarev:
Sure I can do it, I wanted a someone to have a quick look on the problem before I start with changes. I would say it's not obviously a bug, but a change in logic which breaks existing code.

PR #1342, opened 2018-04-25
Defect/acs commons version/evolution context

etugarev:
Extract code into a method in EvolutionContext #1278 

- initially I planned to create a property "currentEvolution" which should represent CurrentEvolutionImpl(). But this would also mean a massive JSP change which I wanted to avoid.

- getEvolutionItems() logic remains intact and new method getVersions() was introduced to contain only a version history.

- added test


bundle/src/main/java/com/adobe/acs/commons/version/package-info.java
```diff
@@ -20,7 +20,7 @@
 /**
  * Version services and helpers.
  */
-@Version("1.0.0")
+@Version("2.0.0")
```
justinedelson: This looks wrong -- we shouldn't need to increment the major version. I suspect that EvolutionContext is missing the `@ProviderType` annotation.

bundle/src/test/java/com/adobe/acs/commons/version/impl/EvolutionContextImplTest.java
```diff
@@ -0,0 +1,187 @@
+package com.adobe.acs.commons.version.impl;
```
justinedelson: missing license header

bundle/src/main/java/com/adobe/acs/commons/version/package-info.java
badvision: Oh, good catch!   Yeah, @ProviderType is definitely a good idea, because then adding a method would not require a major version increment like this.

bundle/src/test/java/com/adobe/acs/commons/version/impl/EvolutionContextImplTest.java
badvision: Can we add a check for this in code climate?

justinedelson: Maybe. Can you create a separate issue for that and assign it to me. I'll take a look.

bundle/src/main/java/com/adobe/acs/commons/version/package-info.java
etugarev: I'll check it now, without major version update and without @providertype there is a en error from maven baseline plugin -> version increase required 1.0.0 -> 2.0.0

justinedelson:
Thanks!

PR #1344, opened 2018-04-25
Make "Ensure Oak Index" Felix plugin URL match the sample provided

tml:
Copied other values from `EnsureOakIndexManagerImpl.java`, because they seemed like reasonable defaults.


justinedelson:
Thanks!

PR #1347, opened 2018-04-27
Redirect Manager Enhancement: Entry Editor

klcodanr:
I added a new feature based on user feedback from the initial version of the tool. The enhancement is a new tab allowing users to edit the redirect map entries directly inside the tool:

screen shot 2018-04-27 at 5 01 36 pm

It also evaluates all of the entries now, so it will highlight invalid entries from the file as well as from the Page / Asset properties.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/redirectmaps/impl/AddEntryServlet.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/redirectmaps/impl/AddEntryServlet.java
@@ -1,0 +1,89 @@
+/*
+ * #%L
+ * ACS AEM Commons Bundle
+ * %%
+ * Copyright (C) 2013 Adobe
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+package com.adobe.acs.commons.redirectmaps.impl;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+
+import javax.servlet.ServletException;
+
+import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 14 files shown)


klcodanr:
@justinedelson or @davidjgonzalez -- Sorry to pester, would you mind taking a look at this PR? Trying to get it in to support a client.

bundle/src/main/java/com/adobe/acs/commons/redirectmaps/impl/AddEntryServlet.java
```diff
@@ -0,0 +1,76 @@
...
+    protected void doPost(SlingHttpServletRequest request, SlingHttpServletResponse response)
+            throws ServletException, IOException {
+        log.trace("doPost");
+
+        int idx = Integer.parseInt(request.getParameter("idx"), 10);
+        String source = request.getParameter("source");
+        String target = request.getParameter("target");
+        log.debug("Removing entry with {} {} at {}",source, target, idx);
+
+        InputStream is = request.getResource().getChild(RedirectMapModel.MAP_FILE_NODE).adaptTo(InputStream.class);
```
justinedelson: unclosed stream?

bundle/src/main/java/com/adobe/acs/commons/redirectmaps/impl/RemoveEntryServlet.java
```diff
@@ -0,0 +1,80 @@
...
+    private static final Logger log = LoggerFactory.getLogger(RemoveEntryServlet.class);
+
+    protected void doPost(SlingHttpServletRequest request, SlingHttpServletResponse response)
+            throws ServletException, IOException {
+        log.trace("doPost");
+
+        int idx = Integer.parseInt(request.getParameter("idx"), 10);
+        log.debug("Removing index {}", idx);
+
+        InputStream is = request.getResource().getChild(RedirectMapModel.MAP_FILE_NODE).adaptTo(InputStream.class);
```
justinedelson: unclosed stream?

bundle/src/main/java/com/adobe/acs/commons/redirectmaps/models/package-info.java
```diff
@@ -17,5 +17,5 @@
  * limitations under the License.
  * #L%
  */
-@aQute.bnd.annotation.Version("3.14.4")
+@aQute.bnd.annotation.Version("4.0.0")
```
justinedelson: Maybe we need to add a comment here that this package is only meant for use by ACS AEM Commons internals. We really need an `@VisibleForScripting` annotation like Guava has `@VisibleForTesting`. But we have a few places where this is true, so no need to necessarily deal with it right now.

justinedelson:
Thanks!

Issue #1346, opened 2018-04-26
Better support for data types in Spreadsheet-based tools (Data import, URL asset import, etc)


badvision:
## Required Information

* [x] AEM 6.3+
* [x] ACS AEM Commons 3.14+
* [x] Reproducible on Latest: *yes*

### Expected Behavior

I would like to add better support for:
- non-string values via hinting in the column names
- Multi-values using either configurable delimiter in a single column with a hint, or using multiple columns that share the same name.

### Actual Behavior

Currently these tools only support String and String[] values, the later is only used if the strings are able to be split using a hard-coded delimiter (comma)

PR #1352, opened 2018-05-02
Feature/sheet data handling improvements

badvision:
#1346 - Add support for data type hints and multi-value arrays.  This improves URL Asset Import and Data Import as well, but other tooling can use this spreadsheet utility as well.


bundle/src/main/java/com/adobe/acs/commons/mcp/util/Variant.java
```diff
@@ -0,0 +1,271 @@
...
+import java.util.Optional;
+import java.text.ParseException;
+import org.apache.commons.lang.StringUtils;
+import org.apache.poi.ss.usermodel.Cell;
+
+/**
+ * Used to represent values that might be provided as one type but used as another. Avoids glue code and switch
+ * statements in other parts of the code expecially dealing with data from spreadsheets.
+ */
+public class Variant {
```
justinedelson: Should be final I think. And marked as @ProviderType (that should cause the package export to only be minor).

bundle/src/main/java/com/adobe/acs/commons/mcp/util/CompositeVariant.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Represents a value which could be either a list of variants or a single variant. The idea is that this supports a
+ * transition from a singular to a multi-value property for easier conversion.
+ *
+ * @param <T> Can be of any class supported by Variant, used for direct conversion in getValue, etc.
+ */
+public class CompositeVariant<T> {
```
justinedelson: should be final and/or @ProviderType

bundle/src/main/java/com/adobe/acs/commons/mcp/util/Variant.java
```diff
@@ -0,0 +1,271 @@
...
+
+    private static final DateFormat STANDARD_DATE_FORMAT = SimpleDateFormat.getDateTimeInstance();
```
justinedelson: Should use FastDateFormat here for thread safety.

justinedelson: Also, all these date instances are locale-specific. Is that the intention?

bundle/src/main/java/com/adobe/acs/commons/mcp/util/package-info.java
```diff
@@ -17,7 +17,7 @@
  * limitations under the License.
  * #L%
  */
-@Version("3.11.0")
+@Version("4.0.0")
```
justinedelson: see other comments on how this can be potentially avoided.

bundle/src/main/java/com/adobe/acs/commons/mcp/util/CompositeVariant.java
justinedelson: and actually... does this need to be part of the API? Seems like it is only used internally.

bundle/src/main/java/com/adobe/acs/commons/mcp/util/Variant.java
justinedelson: Same question about whether or not this needs to be API

davidjgonzalez: The packaging implies it would/should be used in context of MPC tools — you’re going to create MPC tooling outside of commons that use this? 

Perhaps this should be moved to a more generic package space if you plan on generally using it as a dep outside commons?

bundle/src/main/java/com/adobe/acs/commons/mcp/util/package-info.java
badvision: I set the provider type annotations but it still is requiring the major bump (unfortunately) -- any other ideas?

bundle/src/main/java/com/adobe/acs/commons/mcp/util/Spreadsheet.java
```diff
@@ -191,23 +204,135 @@ public int getRowCount() {
...
     /**
      * @return the dataRows
+     *
+     * @deprecated use getDataRowsAsCompositeVariants
      */
+    @Deprecated
     public List<Map<String, String>> getDataRows() {
+        return dataRows.stream().map(m -> {
+            return m.entrySet().stream().collect(Collectors.toMap(e -> e.getKey(), e ->  e.getValue().toString()));
+        }).collect(Collectors.toList());
```
justinedelson: Note that this change isn't strictly backwards compatible. It is *binary* compatible, but the resulting object was originally mutable and is now not guaranteed to be mutable. In general, it would be best to not return mutable objects from methods like this, so I'm not sure if that behavior needs to be guaranteed or not (in which case `Collectors.toCollection(Supplier)` can be used.

content/src/main/.brackets.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
badvision: Done!  Thanks!

bundle/src/main/java/com/adobe/acs/commons/mcp/util/Spreadsheet.java
badvision: Sometimes I use supplier because I want specific behavior.  For example LinkedHashSet maintains the order as-inserted and that tends to provide a lot of advantages if you want a map that actually retains a guaranteed order.

badvision:
Thanks for taking the time to review this @justinedelson!  Let me know if anything else looks funky, other than my normal brand of funky coding style.  ;)

bundle/src/main/java/com/adobe/acs/commons/mcp/util/Variant.java
badvision: I'm all for it -- Spreadsheet parsing and data handling is a more generic concern than MCP.   Any suggestions?

badvision: commons.data?

PR #1353, opened 2018-05-03
replacing folder node name default character with hyphen

kmurugulla:
@badvision 
Modified code to create folder node names with "-" as separator instead of "_". Please review.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/AssetIngestor.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/AssetIngestor.java
@@ -438,7 +438,7 @@ public abstract class AssetIngestor extends ProcessDefinition {
                 String extension = StringUtils.substringAfterLast(name, ".");
                 return JcrUtil.createValidName(baseName) + "." + JcrUtil.createValidName(extension);
             } else {
-                return JcrUtil.createValidName(name);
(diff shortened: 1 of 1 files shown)


bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/AssetIngestor.java
```diff
@@ -438,7 +438,7 @@ default String getNodeName() {
                 String extension = StringUtils.substringAfterLast(name, ".");
                 return JcrUtil.createValidName(baseName) + "." + JcrUtil.createValidName(extension);
```
davidjgonzalez: why wouldn't this instance of createValidName use the same convention?

davidjgonzalez:
I think this makes sense - as hyphens are the common replacement for spaces in AEM node names.
That being said, I'm not sure if this would be problematic/annoying for customers that have ingested a ton of assets using the old format?

An option to select the valid naming scheme seems overkill to support the old behavior, but... idk. WDYT? @badvision ?

davidjgonzalez:
Is it right hat only one instance of getValidName uses this rule? (See my first comment)

badvision:
Oh I didn't catch that @davidjgonzalez  -- @kmurugulla can you update the conversion for all parts of the path so it's more consistent?

PR #1354, opened 2018-05-04
added jmx bean to on-deploy scripts.  

shsteimer:
makes it easier to see the status of all current scripts and makes re-execution easier.

As I've started to play with on-deploy scripts on my current project, I've found myself wishing there was an easier way to see the on-deploy scripts, their status, etc.  Also added a way to re-execute, and to force re-execution of previously succeeded script via force.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutor.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutor.java
@@ -25,9 +25,6 @@ import javax.management.openmbean.TabularDataSupport;
 import com.adobe.granite.jmx.annotation.Description;
 import com.adobe.granite.jmx.annotation.Name;
 
-import aQute.bnd.annotation.ProviderType;
-
-@ProviderType
 @Description("ACS AEM Commons - On Deploy Script Executor MBean")
 public interface OnDeployExecutor {
 
@@ -36,6 +33,6 @@ public interface OnDeployExecutor {
 
     @Description("Execute the script, given it's fully-qualified class name.  If force==true, script is executed even if it has previously succeeded.")
     void executeScript(@Name(value = "scriptName") String scriptName,
-                    @Name(value = "force") boolean force);
+ @Name(value = "force") boolean force);
 
 }
--- a/bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutor.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutor.java
@@ -1,0 +1,41 @@
+/*
+ * #%L
+ * ACS AEM Commons Bundle
+ * %%
+ * Copyright (C) 2016 Adobe
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutor.java
```diff
@@ -0,0 +1,41 @@
...
+
+import javax.management.openmbean.OpenDataException;
+import javax.management.openmbean.TabularDataSupport;
+
+import com.adobe.granite.jmx.annotation.Description;
+import com.adobe.granite.jmx.annotation.Name;
+
+import aQute.bnd.annotation.ProviderType;
+
+@ProviderType
```
justinedelson: Don't need @ProviderType since this interface isn't exported.

bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutorImpl.java
```diff
@@ -78,11 +94,40 @@
...
+        } catch (OpenDataException ex) {
+            logger.error("Unable to build MBean composite types", ex);
+        }
+    }
+
     @Reference
     private ResourceResolverFactory resourceResolverFactory;
+    @Reference(name = "scriptProvider", referenceInterface = OnDeployScriptProvider.class,
+            cardinality = ReferenceCardinality.MANDATORY_MULTIPLE, policy = ReferencePolicy.DYNAMIC)
+    private List<OnDeployScriptProvider> scriptProviders = new ArrayList<>();
```
justinedelson: I think this needs to be a threadsafe list.

bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutor.java
```diff
@@ -0,0 +1,41 @@
...
+@Description("ACS AEM Commons - On Deploy Script Executor MBean")
+public interface OnDeployExecutor {
+
+    @Description("Scripts")
+    TabularDataSupport getScripts() throws OpenDataException;
+
+    @Description("Execute the script, given it's fully-qualified class name.  If force==true, script is executed even if it has previously succeeded.")
+    void executeScript(@Name(value = "scriptName") String scriptName,
```
justinedelson: Would it make sense to return a boolean here indicating whether or not the script was run?

shsteimer:
@justinedelson I'll take another pass based on your feedback later on.  re: formatting, is there a standard formatter you use to keep things standard?  The issue is that I have a habit of reformatting my code as I go, and I think I'm just using the intelliJ default format, so that's whats causing all those changes.  I can try to clean it up, but would be a lot easier if I could just set my IDE formatter to be the same as the existing one.

justinedelson:
TBH, I've never gotten IntelliJ or Eclipse's auto formatters to work correctly. And personally, with the exception of tabs, I've long given up on consistent formatting. What's more important IMO is to not change the formatting on code you aren't changing.

That said, if someone wanted to contribute format rules for IntelliJ & Eclipse (and then let @badvision deal with NetBeans 😄 ) and reformat all the code, I'd be OK with that. But such a pull request should *only* be about formatting, not functional changes.

Or better yet, figure out how to use http://editorconfig.org/.

HitmanInWis:
Thanks for the contrib!

PR #1356, opened 2018-05-08
Revert "replacing folder node name default character with hyphen"

badvision:
Reverts Adobe-Consulting-Services/acs-aem-commons#1353


badvision:
Need to address one additional thing.

PR #1357, opened 2018-05-08
Feature/asset ingestor foldername with hyphen

kmurugulla:
modified two other areas
1. replace the separator in asset name node with "-"
2. folder title property is moved to set on jcr:content node , to be consistent with OOTB
( Title was set on sling:Folder node, earlier)
3. Added "preserveFileName" option to ingestor form

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/AssetIngestor.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/AssetIngestor.java
@@ -303,7 +303,7 @@ public abstract class AssetIngestor extends ProcessDefinition {
         if (s.nodeExists(folderPath)) {
             Node folderNode = s.getNode(folderPath);
             Node folderContentNode = folderNode.hasNode(JcrConstants.JCR_CONTENT) ? folderNode.getNode(JcrConstants.JCR_CONTENT) : null;
-            if (folderPath.equals(jcrBasePath) || (null != folderContentNode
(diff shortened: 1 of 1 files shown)


bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/AssetIngestor.java
```diff
@@ -317,7 +329,11 @@ protected boolean createFolderNode(HierarchialElement el, ResourceResolver r) th
         trackDetailedActivity(el.getNodePath(), "Create Folder", "Create folder", 0L);
         incrementCount(createdFolders, 1L);
         if (!folderPath.equals(jcrBasePath)) {
-            child.setProperty(JcrConstants.JCR_TITLE, name);
```
badvision: Refactor the content of this as something like "setTitle"

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/AssetIngestor.java
```diff
@@ -294,10 +302,14 @@ protected boolean createFolderNode(HierarchialElement el, ResourceResolver r) th
         Session s = r.adaptTo(Session.class);
         if (s.nodeExists(folderPath)) {
             Node folderNode = s.getNode(folderPath);
```
badvision: Refactor the body of this as updateExistingFolderTitle or something like that

Issue #1363, opened 2018-05-09
HTTP Cache - JCR Cache throws Access Denied exception


davidjgonzalez:
JCR HTTP Cache does not have proper permissions to manage the cache under /var/acs-commons/httpcache .. appears the rep:policy is 1 level too far down (is on /var/acs-commons/httpcache/root should be on /var/acs-commons/httpcache)

PR #1364, opened 2018-05-09
#1363 - Corrects permissions allowing HTTP Cache to write to the JCR …

davidjgonzalez:
…Cache space under /var/acs-commons/httpcache


Sc0rpic0m:
Thanks david!!

PR #1365, opened 2018-05-10
Added sling model injector for Shared Component Property values.

HitmanInWis:
Putting this up for review to get feedback.  Will add tests once I get confirmation that nothing looks crazy :D

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/SharedValueMapValueInjectorImpl.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/SharedValueMapValueInjectorImpl.java
@@ -83,6 +83,8 @@ public class SharedValueMapValueInjectorImpl implements Injector {
                         return getSharedProperties(pageRoot, resource).get(name);
                     case GLOBAL:
                         return getGlobalProperties(pageRoot, resource).get(name);
+                    default:
+                        break;
                 }
             }
         }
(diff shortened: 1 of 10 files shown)


bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/SharedValueMapValueInjectorImpl.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @Override
+    public String getName() {
+        return "shared-component-properties-valuemap";
+    }
+
+    @Override
+    public Object getValue(Object adaptable, String name, Type declaredType, AnnotatedElement element, DisposalCallbackRegistry callbackRegistry) {
+        // sanity check
+        if (element.getAnnotation(SharedValueMapValue.class) == null) {
```
HitmanInWis: For some reason this function gets called by with `name = modelFactory` and `name = slingModelFilter`.  I dont know enough about how these injectors work to understand why those values are triggering this code in the first place, but this is my way of early exit when it does happen.

justinedelson: If the intention is to *require* the annotation, this is OK. The idea with injectors is that unless the model developer explicitly defines the source (either through the `@Source` annotation or an injector-specific annotation), all available injectors are checked. In this case, I can see two possible approaches.

1. What you have now -- require the annotation.
2. Don't require the annotation and make sure that this injector is called *after* the regular ValueMap injector (which has a ranking of 2000, per https://sling.apache.org/documentation/bundles/models.html, so you're already good there since this injector's ranking is 4500).

I'd suggest to go with #1 (as you have) as it would be the least surprising *and* the construction of the merged properties has some cost associated with it (not a lot of cost, but some cost).

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/SharedValueMapValueInjectorImpl.java
```diff
@@ -0,0 +1,123 @@
...
+        return null;
+    }
+
+    /**
+     * Get global properties resource for the current resource.
+     */
+    protected ValueMap getGlobalProperties(Page pageRoot, Resource resource) {
+        String globalPropsPath = pageRoot.getPath() + "/" + JcrConstants.JCR_CONTENT + "/" + SharedComponentProperties.NN_GLOBAL_COMPONENT_PROPERTIES;
+        Resource globalPropsResource = resource.getResourceResolver().getResource(globalPropsPath);
+        return globalPropsResource != null ? globalPropsResource.getValueMap() : new ValueMapDecorator(Collections.emptyMap());
```
justinedelson: You might want to change `new ValueMapDecorator(Collections.empty())` to `ValueMapDecorator.EMPTY`. Slightly cheaper.

justinedelson:
Oh, missed that there weren't tests during my review. We're so close to 50% coverage...

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/SharedValueMapValueInjectorImpl.java
HitmanInWis: For sure - didnt realize that exists :)  Will do.

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/SharedValueMapValueInjectorImpl.java
HitmanInWis: Interesting that it checks all annotations.  I never knew or expected that.  Might be best to stick with as-is to avoid unexpected side effects.

HitmanInWis:
No worries - now that you've confirmed that this is basically good to go I'll go back and add some tests.

PR #1366, opened 2018-05-11
Bugfix/1359 retrieve limit properties also from content policies

kwin:


content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/configure-limit-parsys/touchui-limit-parsys.js
```diff
@@ -82,27 +94,19 @@
...
-
-                currentLimit = data[ACS_COMPONENTS_LIMIT];
-
-                var limit = parseInt(currentLimit);
-
-                isWithin = children.length <= limit;
-            });
+        currentLimit = _findPropertyFromDesign(parsysEditable, Granite.author.pageDesign, ACS_COMPONENTS_LIMIT);
+        if (currentLimit === null) {
+            return false;
```
leonardoderks: We might need to return true  in this line instead of false. If no current limit is set for the parsys the insert dialog is not shown.

kwin: Good hint, fixed!

HitmanInWis:
Where can I find documentation on this fancy "Limit Parsys"?  I've always wanted something like this.  Is there an iParsys version that could be used in header/footer?

content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/configure-limit-parsys/touchui-limit-parsys.js
```diff
@@ -34,6 +34,15 @@
     function _findPropertyFromPolicy(editable, design, propertyName) {
         var cell = Granite.author.util.resolveProperty(design, editable.config.policyPath);
 
+        if (!cell || !cell[propertyName]) {
+            // Inherit property also from its parent (if not set in the local policy path)
+            var parent = ns.editables.getParent(editable);
```
leonardoderks: we might need to change "nc" for Granite.author. I don't see where it is defined

kwin: good catch, fixed

kwin:
@justinedelson or @davidjgonzalez Can you have a look at this PR?

kwin:
@HitmanInWis Documentation is still missing (https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io/issues/40) but hopefully I can come up with a PR for that soon.

PR #1380, opened 2018-05-18
Feature/component cloner

brandonmaynard:
This feature is for a new component called the 'Component Cloner'. It gives an author the ability to drop the Component Cloner inside a parsys and then clone an existing component/node-structure as a 1-to-1 match, include all of its children/grandchildren/etc nodes. This allows an author to recreate existing components and only spend time modifying some properties without having to create the entire structure from scratch. This can potentially save a lot of time for authors, depending on how complex the component is, or how many nested components/nodes are included inside of it.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/componentcloner/impl/ComponentClonerServlet.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/componentcloner/impl/ComponentClonerServlet.java
@@ -1,0 +1,114 @@
+/*
+ * #%L
+ * ACS AEM Commons Bundle
+ * %%
+ * Copyright (C) 2018 Adobe
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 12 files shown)


HitmanInWis:
This feature was voted as one of the winners of the 2018 AEM Community Code Contest: https://forums.adobe.com/thread/2490880

davidjgonzalez:
This seems like odd UX to me as well - itL

badvision:
Wouldn't it be a more consistent user experience to add a clone button to the action bar when the user clicks on the component?

davidjgonzalez:
Yeah - in my minds eye it would be more like a how you can create a XF from a set of components ..  and/or this reminds me of the virtual components of yore (not sure if they still work w editable templates)

I can see this being useful across pages, but on a single page it seems like a copy/paste of a component would be as effective? Might be missing something tho..?

sureshcodeit:
This could be a better fit to one of our recently executed projects where the UI can be controlled through margin paddings options in every single components dialog for left, right, top and bottom with small, medium and large values. Once an author adds a component defines its margin values, the component can be cloned **BUT NOT COMPLETELY**, but limited to set of properties on the node. In this case, I would be interested in copying only the margin settings from the component node. This would be really useful layout contaners/column control parent components where the the first child can be cloned to repeat the same margin settings without having to redefine for every instance of the component. Hope this makes sense. 

Please note the highlighted section above -

Issue #1386, opened 2018-05-23
ACS Parsys Placeholder Text feature not working on AEM 6.3


jainankush385:
## Required Information

* AEM Version: 6.3.1.2
* ACS AEM Commons Version: 3.14.8
* Reproducible on Latest:  yes
### Expected Behavior

Parsys Placeholder Text feature not working

### Actual Behavior
`/apps/acs-commons/touchui-widgets/configure-limit-parsys/touchui-limit-parsys.js`
```
function getDesignPath(editable){
        var parsys = editable.getParent(),
            designSrc = parsys.config.designDialogSrc,
            result = {}, param;

        if (designSrc === undefined) {
            return undefined;
        }

        designSrc = designSrc.substring(designSrc.indexOf("?") + 1);

        designSrc.split(/&/).forEach( function(it) {
            if (_.isEmpty(it)) {
                return;
            }
            param = it.split("=");
            result[param[0]] = param[1];
        });

        if (result.content === undefined) {
            return undefined;
        }

        return decodeURIComponent(result.content);
    }
```
`designSrc` not having content attribute.

spauwels:
We have 
ACS AEM Commons Version: 3.14.10
and still a lot of calls to undefined.2.json

badvision:
@jainankush385 can you confirm if the master branch resolves this bug?

jainankush385:
Hi @badvision  sorry for late reply. 
Can you  please provide me aem package of master branch so i can test on my local.
Thanks

jainankush385:
Hi, @badvision tested locally master branch. The feature is still not working.

Thanks

cc: @wimsymons

PR #1391, opened 2018-05-31
Fix Closure Compiler bug in Parsys Placeholder

wimsymons:
The default settings change the minified JS code logic so it exits with the String "undefined" instead of the object undefined causing the JS logic to fail. Probably a bug in the Google Closure Compiler version AEM is using (which is dated, a version of 2016 in our AEM 6.3.1.2). This fixes the undefined.2.json bug #1386

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 
 ## [Unreleased]
 
+### Fixed
+- #1386 - Fixed ajax calls like undefined.2.json when hovering over parsys
+
 ## [3.17.0] - 2018-05-22
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


badvision:
Instead of throwing out the baby with the bathwater, minification is still preferred if possible.  Could you just convince the minifier to leave it alone, e.g. "undefined" -> "un"+"defined" ?

wimsymons:
Minification still happens, only less. /apps/acs-commons/touchui-widgets/configure-parsys-placeholder.min.js goes from 1489 to 2630 bytes. Most likely it's a bug that Adobe needs to fix by upgrading their outdated Google Closure Compiler OSGi bundle (which wraps v20161024) to the latest Google Closure Compiler release (v20180506).

We haven't tested the 'simple' compilationLevel though. Can try that to see if the situation improves.

But what @badvision is saying isn't possible. I can't split the undefined Javascript keyword.

badvision:
OIC, you are trying to get it to leave that keyword alone.  Does the minifier respond more correctly if you declare undefined in the globals hint?

wimsymons:
@badvision I don't know how to do that.

Changing to compilationLevel=simple doesn't fix it, the bug just comes back at that point. (Because compilationLevel=simple is the default AEM setting, as I found on https://helpx.adobe.com/experience-manager/kb/how-to-change-the-minification-engine-for-client-libraries-in-AEM.html

wimsymons:
Another scenario might be to rewrite the JS code to use `!= "undefined"` and `== "undefined"` instead of `!== undefined` and `=== undefined`.

wimsymons:
I went for a third scenario and rewrote the code with a little less return statements. Now it compresses to correct code.

wimsymons:
Should fix #1386.

Now this is approved, will this be included in the next release @badvision ?

PR #1400, opened 2018-06-20
Added AEM 6.3 support for conditional hiding in edit dialogs

RolandDahlemPV:
hiding dialog fields based on condition does not work in AEM 6.3.
This Pull Request fixes the issue

### DIFF
--- a/content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/showhidedialogfields/source/showhidedialogfields.js
+++ b/content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/showhidedialogfields/source/showhidedialogfields.js
@@ -82,6 +82,10 @@
     	if(select){
     		return "select";
     	}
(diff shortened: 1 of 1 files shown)


content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/showhidedialogfields/source/showhidedialogfields.js
```diff
@@ -82,6 +82,10 @@
     	if(select){
     		return "select";
     	}
+	//Check if field is a CoralUI3 checkbox
+	if($(element).hasClass("coral-Checkbox")){
```
justinedelson: Per the comment at the top of https://helpx.adobe.com/experience-manager/6-3/sites/developing/using/reference-materials/coral-ui/coralui3/Coral.Checkbox.html, this class shouldn't be referenced directly. I believe the correct check would be on the element name.

PR #1402, opened 2018-06-20
Added AEM 6.3 support for conditional hiding in edit dialogs

RolandDahlemPV:
hiding dialog fields based on condition does not work in AEM 6.3.
This Pull Request fixes the issue

### DIFF
--- a/content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/showhidedialogfields/source/showhidedialogfields.js
+++ b/content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/showhidedialogfields/source/showhidedialogfields.js
@@ -83,9 +83,10 @@
     		return "select";
     	}
 	    
-	if($(element).hasClass("coral-Checkbox")){
+        //Check if field is a CoralUI3 checkbox
+        if(element && element.tagName==="CORAL-CHECKBOX"){
         	return "checkbox";
-    	}
+        }
     }
 
     // make sure all unselected target elements are hidden.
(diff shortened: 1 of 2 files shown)


content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/showhidedialogfields/source/showhidedialogfields.js
```diff
@@ -82,6 +82,10 @@
     	if(select){
     		return "select";
     	}
+	    
+	if($(element).hasClass("coral-Checkbox")){
```
justinedelson: (commented on wrong PR)
Per the comment at the top of https://helpx.adobe.com/experience-manager/6-3/sites/developing/using/reference-materials/coral-ui/coralui3/Coral.Checkbox.html, this class shouldn't be referenced directly. I believe the correct check would be on the element name.

RolandDahlemPV: @justinedelson:

justinedelson: According to https://stackoverflow.com/questions/8852665/how-can-i-get-name-of-element-with-jquery, you should use attr

RolandDahlemPV: But "./frame" has nothing to do with checkbox. I don't think we can identify a checkbox with this in a reliable way

justinedelson: Sorry, I edited my comment above to clarify that you should look at the element name, not the name property (or attribute). Strictly speaking, in DOM terms this is "tag name".

RolandDahlemPV: I changed it to use tag name

justinedelson:
Thanks!

PR #1406, opened 2018-06-25
allow to blacklist resources by path for replication status updater

### LINKED ISSUE #1334: Package Replication Status Updater does not treat initialContent below editable templates correctly
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.3.1.2
* [x] ACS AEM Commons Version: 3.15.0
* [X] Reproducible on Latest? yes/no

### Expected Behavior

Pages being created from an editable template are not marked as ever being activated (i.e. they should not have a `jcr:content/cq:lastReplicated` property with a value).

### Actual Behavior

The default configuration for the [Package Replication Status Updater](https://adobe-consulting-services.github.io/acs-aem-commons/features/package-replication-status-updater/index.html) will update the replication status below `/conf/.../<mytemplate>/initial/jcr:content` (because this is a node of type `cq:PageContent`), i.e. it will create a property `cq:lastReplicated`.

Pages being created from that editable template afterwards have an incorrect activation status (i.e. they are marked as previously being activated), although they have just been created. This happens because AEM by default copies over all properties from `.../initial/jcr:content` (even including operational properties like `cq:lastReplicated`)

HINT: This is only an issue with the Package Replication Status Updater as manually activating a template via http://localhost:4502/libs/wcm/core/content/sites/templates.html/conf/we-retail will not touch the `jcr:content` node below the `initial` Page of the template, but only the `jcr:content` node directly below the template.

### Steps to Reproduce

1. Configure the Package Replication Status Updater via http://localhost:4502/system/console/configMgr/com.adobe.acs.commons.replication.status.impl.JcrPackageReplicationStatusEventHandler (just creating a default configuration is enough)
2. Open the Package "we.retail.ui.content-2.0.2.zip" in the package manager (http://localhost:4502/crx/packmgr/index.jsp#/etc/packages/adobe/aem6/sample/we.retail.ui.content-2.0.2.zip) and click on "Replicate"
3. Create a new page of any of the we-retail templates. 
4. Observe the incorrectly set `cq:lastReplicated` property below the newly created page

Most probably there is another configuration possibility necessary to also exclude resources by  path (from being covered by the Package Replication Status Updater). Just whitelisting by node types is not enough for this edge case.

kwin:
This closes #1334.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 - #1386 - Fixed ajax calls like undefined.2.json when hovering over parsys
 - #1301 - Fixed issue in MCP process forms where CoralUI bindings happened twice per form breaking some functionality (like file uploads).
 
+### Added
+- #1334 - Added blacklist possibility by path regex for replication status updater (to make sure initial content below editable templates never has replication metadata)
+
 ## [3.17.0] - 2018-05-22
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


kwin:
I will instead come up with a PR which fixes both #1407 and #1334.

kwin:
This should be fixed together with #1407

PR #1411, opened 2018-06-28
Feature/package replication status updater path restrictions

kwin:


justinedelson:
@kwin I'm a little concerned about the inclusion of the configuration here. Due to the way the OSGi installer works, when there is a configuration file in ACS AEM Commons, it becomes hard to reliably override it. This is OK for something like the service user mapping since that isn't meant to be changed downstream. Does the queue configuration here fall into the same bucket?

Adobe US in on holiday for the next week, so I'm not sure if @davidjgonzalez will be able to review until the 9th.

kwin:
We can also leave out the configuration (which means we rely on the default queue configuration, which also has 5 retries in my AEM 6.3 instance by default). That way every consumer could decide to configure a custom queue configuration only if really necessary.

bundle/src/main/java/com/adobe/acs/commons/replication/status/impl/JcrPackageReplicationStatusEventHandler.java
```diff
@@ -459,13 +491,34 @@ protected void activate(final Map<String, String> config) throws LoginException
...
         log.info("Package Replication Status - Replicated By Override User: [ {} ]", this.replicatedByOverride);
         log.info("Package Replication Status - Replicated At: [ {} ]", this.replicatedAt.toString());
-        log.info("Package Replication Status - Node Types: [ {} ]",
-                StringUtils.join(this.replicationStatusNodeTypes, ", "));
+        log.info("Package Replication Status - Node Types and Path Restrictions: [ {} ]", pathRestrictionByNodeType);
+    }
+
+    protected Map.Entry<String, Pattern> extractNodeTypeRestrictionAndPathRestrictionFromConfigEntry(String entry) {
+        // split up between node types and paths
+        int separatorPos = entry.indexOf(' ');
```
davidjgonzalez: More just an FYI as this doesn't quite fit your use case.. but we have ParameterUtil which helps w stuff like this (again, don't think it quite fits your use case without modification).

https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/665faaa8beff5f314f0d384ff4d2abbc820f1691/bundle/src/main/java/com/adobe/acs/commons/util/ParameterUtil.java

content/src/main/content/jcr_root/apps/acs-commons/config.author/org.apache.sling.event.jobs.QueueConfiguration-packagereplicationstatus.config
```diff
@@ -0,0 +1,12 @@
+# Configuration created by Apache Sling JCR Installer
```
davidjgonzalez: As JE mentioned, move to docs.

kwin:
I migrated to using the `ParameterUtil` now and removed the config. Please check again @davidjgonzalez .

Issue #1415, opened 2018-07-03
Error Page Handler first real resource look-up fails with etc maps


davidjgonzalez:
## Required Information

* [X] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.x
* [X] ACS AEM Commons Version: 3.x
* [X] Reproducible on Latest? yes

Via @jochen314 

 don't get it: We already have resolved the url to a resource path.
Now we check for the parent.
Why do we need to do a resolve with the request and not just a getResource?

> For us, this does not work:
> We have multiple vhosts on the same instance. So out /etc/map maps:
> http://host1/ -> /content/host1 and
> http://host2/ -> /content/host2
> When the url http//host1/en/not-existing.html:
> the Url get mapped to the NonExistingResource /content/host1/en/not-existing
> At this line, candidatePath is /content/host1/en
> And this gets resolved to /content/host1/content/host1/en, which is not, want we need.
> 
> We need to either:
> 
> * walk up the URL and resolve on each stop or
> * walk up the resolved path and then got a getResource
> Currently is seems to be mixed up

https://github.com/Adobe-Consulting-Services/acs-aem-commons/commit/aab03112d399fc81eb031b5426602db4e12be1a2#diff-730590a3bca3e92d4ed6242cf2e510a5

davidjgonzalez:
Fixed in 3.17.2

jochen314:
Hi,

sorry to bother, but my problem is not fixed.

My problem is in the `for` loop (line 560):
`for (int i = parts.length - 1; i >= 0; i--) {`

`parts` is 

> [content, test, www, global, de, errors, 40dddddd4]

(/content/test/ww/global/de exists)

But as there is a resolve with request, the existing resource is not found. The mapping added another /content/test/www in front.

PR #1418, opened 2018-07-09
#1415 - Fixed bug in Error page handler real-resource lookup when /et…

davidjgonzalez:
…c/map is used.


bundle/src/main/java/com/adobe/acs/commons/errorpagehandler/impl/ErrorPageHandlerImpl.java
```diff
@@ -530,8 +530,14 @@ private Resource findFirstRealParentOrSelf(SlingHttpServletRequest request, Reso
...
-            resource = resourceResolver.resolve(request, path);
-        }
+            // #1415 - First try to get get the resource at the direct path; this look-up is very fast (compared to rr.resolve and often what's required)
+            resource = resourceResolver.getResource(path);
+
+            if (resource == null) {
+                // #1415 - If the resource is not available at the direct path, then try to resolve (handle sling:alias).
+                // First map the path, as the resolve could duplicate pathing.
+                resource = resourceResolver.resolve(request, resourceResolver.map(request, path));
+            }        }
```
bstopp: Whats with the extra curly brace?

ireasor:
Looks good to me!

Issue #1431, opened 2018-07-26
Multifield Extension not compatible with Coral3 Select


Narayan15:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs,  etc: AEM 6.3.2 (6.3 SP2)
* [ ] ACS AEM Commons Version: 3.17.0
* [ ] Reproducible on Latest? yes

### Expected Behavior

When using the deprecated "Multifield Extension" it does not support granite/ui/components/coral/foundation/form/select component. When it's used, the Select field is showing a default value every time rather than displaying the option that I selected earlier.

### Actual Behavior

Although this feature has been deprecated, we can't move to the recommended approach as we require both classic and touch UI to function and currently the classic UI only supports saving the data as json values. We thus need same approach for touch ui, but getting stuck with coral3 select issue.

### Steps to Reproduce
Add granite/ui/components/coral/foundation/form/select as one of the items in the multifield.

### Links

I will be providing a patch to the existing codebase to fix this particular issue.

davidjgonzalez:
Fixed in 3.17.2! thanks @Narayan15

PR #1432, opened 2018-07-26
Fix to bug #1431. Coral3 Select field now works in Composite MultiField

Narayan15:


davidjgonzalez:
Thanks for the fixed @Narayan15 !

Issue #1435, opened 2018-07-26
Increase FAM Watchdog time limit


davidjgonzalez:
When watchdog interrupts writing threads, bad things can happen to the repository. The default for this should be set to something that will never be hit, until we can figure out how to only perform "safe" interrupts (if thats even possible).

This should be back-ported to 2.x.x line as well (assuming FAM is in 2.x.x)

Chatted w tom blackford, and he suggests:

* Update watchdog timeout default to: 1hr
* Decrease default CPU throttle to: 50%
* Increase Mem throttle to: 85%

PR #1440, opened 2018-07-31
#1435 - Updated Throttled Task Runner

davidjgonzalez:
- Update watchdog timeout default to: 1hr
- Decrease default CPU throttle to: 50%
- Increase Mem throttle to: 85%

(Note that the mem throttle was already 85% in the default value, however, the displayed value was .75. These now are in sync with 85%)

Added a WARNING to the OSGi configuration's description about Watchdog timeouts.


badvision:
50% feels kind of low, imho

davidjgonzalez:
We can let Tom comment on this; His suggestion IIUC is based on common customer experience - and he has a unique position of being exposed to many customers (relative to your average bear).

Now only if he has a github account I could @ ...

Issue #1441, opened 2018-07-31
Error Creating a Report Executor


akranjec:
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: AEM 6.4
* [x] ACS AEM Commons Version: 3.17.0
* [x] Reproducible on Latest? yes

### Expected Behavior

According to documentation on Extending Reports it should be possible to create a custom Report Executor.

### Actual Behavior

Custom Report Executor can't be created.

### Steps to Reproduce

1. I used fresh 6.4 AEM instance and set up a fresh project based on AEM archetype 13.
2. Created TestReportExecutor.java class with `@Model` annotation and implemented ReportExecutor interface. 
3. Also, I created a component which uses TestReportExecutor
```
                <reportExecutor
                        jcr:primaryType="nt:unstructured"
                        sling:resourceType="granite/ui/components/foundation/form/hidden"
                        name="./reportExecutor"
                        value="clean.aem.core.models.TestReportExecutor" />
```
4. After using TestReportExecutor in newly created report I recive folowing error in error.log

```
> java.lang.ClassNotFoundException: clean.aem.core.models.TestReportExecutor not found by com.adobe.acs.acs-aem-commons-bundle [546]
> 	at org.apache.felix.framework.BundleWiringImpl.findClassOrResourceByDelegation(BundleWiringImpl.java:1639)
> 	at org.apache.felix.framework.BundleWiringImpl.access$200(BundleWiringImpl.java:80)
> 	at org.apache.felix.framework.BundleWiringImpl$BundleClassLoader.loadClass(BundleWiringImpl.java:2053)
> 	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
> 	at com.adobe.acs.commons.reports.models.ReportRunner.executeConfig(ReportRunner.java:66) [com.adobe.acs.acs-aem-commons-bundle:3.17.0]
> 	at com.adobe.acs.commons.reports.models.ReportRunner.init(ReportRunner.java:118) [com.adobe.acs.acs-aem-commons-bundle:3.17.0]
> 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
> 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
> 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #1445, opened 2018-08-02
#1441 - Fixed issue with the Report Runner loading custom Report Exec…

davidjgonzalez:
…utors

It appears the problem is the class loader being used.

This fix switches it to use the DynamicClassLoaderManager.

@klcodanr i don't love the new cstor, but i couldn't figure out how to inject the DCLM for the tests since there is a cstor that takes a request. I wanted to remove that, and just use the @Self annotation for the Request (im not sure why it doesn't, the only use of the cstor I can find is in the test) but that is an API breaking change :(

The new cstor is default, so at least its not usable as a public API.

If you know how to use mockito to force inject a field w/ a mock that'd be a nice enhancement.


klcodanr:
@davidjgonzalez the only thing I could think to do would be to inject the object into the field with reflection, but it's hack-y.

davidjgonzalez:
@klcodanr no worries; just an FYI you can use the @Mock and @InjectMocks to inject into fields (this is how we normally do it).

davidjgonzalez:
@justinedelson any issues w this?

PR #1451, opened 2018-08-03
Add ns-scoped flags to shared-component-properties function

adamcin:
as workaround for ns.edit.Toolbar behavior which allows registering
multiple instances of the same action name, resulting in an accumulation
of buttons when this script is embedded in multiple clientlibs or when
cq-layer-activated messages are forged by third-party clientlibs.


HitmanInWis:
👍

Issue #1428, opened 2018-07-25
MCP - URL Asset Import - Column names from spreadsheet are imported without retaining case-sensitivity


ksurendra:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: AEM 6.3 CFP 2
* [ ] ACS AEM Commons Version: 3.17.0
* [ ] Reproducible on Latest? Yes

### Expected Behavior
Column names (which will be the metadata keys) imported retaining the case as entered in the spreadsheet.

### Actual Behavior
Column names are converted to lowercase when saved into AEM.

![screen shot 2018-07-25 at 11 06 06 am](https://user-images.githubusercontent.com/902972/43213363-0c0349b8-8ffc-11e8-92d8-0870a8c77da9.png)

### Steps to Reproduce
1. Login to AEM > Tools > ACS Commons > Managed Controlled Process 
2. Choose Start > URL Asset Import
3. Use the sample spreadsheet attached to this bug
4. Run the import
5. Go to AEM Home > Assets > Files > devpoc3 > sxm-apps > others
6. Open any asset and under Basic tab.. the "Alternate Text/Caption" the value will not be shown

7. go to CRX/DE and open this path **/content/dam/devpoc3/sxm-apps/others/004_2.jpg/jcr:content/metadata** and see under metadata we can see that the values are saved as **sxm:alternatetextcaption** 

8. If we enter the same as **sxm:alternateTextCaption** the value shows up on the UI. This is because the **Metadata Schema** has this field defined as **sxm:alternateTextCaption**

### Links
1. Sample import spreadsheet 
[sample.xlsx](https://github.com/Adobe-Consulting-Services/acs-aem-commons/files/2228651/sample.xlsx)

kaushalmall:
@ksurendra mind looking at the issue @msiraj20 is reporting?

ksurendra:
@badvision or @davidjgonzalez can we re-open this issue, i created a pull-request for this at https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/1518

ksurendra:
> This seems to be not working as expected.
> 
> Steps:
> 
> 1. Upload the attached excel in URL asset Import
> 
> ## Actual Result
> 1. Not honoring the case of the property, example in excel the property was "assetType", but in CRX it stored as assettype
[...]

PR #1463, opened 2018-08-21
Updated MCP - DataImporter class to call Spreadsheet class without converting header names to lower case

ksurendra:
This is to fix the issue: https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/1428 

Before the change, Data Importer is calling the `data = new Spreadsheet(importFile, PATH);` method where the header names from the user entered spreadsheet are converted to lower case. 

Changed the code to call another method `public Spreadsheet(boolean convertHeaderNames, RequestParameter file, String... required)` in `com.adobe.acs.commons.data.Spreadsheet.java` to write the column names as listed in the spreadsheet by user. 

This will allow the columns names to be written as listed by user to AEM. 

Users should be aware that the column names in the spreadsheet must match the case (camel case) for property names (key's), the way they need to be written to AEM. The reason for this is that if the case-sensitivity is not maintained, AEM UI is not identifying these properties and is not displaying the values accordingly.

For example, if the metadata schema has a field named `abc:alternateTextCaption` but if key is written to AEM as `abc:alternatetextcaption` (case-sensitivity not maintained), then UI will not map the value to that property to be displayed on the UI.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/DataImporter.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/DataImporter.java
@@ -58,6 +58,7 @@ public class DataImporter extends ProcessDefinition {
 
     private static final Logger LOG = LoggerFactory.getLogger(DataImporter.class);
     private static final String PATH = "path";
+    private boolean enableHeaderNameConversion = false;
 
     public enum MergeMode {
         CREATE_AND_OVERWRITE_PROPERTIES(true, true, true),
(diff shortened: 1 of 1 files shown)


ksurendra:
Attaching sample test file 
[Publish_udpate-sample.xlsx](https://github.com/Adobe-Consulting-Services/acs-aem-commons/files/2308137/Publish_udpate-sample.xlsx)

ahmed-musallam:
I see this issue happening in 3.19.0

Issue #1459, opened 2018-08-16
Redirect Map Manager - Edit Individual Entries


klcodanr:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: All
* [ ] ACS AEM Commons Version: 3.17.0
* [ ] Reproducible on Latest? yes

### Expected Behavior

it would be great to allow authors to directly edit the redirect map entries

### Actual Behavior

Instead they have to remove and re-add.

### Steps to Reproduce

N/A

### Links

N/A

PR #1465, opened 2018-08-24
Resolving #1459 - Redirect Entry Editing

klcodanr:
adding the ability to edit an individual entry and a confirmation when deleting an entry.

Validated in 6.2 - 6.4.


justinedelson:
@klcodanr this is part of #1466 now right?

klcodanr:
@justinedelson -- yeah, kind of awkward with multiple changes against the same module. Next time I'll hopefully plan better and open a single PR with both changes.

PR #1466, opened 2018-08-24
Defect/1458 Redirect Map Entry Update not updating date

klcodanr:


klcodanr:
FYI -- the issue reported by code climate is the name of a method I'm overriding in a test from the ResourceResolver.

Issue #1467, opened 2018-08-24
Versioned ClientLibs cause WARN log messages on AEM 6.3


jsedding:
## Required Information
* AEM 6.3
* ACS AEM Commons latest

### Expected Behavior
Using Versioned ClientLibs should not cause any WARN log messages.

### Actual Behavior
Versioned ClientLibs cause frequent WARN log messages, [as documented](https://adobe-consulting-services.github.io/acs-aem-commons/features/versioned-clientlibs/index.html#log-warnings-under-aem-63).

### Steps to Reproduce
Not needed, the undesired behaviour is documented.

Issue #1216, opened 2017-12-15
Versioned clientlibs causes error log to be filled


royteeuwen:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.3 SP1
* [ ] ACS AEM Commons Version: 3.12.1
* [ ] Reproducible on Latest? yes

### Expected Behavior

There should be no error log lines from the versioned clientlibs when opening ANY page in the author environment

### Actual Behavior

For every page that is opened on author, the following log lines are produced:

> 15.12.2017 17:34:40.248 *WARN* [10.45.0.1 [1513355680163] GET /MB/aem/start.html HTTP/1.1] com.adobe.granite.ui.clientlibs.impl.HtmlLibraryManagerImpl No library configured at /apps/clientlibs/granite/typekit
15.12.2017 17:34:40.248 *WARN* [10.45.0.1 [1513355680163] GET /MB/aem/start.html HTTP/1.1] com.adobe.granite.ui.clientlibs.impl.HtmlLibraryManagerImpl No library configured at /apps/clientlibs/granite/jquery
15.12.2017 17:34:40.248 *WARN* [10.45.0.1 [1513355680163] GET /MB/aem/start.html HTTP/1.1] com.adobe.granite.ui.clientlibs.impl.HtmlLibraryManagerImpl No library configured at /apps/clientlibs/granite/utils
15.12.2017 17:34:40.248 *WARN* [10.45.0.1 [1513355680163] GET /MB/aem/start.html HTTP/1.1] com.adobe.granite.ui.clientlibs.impl.HtmlLibraryManagerImpl No library configured at /apps/clientlibs/granite/jquery/granite
15.12.2017 17:34:40.248 *WARN* [10.45.0.1 [1513355680163] GET /MB/aem/start.html HTTP/1.1] com.adobe.granite.ui.clientlibs.impl.HtmlLibraryManagerImpl No library configured at /apps/clientlibs/granite/moment
15.12.2017 17:34:40.249 *WARN* [10.45.0.1 [1513355680163] GET /MB/aem/start.html HTTP/1.1] com.adobe.granite.ui.clientlibs.impl.HtmlLibraryManagerImpl No library configured at /apps/clientlibs/granite/coralui3
[...]

PR #1468, opened 2018-08-24
#1467 - Versioned ClientLibs cause WARN log messages on AEM 6.3 (#1467)

jsedding:
Avoids the documented warnings for proxied clientlibs by checking whether the respective clientlib exists using the method `HtmlLibraryManage#getLibraries()` before loading the HtmlLibrary as before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 
 ## [3.17.4] - 2018-08-15
 
+### Fixed
+- #1467 - Versioned ClientLibs cause WARN log messages on AEM 6.3
 - #1413 - Added ACL to make the redirect maps globally readable
 
 ## [3.17.2] - 2018-08-13
--- a/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/VersionedClientlibsTransformerFactory.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/VersionedClientlibsTransformerFactory.java
@@ -23,6 +23,7 @@ import com.adobe.acs.commons.rewriter.AbstractTransformer;
 import com.adobe.acs.commons.util.impl.AbstractGuavaCacheMBean;
 import com.adobe.acs.commons.util.impl.CacheMBean;
 import com.adobe.acs.commons.util.impl.exception.CacheMBeanException;
+import com.adobe.granite.ui.clientlibs.ClientLibrary;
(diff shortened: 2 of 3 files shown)


jsedding:
2/5 travis builds failed due to problems connecting to a maven repository. So the failure does not seem to be a reflection of this PR but rather an infrastructure/load problem at travis.

Is it possible to re-run travis? If yes, how?

royteeuwen:
Hey @jsedding . How will this fix the problem?
See the issue I once created a while back: https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/1216

jsedding:
Hey @royteeuwen - it fixes the issue by checking if the library is available under the calculated path before actually requesting it. The check is done using `HtmlLibraryManage#getLibraries()`, which returns a map of `ClientLibrary` instances. These seem to be the more internal representation of a `HtmlLibrary` but do refer to the same node structures.

So in short:
1. check if the proxied library exists using `getLibraries().get(path)`
2. only if it exists, call `HtmlLibraryManager#getLibrary(type, path)`, which then doesn't log a warning

royteeuwen:
Hey @jsedding ,

To my understanding, it would still log the error when the proxied clientlib is first looked up in /apps but is actually in /libs?

justinedelson:
@jsedding this looks like a very elegant solution. I've retriggered the travis build.

One (lazy) question - is there any (significant) performance impact of this change?

jsedding:
FWIW, it turns out that my gut feeling was wrong. I did some simple Apache Bench tests calling `/sites.html/content`, which (on my development machine) take 80-100ms (`ab -A admin:admin -n1000 -c5 http://localhost:4502/sites.html/content`). After my changes, it's about 2x that.

I will update the PR with a cache in order to mitigate the slowdown.

justinedelson:
@jsedding does it help to only call `getLibraries()` once, i.e. outside of the search path loop. That would be simpler than a cache.

jsedding:
@justinedelson not so much, because the search paths loop only has two iterations (in a standard setup). So the cost would be cut in half. Now the rewriter may rewrite e.g. 10 clientlibs per rendered HTML page, thus incurring the cost of copying the map 10 instead of 20 times.

With the previous solution the cost of copying the map is never incurred, thus we would still have a significant overhead. With a cache, we would get (almost) on par with the previous solution. For a system with no changes to clientlibs (i.e. normal publish) the cost would be incurred once.

Of course cache invalidation increases complexity somewhat. So there's a trade off.

jsedding:
Thanks for merging and you're welcome for the fix!

Issue #1462, opened 2018-08-18
colorfield can not work well on multifield


xuanuy:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: Adobe Experience Manager (6.3.2.2)
* [ ] ACS AEM Commons Version: acs-aem-commons-content-3.17.0
* [ ] Reproducible on Latest? yes

### Expected Behavior

granite/ui/components/coral/foundation/form/colorfield should be load in granite/ui/components/foundation/form/multifield correctly.

### Actual Behavior

granite/ui/components/coral/foundation/form/colorfield doesn't load correct in granite/ui/components/foundation/form/multifield 

### Steps to Reproduce

Use granite/ui/components/coral/foundation/form/colorfield in granite/ui/components/foundation/form/multifield, config to save value as json.
Can save value when submit dialog but when re-open dialog the value of colorfield is not loaded correctly.

xuanuy:
Modified touchui-widgets-init.js to fix.
[touchui-widgets-init.js.zip](https://github.com/Adobe-Consulting-Services/acs-aem-commons/files/2299574/touchui-widgets-init.js.zip)

justinedelson:
@xuanuy can you submit this as an actual pull request?

jeffreyniu:
do you add this hotfix to package 3.18.2 or 3.19.0? i tried these 2 packages, but i found it was not be fixed.

PR #1474, opened 2018-08-30
colorfield can not work well on multifield #1462

xuanuy:


davidjgonzalez:
@schoudry / @justinedelson  - this look good (less the missing changelog) .. i can no longer eyeball extjs stuff.

schoudry:
changes look good, btw  composite multifield in ACS commons was deprecated (in favor of otb one introduced in 63)

PR #1479, opened 2018-09-07
replication status updater: start one job per package to let an

kwin:
exception not affect other packages


chu690214:
[scripting_sdk_076.zip](https://github.com/Adobe-Consulting-Services/acs-aem-commons/files/2361216/scripting_sdk_076.zip)

Issue #1480, opened 2018-09-09
Adobe I/O healthcheck must only check 1 onepoint


heervisscher:
## Required Information

* [6.3 ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: _______
* [3.17.5 ] ACS AEM Commons Version: _____
* [yes] Reproducible on Latest? yes/no

### Expected Behavior

When doing a Adobe I/O healthcheck you just want to check one endpoint.
Now you have the risk of POST/DELETE api-calls

### Actual Behavior

All endpoints are checked in the healthcheck

### Steps to Reproduce

Run the Adobe I/O healthcheck

### Links

_Links to related assets, e.g. content packages containing test components_

PR #1481, opened 2018-09-09
Only checking for 1 endpoint

heervisscher:
fix for https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/1480

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/adobeio/service/EndpointService.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/adobeio/service/EndpointService.java
@@ -54,7 +54,8 @@ public interface EndpointService {
 
 
     /**
-     * Performs the GET-action connected to the endpoint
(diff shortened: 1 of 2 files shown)


heervisscher:
thanks @justinedelson, do you agree that != GET should not be tested like this?

justinedelson:
for sure. I'm just disagreeing with filtering the services using a specific id.

but that said, I noticed that the JavaDoc for performIO_Action() says that it always uses GET (see https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/adobeio/service/EndpointService.java#L57). So maybe the problem isn't the HC but rather the implementation of this method. Or the JavaDoc is wrong :)

heervisscher:
@justinedelson two changes:
1. only GET-requests in the hc
2. javadoc change

Issue #1464, opened 2018-08-23
ResourceResolverMapTransformer should decode URI-encoded attribute values


jsedding:
## Required Information
* AEM 6.3 (any SP or CFP)
* ACS AEM Commons 3.17.0 / Latest
* ResourceResolverMapTransformerFactory configured to rewrite img:src

### Expected Behavior
On author the assets UI should display images. 

### Actual Behavior
On author the assets UI displays no images, neither on tiles/cards or in the asset-detail view of an  individual image.

### Steps to Reproduce
- Set up ResourceResolverMapTransformerFactory to rewrite img:src on an AEM 6.3 author.
- Navigate to /assets.html into a folder containing images.

### Analysis
The code rendering the assets UI calls org.apache.jackrabbit.util.Text#escapePath on the thumbnail URI (actually the web rendition in this case).

Hence the ResourceResolverMapTransformer gets an encoded path, more specifically with the "jcr:content" part encoded as "jcr%3acontent". Calling ResourceResolver#map on this doesn't achieve anything, as the path cannot be found in the repository. So no namespace mangling or other mappings are applied.

For some reason, I'm not quite sure if that's the rewriter pipeline, the URI-encoded "jcr:content" ends up rendered double-encoded, i.e. "jcr%253acontent", leading to a 404 when the browser loads the image.

I suggest to URI-decode the attribute value before passing it to ResourceResolver#map in [ResourceResolverMapTransformerFactory.java#L95](
https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/78f849c2db910503a27035f3bc6e8af6cdc4acd2/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/ResourceResolverMapTransformerFactory.java#L95).

PR #1492, opened 2018-09-21
Decode attribute value before mapping

rbotha78:
The resource resolver map transformer needs to first decode the attribute before passing it on to be mapped since the resolver map function is returning an encoded value. 
If the attribute is encoded before passing to map, the result is a double encoded value which results in 404 errors.


rbotha78:
This fixes #1464

PR #1494, opened 2018-09-25
allow configured 404 page to not include higher components/templates

kaushalmall:
… reset method to remove TopLevelComponent context and let the configured 404 page load correctly. The 404 jsp now calls the reset method.


kaushalmall:
@davidjgonzalez @justinedelson do we need a change log entry for this?

justinedelson:
@kaushalmall always :)

kaushalmall:
done. created issue #1495 to track as well.

PR #1503, opened 2018-09-27
Bugfix/1501 poi compatibility with 6330

josephrignanese:
Due to a change in version of the POI library, there are issues with the previous calls.
This seems to work (for now) with 6.2 -> 6.4.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ after the 3.9.0 release. All changes up until the 3.9.0 release can be found in 
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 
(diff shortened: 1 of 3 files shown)


badvision:
Thanks @josephrignanese!!!

josephrignanese:
@badvision, yeah it is a bit weird. The POI library seemed to have some undocumented breaking changes and this was the only approach I could see available across all the versions.

Right now the versions to be compatible with are:
AEM 6.2 - 3.11
AEM 6.3 - 3.15, 3.17
AEM 6.4 - 3.16

Note that the constructor is removed in future versions of POI (>=4.0.0).

PR #1518, opened 2018-10-12
Fix #1428 - MCP> URL Asset Import - Column names from spreadsheet not retaining case-sensitivity

### LINKED ISSUE #1428: MCP - URL Asset Import - Column names from spreadsheet are imported without retaining case-sensitivity

ksurendra:
Updated code to not convert column names to lower case. Retains case-sensitivity. 
Removed 'toLowerCase' from 'convertHeaderName' method in 'Spreadsheet'.

We have to ask user to make sure the column name for "**source, target, rendition, original**" are in **lower-case**, as described in the documentation at https://adobe-consulting-services.github.io/acs-aem-commons/features/mcp-tools/asset-ingestion/url-asset-ingestor/index.html
 
_Ignored columns
A column will be ignored in the following cases:

The name is not one of the following: source, target, rendition, original
The name does not have a colon : character in it denoting some kind of metadata property._

If the column names are not in lower-case, the "URL Asset Ingestor" will not be able to pick them up for processing.

Attached sample spreadsheet to run/test.
[TestAssetImportUsingMCP-UpdatedFormat.xlsx](https://github.com/Adobe-Consulting-Services/acs-aem-commons/files/2474140/TestAssetImportUsingMCP-UpdatedFormat.xlsx)


davidjgonzalez:
@ksurendra can you fix code climate please?

bundle/src/main/java/com/adobe/acs/commons/data/Spreadsheet.java
```diff
@@ -229,7 +230,7 @@ public String convertHeaderName(String str) {
             if (str.contains("@")) {
                 str = StringUtils.substringBefore(str, "@");
             }
-            return String.valueOf(str).toLowerCase().replaceAll("[^0-9a-zA-Z:\\-]+", "_");
+            return String.valueOf(str).replaceAll("[^0-9a-zA-Z:\\-]+", "_");
```
davidjgonzalez: There are a number of methods on JcrUtil for creating valid JCR names.. Could/should one of these be used instead of this more bespoke (regex) solution?

https://helpx.adobe.com/experience-manager/6-2/sites/developing/using/reference-materials/javadoc/com/day/cq/commons/jcr/JcrUtil.html

davidjgonzalez: Also, perhaps a quick unit test or two (esp if we have known failing cases which i assume drove this update?)

badvision:
Please re-open after resolving the concerns in this PR

PR #1522, opened 2018-10-17
Fix #1516 Undefined exception on configure-limit-parsys.min.js

### LINKED ISSUE #1516: Undefined exception on configure-limit-parsys.min.js
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.4
* [ ] ACS AEM Commons Version: 3.17.0

### Expected Behavior

With a custom component that has a resource of 'wcm/foundation/components/parsys', when we drag and drop any component no console error is expected.

### Actual Behavior

One error is logged every time that I drag and drop one component. This error is an undefined exception on /apps/acs-commons/widgets/configure-limit-parsys.min.js
This error happens on the: cellSearchPathInfo = cellSearchPathInfo[parNames[c]]; Since the pasys was created dynamically using sly resource and for the component the firt time on the fir it gets undefined, and in the second it breaks. (almost as a nullpointer issue)

image

### Steps to Reproduce

Some component with a resource parsys:
<sly data-sly-resource="${ column.parsysKey @ resourceType='wcm/foundation/components/parsys'}"/>

The final structure would be something like:

image

Drag and drop some component on the last parsys and see the error being logged on the console.

First iteration:
image

Second one:
image

Third and last one:
image

ParNames:
0: "common_content_iparsys"
1: "responsivecolumncont"
2: "column0"

Mupc:
This prevent the exception since it would not have the  ACS_COMPONENTS_LIMIT configured for it. So in this case this should not trow any error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 - #1497 - Javadoc improvement in EndpointService
 - #1506 - Fixed path browser input fields in MCP to work on AEM 6.4
 - #1501 - Error downloading reports from MCP processes with 6.3.3.0
+- #1516 - Undefined exception on configure-limit-parsys.min.js
 
 ## [3.18.2] - 2018-09-26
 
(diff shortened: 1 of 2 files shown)


davidjgonzalez:
Thanks!

PR #1526, opened 2018-10-26
Feature/action manager priority queue

shsteimer:
I think this is mostly self explanatory.  By adding an optional parameter to the action manager for the priority, we can give tasks a priority and order their execution based on this priority.

While this does not exactly guarantee the order of execution (could vary depending on how long various tasks take), it does allow you to order tasks that are more important first.  For example, you may want to execute more resource intensive tasks first to free up those resources for other things.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/fam/impl/ActionManagerFactoryImpl.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/fam/impl/ActionManagerFactoryImpl.java
@@ -20,6 +20,7 @@
 package com.adobe.acs.commons.fam.impl;
 
 import com.adobe.acs.commons.fam.ActionManager;
+import com.adobe.acs.commons.fam.ActionManagerConstants;
 import com.adobe.acs.commons.fam.ActionManagerFactory;
 import com.adobe.acs.commons.fam.ThrottledTaskRunner;
 import com.adobe.acs.commons.fam.mbean.ActionManagerMBean;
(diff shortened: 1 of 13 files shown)


bundle/src/main/java/com/adobe/acs/commons/fam/impl/ActionManagerFactoryImpl.java
```diff
@@ -54,12 +56,17 @@ public ActionManagerFactoryImpl() throws NotCompliantMBeanException {
...
     public ActionManager createTaskManager(String name, ResourceResolver resourceResolver, int saveInterval) throws LoginException {
+        return this.createTaskManager(name, resourceResolver, saveInterval, ActionManagerConstants.DEFAULT_ACTION_PRIORITY);
+    }
+
+    @Override
+    public ActionManager createTaskManager(String name, ResourceResolver resourceResolver, int saveInterval, int priroty) throws LoginException {
         String fullName = String.format("%s (%s)", name, UUID.randomUUID().toString());
         
-        ActionManagerImpl manager = new ActionManagerImpl(fullName, taskRunner, resourceResolver, saveInterval);
+        ActionManagerImpl manager = new ActionManagerImpl(fullName, taskRunner, resourceResolver, saveInterval, priroty);
```
badvision: Misspelled priority

shsteimer: This is exactly the sort of high quality code review I expect.

Snark aside, will fix this, thanks for pointing it out.

badvision: Hey, it's open-source.  You're getting so much more than you paid for already!  ;P  But seriously, thanks for the awesome PR!

PR #1529, opened 2018-10-30
Instant Package Utility

edubey:
demo

image

*Description:* Allows user to instantly create and download a package from selected resource and including child resources. Purpose is create package when its quickly required without worrying about name, title, version, description.

*Current Steps*

• Create package
• Add filter by copy / paste I guess
• Build and download

*Instant Package*

• Select the path in UI 
• Click to download an AEM Package

*Purpose:* Save time and simplified process

- Target audience is mainly developers to save time, though can be used by authors also.
- By default will be **disable**, can be enable /disable from _Instant package config page_
- Support Sites and Assets as of now.

 **Share any improvements and feedback**

**Possible Improvement to this feature :-** 
Deletion of packages post download or a scheduler job 
Restrict to admin only


badvision:
This looks pretty exciting, though please review the code climate issues and consider if there's any way to add some test coverage if possible (-0.1% isn't a deal-breaker though)

PR #1532, opened 2018-10-31
Request throttler

joerghoh:
I'd like to contribute the request throttler, a mechanism which allows you to throttle certain requests based on the current CPU load.


badvision:
This is very interesting @joerghoh!  Thank you for this.  It looks like code climate is still being picky about a few things.  If I read this correctly, this doesn't do anything by default?  Is it possible to have the service not actively filtering if no paths are configured?

joerghoh:
@badvision It's a service factory, so if there's no config it shouldn't be active at all. But I haven't considered the case if the configuration has no path configured.

badvision:
Thanks @joerghoh for another quality submission.  I'll merge in and resolve the remaining code climate issues.

joerghoh:
@badvision thanks for your work and accepting it!

PR #1536, opened 2018-11-01
[Review] Feature/pwa

davidjgonzalez:


bundle/src/main/java/com/adobe/acs/commons/cloudservices/pwa/impl/ConfigurationImpl.java
```diff
@@ -0,0 +1,125 @@
...
+
+            if (StringUtils.isNotBlank(confPath)) {
+                rootPage = page;
+                Resource cloudConfigsResource = resourceResolver.getResource(confPath + "/settings/cloudconfigs");
+                if (cloudConfigsResource != null) {
+                    for (final Resource cloudConfigResource : cloudConfigsResource.getChildren()) {
+                        Resource jcrContentResource = cloudConfigResource.getChild(JcrConstants.JCR_CONTENT);
+                        if (jcrContentResource != null && jcrContentResource.isResourceType(RESOURCE_TYPE)) {
+                            confPage = pageManager.getContainingPage(jcrContentResource);
+                            //rootPage = page;
```
davidjgonzalez: @avitripathy01 So this will not support nested cq:conf, right? Is there a use-case for nesting cq:conf applications in a site tree?

bundle/src/main/java/com/adobe/acs/commons/cloudservices/pwa/impl/ConfigurationImpl.java
```diff
@@ -0,0 +1,125 @@
...
+
+            page = page.getParent();
+        }
+    }
+
+    @Override
+    public boolean isReady() {
+        return rootPage != null;
+        // TODO Uncomment
+        //return slingSettingsService.getRunModes().contains("publish");
```
davidjgonzalez: SW caching on Author will break things, so unless this check has been moved, this should be incorporated into the isReady check. TBH, i'd prefer to loading things on author we dont need to.

bundle/src/main/java/com/adobe/acs/commons/cloudservices/pwa/impl/Constants.java
```diff
@@ -0,0 +1,81 @@
...
+
+    /** JSON Keys **/
+    public static final String KEY_CACHE_NAME = "cache_name";
+    public static final String KEY_VERSION = "version";
+    public static final String KEY_FALLBACK = "fallback";
+    public static final String KEY_PRE_CACHE= "pre_cache";
+    public static final String KEY_NO_CACHE = "no_cache";
+    public static final String KEY_PATTERN = "pattern";
+    public static final String KEY_PATH = "path";
+    public static final String KEY_SW_Scope = "scope";
```
davidjgonzalez: Wrong casing for Java Constant

bundle/src/main/java/com/adobe/acs/commons/cloudservices/pwa/impl/PWAProxyServlet.java
```diff
@@ -0,0 +1,295 @@
...
+                "sling.servlet.extensions=load",
+                "sling.servlet.extensions=js",
+                "sling.servlet.selectors=pwa",
+                "sling.servlet.selectors=manifest",
+                "sling.servlet.selectors=service-worker",
+                "sling.servlet.methods=" + HttpConstants.METHOD_GET,
+                "sling.servlet.resourceTypes=cq:Page"
+        }
+)*/
+public class PWAProxyServlet extends SlingSafeMethodsServlet implements OptingServlet {
```
davidjgonzalez: I think we can delete this file now? It should've been moved out to other servlets (separation of concerns)

bundle/src/main/java/com/adobe/acs/commons/cloudservices/pwa/impl/PwaServiceWorkerConfigServlet.java
```diff
@@ -0,0 +1,157 @@
...
+ 
+    }
+
+    private JsonObject getConfig(SlingHttpServletRequest request) throws ServletException {
+        final Configuration configuration = modelFactory.createModel(request, Configuration.class);
+        final ValueMap properties = configuration.getProperties();
+        final JsonObject json = new JsonObject();
+        if(properties !=null ){
+            final int version = properties.get(PN_VERSION, 1);
+            final String swScope = properties.get(KEY_SW_Scope, "/");
```
davidjgonzalez: Are we requiring the scope to be selected by the Config author? Should this be implicit based on where the cq:conf is? Would this be a pre-RR Mapped value (ie. JCR content path vs mapped path)?

bundle/src/main/java/com/adobe/acs/commons/cloudservices/pwa/impl/PwaServiceWorkerJavaScriptServlet.java
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }
+    }
+
+    private void writeJavaScript(SlingHttpServletRequest request, SlingHttpServletResponse response) {
+
+        final Configuration configuration = modelFactory.createModel(request, Configuration.class);
+        boolean serviceWorkerJsFlag = false;
+        Collection<ClientLibrary> htmlLibraries ;
+        for (String selector : request.getRequestPathInfo().getSelectors()) {
```
davidjgonzalez: Why are we checking for selectors in servlets? We should be letting Sling resolve these for us.

content/src/main/content/jcr_root/apps/acs-commons/cloudservices/pwa/clientlibs/clientlib-idb/.content.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<jcr:root xmlns:cq="http://www.day.com/jcr/cq/1.0" xmlns:jcr="http://www.jcp.org/jcr/1.0"
+    jcr:primaryType="cq:ClientLibraryFolder"
+    jcr:title="ACS Commons IndexedDb"
+    allowProxy="{Boolean}true"
+    categories="[acs-commons.idb]"/>
```
davidjgonzalez: IIRC we mark vendor resources in vendor folders (skips code quality checks) and also add vendor to the category so its clear we (acs commons) arent providing this JS.

content/src/main/content/jcr_root/apps/acs-commons/cloudservices/pwa/clientlibs/clientlib-serviceworker/.content.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<jcr:root xmlns:cq="http://www.day.com/jcr/cq/1.0" xmlns:jcr="http://www.jcp.org/jcr/1.0"
+    jcr:primaryType="cq:ClientLibraryFolder"
+    jcr:title="ACS Commons Generic Service Worker"
```
davidjgonzalez: This should have a better name - this is a very elaborate SW now.

bundle/src/main/java/com/adobe/acs/commons/cloudservices/pwa/impl/PwaManifestServlet.java
```diff
@@ -0,0 +1,167 @@
...
+
+    @Override
+    protected final void doGet(SlingHttpServletRequest request, SlingHttpServletResponse response) throws ServletException, IOException {
+        response.setContentType("application/x-web-app-manifest+json; charset=UTF-8");
+        response.setCharacterEncoding("UTF-8");
+
+
+        ResourceResolver serviceResourceResolver = null;
+        try {
+            serviceResourceResolver = resourceResolverFactory.getServiceResourceResolver(AUTH_INFO);
```
davidjgonzalez: Why are we using service users in the context of a Request? The user's permissions should be used, and ACLs should be applied accordingly (like we do for Templates, etc.) Or, CUGs could be set on the resource so when its replicated, they are accessible as needed.

This should only be needed on Publish as well.

Now, anyone that hits this URL can access the payload (ie. if this is a secured site, this URL is not secure anymore).

content/src/main/content/jcr_root/apps/acs-commons/cloudservices/pwa/components/scripttag/header/header.html
```diff
@@ -0,0 +1,10 @@
+<sly data-sly-use.configuration="${'com.adobe.acs.commons.cloudservices.pwa.Configuration'}"
+     data-sly-test.ready="${configuration.ready}"
+     data-sly-test.path="${configuration.scopePath}">
+    <link rel="manifest" href="${path}.pwa.manifest.json" type="application/manifest+json" crossOrigin="use-credentials">
```
davidjgonzalez: So scope paths MUST reside at or under a page tree, that has a cq:conf that has a valid PWA Configuration, right? What happens if it does not? Does it gracefully fail?

badvision:
Looks cool but much left to be done (code climate, test coverage, etc)

joerghoh:
Ok, maybe I've missed something, but what's this all about? Is there an earlier PR which contains much more documentation?

avitripathy01:
This PR creates a cloud service configuration that could be applied to a site or context under /conf.  Creating the configuration includes filling out details needed for creating the Service Worker(SW) and the manifest file along with the few other details like what paths/patterns of the url to cache and uncache.
The manifest configuration details helps making the site function like a native mobile app allowing to add the site as a shortcut to the device homescreen.
The SW acts a a proxy to every request and starts caching the response on the browser. So the next time a request is made for the same url, the response is served from the cache in the browser.

The manifest details are loaded as part of the header of the site through a manifest servlet and the pwa-clientlib loads at the footer.
The pwa-clientlibs makes a servlet call to know whether it should and under what location of the site it should create the SW. Whenerver there is a new content available at the server side, the cache version is checked from the cloud service configuration and the browser cache is updated. The cache busting is implemented using IndexedDb.

davidjgonzalez:
Closing for @avitripathy01 can re-open as his

schoudry:
some screenshots or video recording please....

PR #1537, opened 2018-11-01
Creation of ResourceResolvers in a try-with-resource clause

joerghoh:
Use try-with-resource for all ResourceResolvers, so we don't need the "finally" Block anymore and never forget to close the ResourceResolver.

Regarding the change to FastActionManagerRunnerImpl:
I am not sure if my change is correct, because it seems to me that a lot of async activities
are started, and therefor the automatic close of the RR when leaving the block might not be correct.

But even the original implementation had a flaw, because I haven't found a matching "close" of
this RR at first sight. @davidjgonzalez Can you help here?

converted JcrPackageReplicationStatusEventHandler
converted crPackageReplicationStatusEventHandler
converted AutomaticPackageReplicatorJob
converted WCMInboxWebConsolePlugin
converted EmailServiceImpl
converted SystemNotificationsImpl
converted SharedComponentPropertiesPageInfoProvider
converted ComponentErrorHandlerImpl
converted ErrorPageHandlerImpl
converted DispatcherFlushRulesImpl
converted ReviewTaskAssetMoverHandler
converted JCRHttpCacheStoreImpl
converted HttpCacheInvalidationJobConsumer
converted EnsureServiceUser
converted EnsureGroup
converted WorkflowInstanceRemoverScheduler
converted AEMTransientWorkflowRunnerImpl
converted FastActionManagerRunnerImpl -- although I am not sure if that works at all here ...
converted AEMWorkflowRunnerImpl
converted SyntheticWorkflowRunnerImpl
converted ProcessInstanceImpl
converted AbstractFormHelperImpl
converted EnsureOakIndexJobHandler

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/dam/impl/ReviewTaskAssetMoverHandler.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/dam/impl/ReviewTaskAssetMoverHandler.java
@@ -159,10 +159,8 @@ public class ReviewTaskAssetMoverHandler implements EventHandler {
 
     @Override
     public void handleEvent(Event event) {
-        ResourceResolver resourceResolver = null;
 
-        try {
-            resourceResolver = resourceResolverFactory.getServiceResourceResolver(AUTH_INFO);
+        try (ResourceResolver resourceResolver = resourceResolverFactory.getServiceResourceResolver(AUTH_INFO)){
             final String path = (String) event.getProperty("TaskId");
             final Resource taskResource = resourceResolver.getResource(path);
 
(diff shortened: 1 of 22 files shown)


joerghoh:
@badvision @davidjgonzalez So there was indeed a leaking ResourceResolver in FastActionManagerRunnerImpl?

badvision:
In the Bulk Workflow implementation that used it, looks like.

davidjgonzalez:
Anything that uses QueryBuilder has high-risk for leaking the RR

davidjgonzalez:
@joerghoh  sorry - what what the help request? :) To update the impl's or review the updates?

joerghoh:
I guess it was about reviewing if the implicit closing of the RR is correct here.

Issue #1560, opened 2018-11-14
Switch to OSGI annotations


joerghoh:
Although the SCR annotations are still supported, we should switch over to OSGI annotations.

Carsten Ziegeler's blog:
* https://blog.osoco.de/2016/05/migrating-from-the-apache-felix-scr-annotations-to-the-osgi-declarative-services-annotations/ (Overview)
* https://blog.osoco.de/2015/08/osgi-components-simply-simple-part-i/  (Components, References)
* https://blog.osoco.de/2015/08/osgi-components-simply-simple-part-ii/ (Configuration, Services)
* https://blog.osoco.de/2015/11/osgi-components-simply-simple-part-iii/ (Metatype)

Specific for AEM:
* @heervisscher: http://blogs.adobe.com/experiencedelivers/experience-management/using-osgi-annotations-aem6-2/ (focus on Configuration)
* @nateyolles: http://www.nateyolles.com/blog/2017/05/osgi-declarative-services-annotations-in-aem

Registering Sling Servlets (new annotations): https://sling.apache.org/documentation/the-sling-engine/servlets.html#registering-a-servlet-using-java-annotations
* For the different ways to register servlets (resourcetype-based or path-based) there are different annotations!

References:
* #1567 how to deal with configuration ids containing a hyphen ("-")

Learnings regarding Labels and Descriptions
* SCR annotations mixed all up a bit; especially the label and description of the component was taken for the component itself and its configuration. This is clearly separated with OSGI annotations:
* ```@Component``` allows you to have a description, but I haven't found where it's reflected. Anyhow, just the full qualified classname in the /system/console/components isn't a bad thing either, isn't it?
* Use the name and the description on the ```ObjectClassDefinition``` annotation to provide metadata for the Configuration.

Things I learned when dealing with Configurations and properties:
* While with SCR you play with ```@Property``` and ```@Properties``` as you like (write on top of the class or annotate constant strings with it), OSGI annotations give some clear recommendations:
* If you are not supposed to change a value (with SCR you used ```propertyPrivate=true```), add it to the ```@Component``` annotation. And propertyPrivate is not needed anymore.
* If you should be able to change them, add them to the ```@ObjectClassDefinition``` interface (tpyically called ```Config```).
* You don't need the ```metatype=true|false``` properties in the ```@Component``` annotation anymore. If you have a ```@Designate(ocd=...class)``` annotation on your service, it will generate metatype information automatically for it.

[...]

PR #1565, opened 2018-11-15
Switch to OSGI annotations

joerghoh:
I started the conversion of the first services. Mostly straight forward.

**Learnings:**
* Haven't found a good way to convert the \@Component(label="foo") to OSGI. Dropped it for the moment.
* Properties marked as private remain as properties with the \@Components annotation; all other properties move to the Configuration class; even if they might not be used afterwards, but they just remain as OSGI properties.

Please provide feedback; If you are ok with it, let's add this to #1560.


badvision:
This looks fine with me, can you resolve the minor conflicts and minor code climate complaint?

joerghoh:
Tests are running well, I just have 6 baseline errors ... @badvision: are you interested in fixing them

PR #1568, opened 2018-11-17
Feature/remove deprecated methods

badvision:


joerghoh:
That's a lot of stuff in there. I would prefer to have it splitted into multiple smaller PRs with different scopes (maybe even a few PRs to do the JSON stuff, another one for clustering, etc). But here we are :-)

Most stuff looks reasonable, changes are mostly trivial. I would do a +0 (the only downside being the size of this PR).

badvision:
@joerghoh I agree on principle but most of the changes were literally the same code copy/pasted in many places using the Sling JSON classes.  I tried to minimize the redundancy by abstracting the deprecating/missing/changed functionality into some new utility classes.

Issue #1570, opened 2018-11-19
3.19.1 master doesn't deploy on 6.3/6.4


heervisscher:
## Required Information

* [AEM6.3 ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: ___
* [3.19.1 ] ACS AEM Commons Version: _____
* [yes ] Reproducible on Latest? yes/no

### Expected Behavior

Bundle is active

### Actual Behavior

Bundle is 'installed'

heervisscher:
![image](https://user-images.githubusercontent.com/1647181/48711360-61898d00-ec0b-11e8-8143-fc0c062cf4ed.png)

heervisscher:
Also on 6.4 there is the same issue

davidjgonzalez:
It appears the MCP SFTP code introduced this dependency. I don't see AEM providing this in AEM 6.4.2 either. 

I'd much prefer if we can re-write without dependency (or use a dep already provided by AEM).
* Our optionals list is already too big (requiring yet-another-install to get features to work)
* and we don't want to go back to embedding 3rd party libraries (a bad habit we've already broken).

Might be smart to just revert this feature until we can figure out a good solution.

joerghoh:
splitting the huge package of acs-aem-commons into smaller ones? especially if the code adds 3rd party dependencies.

davidjgonzalez:
that's what we did for Twitter, but TBH history has proven that is a PITA for just about everyone.

Spitting ACS Commons, in general, has been discussed in previous issues.

davidjgonzalez:
Making jcraft optional, and gracefully handling this lack of dependency "in product", is IMO is the best work-around. The right solution is, of course, to figure out how to not rely on 3rd party deps.

justinedelson:
FWIW, the problem with the Twitter dependency is pretty specific and not something I'd expect to be necessary in general. The issue there is that Twitter4J is used in an AdapterFactory and AdapterFactories are eagerly instantiated.

In this case, my understanding is that this dependency is used in a MCP process; a similar situation already exists with the S3 Ingestor. Seems like it would apply here.

Actually, this gives me an idea for how to fix the twitter thing...

heervisscher:
On 6.4.2 I have this
![image](https://user-images.githubusercontent.com/1647181/48714777-b9c48d00-ec13-11e8-9e75-04eb7f70b246.png)

justinedelson:
[...]

PR #1571, opened 2018-11-19
remove separate twitter bundle; use exception trapping to only regist…

justinedelson:
…er AdapterFactory when Twitter4J is available


justinedelson:
Upon installation, if Twitter4J isn't available, you get a single log message:

```
com.adobe.acs.commons.twitter.impl.TwitterAdapterFactory Twitter4J Library not found. Not registering TwitterAdapterFactory.
```

Note that due to #1570, this isn't actually testable as-is.

Also, this pull request doesn't remove the `min` packaging. I think that warrants keeping around for legacy reasons.

badvision:
Code climate strikes again... 🤕

badvision:
We're doing a major version bump so we could remove it now, but after 4.0.0 it will be a while before we get another option to consider it.

joerghoh:
+1 for removal of the twitter4j bundle, thus eliminating the need for the min bundle

badvision:
I'll merge this in once the confirmation build passes (which I expect it to)

PR #1572, opened 2018-11-19
Embed JSCH and use a newer version than previously found

badvision:
This should help with the deployment issues caused by the SFTP pull request

### DIFF
--- a/bundle/pom.xml
+++ b/bundle/pom.xml
@@ -67,7 +67,7 @@
                             !android.util.*,
                             *
                         </Import-Package>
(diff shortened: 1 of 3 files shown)


badvision:
@joerghoh , @justinedelson adding the JSCH as an embedded dependency didn't increase the size of the build too much at all.  In case there is some module exporting it, e.g. sample content, this is a better option because that exported version was rather old (as if the library itself isn't already rather old anyway!)  This newer version that I'm embedding at least doesn't require a deprecated hashtable type to pass properties during initialization.

PR #1577, opened 2018-11-20
Taking into account how many paragraphs are pasted

livvius:


content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/configure-limit-parsys/touchui-limit-parsys.js
```diff
@@ -124,7 +124,10 @@
         }
         var limit = parseInt(currentLimit);
         var children = getChildEditables(parsysEditable);
-        isWithin = children.length <= limit;
+        //in case is a paste action calculates the final number or paragraphs by adding how many are on the clipboard
```
kwin: rather "Take into account also the number of components in the clipboard in case this is a "paste"

content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/configure-limit-parsys/touchui-limit-parsys.js
```diff
@@ -124,7 +124,10 @@
...
+        //and subtracting always 1 because Granite.author.edit.findEditables() returns also the * component
+        var finalNumberOfParagraphs = pasteAction ? children.length - 1 + Granite.author.clipboard.length : children.length;
+        isWithin = finalNumberOfParagraphs <= limit;
```
kwin: the logic is a bit convoluted, why not making it simpler by calculating 
a) the number of components to be inserted (1 by default)
b) the number of slots already taken (independent of pasteaction)

Then the logic for checking whether you are inside the limits is the same for both cases. That should simplify this code a lot.

content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/configure-limit-parsys/touchui-limit-parsys.js
```diff
@@ -133,7 +136,7 @@
         return isWithin;
     }
 
-    function extendComponentDrop(){
+    function extendComponentDrop() {
```
kwin: non related formatting change

Issue #1593, opened 2018-11-29
Sftp Asset Ingestor throws URISyntaxException if item contains special characters


cylinder-y:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.4.1, 6.4.2
* [ ] ACS AEM Commons Version: 3.19.1
* [ ] Reproducible on Latest? yes

### Expected Behavior

If migrated folder contains files or folders with special characters, those items should be imported

### Actual Behavior

When we have space characters inside of imported folder, we have the following:

1. Asset is not imported, only exception is written into log
2. Folder not imported, and process is running in endless loop

### Steps to Reproduce

Run Sftp Asset Injector(file or url), with special chars in names

cylinder-y:
working on it

badvision:
Dude you rock! 🗡 Let me know if you need any assistance with this patch.  Also, if possible, could you update the unit test to assert it works?  I think I have a few mocks in there already that help bypass the SFTP library to at least test the rest of the code.

PR #1594, opened 2018-11-29
#1593 - encoding of url parts is added for FileAssetIngestor.SftpHier…

cylinder-y:
encoding of url parts is added for FileAssetIngestor.SftpHierarchicalElement and for FileOrRendition.SftpConnectionSource

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/FileAssetIngestor.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/FileAssetIngestor.java
@@ -46,20 +46,20 @@ import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.net.URLDecoder;
 import java.util.Objects;
 import java.util.Vector;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Stream;
 
+import static com.adobe.acs.commons.mcp.impl.processes.asset.HierarchicalElement.UriHelper.decodeUriParts;
+import static com.adobe.acs.commons.mcp.impl.processes.asset.HierarchicalElement.UriHelper.encodeUriParts;
+
 /**
  * Asset Ingestor reads a directory structure recursively and imports it as-is
  * into AEM.
  */
 public class FileAssetIngestor extends AssetIngestor {
-
-    static String SFTP_URL_ENCODING = "utf-8";
 
     public FileAssetIngestor(MimeTypeService mimeTypeService) {
         super(mimeTypeService);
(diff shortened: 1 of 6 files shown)


cylinder-y:
Failed because of #1595

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/FileAssetIngestor.java
```diff
@@ -57,6 +59,8 @@
  */
 public class FileAssetIngestor extends AssetIngestor {
 
+    static String SFTP_URL_ENCODING = "utf-8";
```
badvision: Can we move this into a more general location, e.g. since the HierarchialElement class is referencing this, maybe that's a more suitable place for this constant?

badvision:
Looks fantastic, and great work on the added test cases.  Can we move that one constant?  (see my review comments.)

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/FileAssetIngestor.java
cylinder-y: Sure, bu code climate has forced me to do so : cba38410d96b66ffba5d304dc8b8a62dd1c5e86e because of constant in Interface. I will move it into separate class as well as encodeUriParts method in HierarchialElement. It require additional work for urls with port

badvision: Yes please.  Thanks so much!  That picky code climate!!

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/HierarchicalElement.java
```diff
@@ -120,4 +124,26 @@ default void visitAllFiles(CheckedConsumer<HierarchicalElement> visitor,
...
+    class UriHelper {
+        static String SFTP_URL_ENCODING = "utf-8";
+
+        private UriHelper(){}
+
+        static String encodeUriParts(final String uri) throws UnsupportedEncodingException {
+            String[] uriParts = uri.split("/");
+
+            //path parts started from index 3
+            for (int i = 3; i < uriParts.length; i++) {
```
grakovama: Would be nice to replace a magic number 3 with a constant

badvision: But then it would lose its magic!

cylinder-y:
hello @badvision, could you please review?

badvision:
I am sad that the magic is gone 😭 but I will endeavor to move on.   On a serious note, great work especially with the test cases! 😀

PR #1600, opened 2018-12-03
Feature/httpcache configurable writing

Sc0rpic0m:
This will make it possible to choose the printwriter instead of output stream based on the cache config.

Some projects need the printwriter instead.


badvision:
Please add a changelog entry when you get a moment @Sc0rpic0m.

Sc0rpic0m:
Done.

Sc0rpic0m:
Good thought @davidjgonzalez . I wonder what @justinedelson has to say on this. Should we write such an auto detect option? Problem was that with getOutputStream(), the rewriter pipeline wasn't called. What I could do is add a 2nd ResponseWriteMethod  option (that is default) AUTO, that does what you describes. If we all agree I will write that. Let me know

davidjgonzalez:
IIRC the original discussion was around detecting if one or the other had already been engaged, which and we concluded there is not a way.


Another thought, which might be more accurate, it to just track in HttpCacheContent [1] by way of the ResponseWrapper [2] which (writer vs stream) was used to originally collect the content being cached. I can't think of a case where the cache value would be written to one, but we'd want to serve it from cache using the other..? This avoids having to deal w detections via content types/encodings etc... however, the content was served into the cache, is how it leaves the cached.

WDYT? 

[1] https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/httpcache/engine/CacheContent.java

[2] https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/httpcache/engine/HttpCacheServletResponseWrapper.java#L43

PR #1601, opened 2018-12-03
Asset Ingestor and Valid Jcr Name #1597If Preserve File name unchecked, asset name will support only the fol…

cylinder-y:
If Preserve File name unchecked, asset name will support only the following characters: letters, digits, hyphens, underscores, another chars will be replaced with hyphens


cylinder-y:
Bag my pardon for failed test, working on it

badvision:
Happens to the best of us @cylinder-y 😄

PR #1604, opened 2018-12-04
file asset import and url asset imports saves source path as migrated…

cylinder-y:
file asset import and url asset imports saves source path as migratedFrom property into assets jcr:content node. If asset is skipped the message in the format "source -> destination" is written into report


cylinder-y:
Hello @badvision, Could you please review it?

PR #1606, opened 2018-12-10
Url Asset Ingestor saves the same path in migratedFom property

cylinder-y:
getItemName returns path of the migrated item, if exception occurs - url will be used as Item Name


cylinder-y:
I would create another pull request

PR #1607, opened 2018-12-10
Feature/httpcache response write method

Sc0rpic0m:
see : https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/1600

Persists the write method being used by the servlet to serve up cache content.
Tested locally.


badvision:
+1 from me.  @davidjgonzalez does this one work for you?

PR #1612, opened 2018-12-12
Retries count and retry pause is configurable for all Asset Ingestors

cylinder-y:
new configuration field was added for FAM tasks:
-retry count
-retry pause

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/FileAssetIngestor.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/FileAssetIngestor.java
@@ -68,18 +68,9 @@ public class FileAssetIngestor extends AssetIngestor {
     @FormField(
             name = "Source",
             description = "Source folder for content ingestion which can be a local folder or SFTP url",
-            hint = "/var/mycontent, /mnt/all_the_things, sftp://host[:port]/base/path...",
-            required = true
+            hint = "/var/mycontent, /mnt/all_the_things, sftp://host[:port]/base/path..."
     )
     String fileBasePath;
-
-    @FormField(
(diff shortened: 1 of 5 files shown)


badvision:
I'm not sure that having the timeout replaced with retry_pause is a good idea.  timeout is more of a connection timeout interval and that could be several seconds.  Retry pause in the FAM actions class is something that is usually in milliseconds.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/FileAssetIngestor.java
```diff
@@ -333,7 +324,7 @@ private ChannelSftp openChannel() throws JSchException {
 
                 com.jcraft.jsch.Session session = jsch.getSession(username, uri.getHost(), port);
                 session.setConfig("StrictHostKeyChecking", "no");
-                session.setTimeout(timeout);
```
badvision: I think that timeout is not the same as retry pause

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/UrlAssetImport.java
```diff
@@ -140,11 +131,11 @@ public void init() throws RepositoryException {
             HttpClientBuilder clientBuilder = httpFactory.newBuilder();
             clientBuilder.setDefaultSocketConfig(
                     SocketConfig.custom()
-                            .setSoTimeout(timeout)
```
badvision: I think that timeout should be its own setting

cylinder-y:
Hi @badvision,
Thanks a lot for reviewing, I have rolled back timeout field
--Yaraslau

PR #1621, opened 2018-12-13
add dependency-check-plugin to identify potential vulnerabilities in …

justinedelson:
…dependencies

* Ignore guava since Oak is not compatible with newer Guava versions due to removal of Iterators.emptyIterator()
* Update jjwt and jsch to latest versions


BillDinger:
Love the addition of the dependency check, much appreciated!

justinedelson:
Note that part of this is also changing how the enforcer plugin and Embed-Dependency are configured.

Previously -- all dependencies were required to be compile scope and an explicit list was embedded.
Now - compile scope is still banned by default, but specific dependencies are allowed to be compile scope and the embed uses that list.

The net result is basically the same, but enabled the same set of dependencies to be used for embedding and vulernability checking.

justinedelson:
This last commit is something I'm not happy about -- the test all pass with Guava 27.0.1 (and 24.1.1), but there's garbage being logged and Travis errors out because the stderr is too big (fair call). So I'm now including a fixed copy of the problematic class (`org.apache.jackrabbit.oak.spi.commit.CommitInfo`). Don't see another way around this, but open to suggestions.

justinedelson:
@BillDinger  the one last thing I'm not happy about is that this slows down the Travis build. It should be possible to cache the database for at least 24 hours, but in order to do that, I'd need to know where the database is stored on disk. Do you happen to know?

BillDinger:
Agreed, we do the same for our builds figuring 24 hours is 'good enough' and refreshing it often is pretty slow. @justinedelson it's in  the `.m2` directory. so on my windows box 
 `C:\Users\bdinger\.m2\repository\org\owasp\dependency-check-data\3.0` and in there it's a file called `dc.h2.db` on *nix box it should be in the user's m2 direcotry in same location

patrickhousley:
@justinedelson You can set the path inside the configuration block of the plugin and then cache it from there.

```
<configuration>
  <cveValidForHours>24</cveValidForHours>
  <failBuildOnAnyVulnerability>true</failBuildOnAnyVulnerability>
  <skipProvidedScope>true</skipProvidedScope>
  <dataDirectory>target/owasp-data</dataDirectory>
</configuration>
```

patrickhousley:
What was the issue with the `org.apache.jackrabbit.oak.spi.commit.CommitInfo` class? I don't see it used directly in the repository. Could you point me to where it is used indirectly?

.travis.yml
```diff
@@ -2,10 +2,10 @@ language: java
 jdk: oraclejdk8
 env:
 - MAVEN_SKIP_RC=true MAVEN_OPTS="-Xmx512m -XX:MaxPermSize=512m -Duser.language=en -Duser.country=US" POST_COVERAGE=true
-- MAVEN_SKIP_RC=true MAVEN_OPTS="-Xmx512m -XX:MaxPermSize=512m -Duser.language=de -Duser.country=DE" POST_COVERAGE=false
-- MAVEN_SKIP_RC=true MAVEN_OPTS="-Xmx512m -XX:MaxPermSize=512m -Duser.language=es -Duser.country=ES" POST_COVERAGE=false
-- MAVEN_SKIP_RC=true MAVEN_OPTS="-Xmx512m -XX:MaxPermSize=512m -Duser.language=fr -Duser.country=FR" POST_COVERAGE=false
-- MAVEN_SKIP_RC=true MAVEN_OPTS="-Xmx512m -XX:MaxPermSize=512m -Duser.language=hi -Duser.country=IN" POST_COVERAGE=false
+- MAVEN_SKIP_RC=true MAVEN_OPTS="-Xmx512m -XX:MaxPermSize=512m -Duser.language=de -Duser.country=DE" POST_COVERAGE=false SKIP_DEPENDENCY_CHECK=true
```
adamcin: While you are at it, should you consider changing `-XX:MaxPermSize=512m` to `-XX:MaxMetaspaceSize=512m`?

adamcin:
@patrickhousley https://travis-ci.org/Adobe-Consulting-Services/acs-aem-commons/jobs/467732680#L3208 via `org.apache.sling.testing.sling-mock-oak`

.travis.yml
justinedelson: good idea, but that's a separate PR :)

justinedelson:
Thanks @BillDinger. If it is ~/.m2, we're already caching it (see https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/.travis.yml#L17).

justinedelson:
@patrickhousley  I think @adamcin answered your question - there's around a bazillion of those in the log output. I looked a bit at intercepting the way Logback handles this situation, but it'd worse than copying/tweaking this one class.

badvision:
Is this one safe to merge @justinedelson?  Were there any outstanding concerns?

PR #1622, opened 2018-12-13
Update oakpal version to allow guava variation vis. OAK-7182

adamcin:
See adamcin/oakpal#3


adamcin:
@justinedelson oakpal 1.1.12 released with oak 1.8.9 dependency

badvision:
The confirmation build technically passed, it was that email health check that borked on one of the sub-builds.  I restarted it so we have a clean sign-off.

PR #1623, opened 2018-12-14
Potential fix for Health Check Status Emailer Test failures

justinedelson:
This does change the behavior slightly of the Health Check Status Emailer. It now sends the email if the threshold has been reached exactly (whereas before it would happen only if it was at least 1 millisecond after the threshold time). My theory is that this is failing on very fast machines. I've never seen this fail locally, but I guess that just means I need a new computer.


badvision:
Thanks!  This one was a bit of a pest.  Can we run the travis build a few times to confirm this one?

justinedelson:
for sure

badvision:
I've restarted it 3 times and it seems fine.  Thanks!! 👍

badvision:
+1 from me... @davidjgonzalez ?

PR #1626, opened 2018-12-18
Moving to OSGi-annotations.

heervisscher:
This PR is moving acs-commons to OSGi-R6


heervisscher:
This PR is a non-compat-change. Because we can't support anymore '-' in OSGi properties

badvision:
Do we still need that other PR?  There are still merge conflicts on it.

heervisscher:
I closed the other

joerghoh:
just for the records:
* all the references to the scr annotations have been purged
* also the scr plugin is already removed

that means, starting now we only support OSGI annotations

justinedelson:
which are the specific PIDs which have dashes in them? that should go in the CHANGELOG in my opinion.

davidjgonzalez:
All breaking changes should go into a VERY clear documentation page that lists:
1) What Features AND PIDs have changed
2) Old OSGi Property Name
3) New OSGi Property Name

These, of course, need to update in all the feature-specific documentation pages (while moving any old ones to the bottom/out of the way but still available for reference if you're on a < 4.0.0 version.

PR #1627, opened 2018-12-18
Defect/asset ingestor/folders with spec

anasidorovich:
If Preserve File name unchecked, asset and folder names will support only the following characters: letters, digits, hyphens, underscores, another chars will be replaced with hyphens


anasidorovich:
Hi @badvision could you please review changes? 

Thx

cylinder-y:
hello @badvision,
Could you please take a look? We need that feature for our migration
--Yaraslau

PR #1628, opened 2018-12-18
Revert "Moving to OSGi-annotations."

davidjgonzalez:
Reverts Adobe-Consulting-Services/acs-aem-commons#1626

This breaks backwards compatibility; we should think about if we can prevent this from happening.

This merge also forces a 4.0.0 release due to the breaking changes.

We also probably want to put a test/validation strategy around this.


badvision:
The goal is to go to 4.0. I've already introduced breaking changes in the API refactor.

davidjgonzalez:
Lets try to mitigate this if we can - as a user i would not be happy even with a 4.0.0.

davidjgonzalez:
Let's also cut the last bug-fix update before we merge in our breaking changes for 4.0.0 (there's already a number of PR's piling up).  IIRC the other breaking changes are technical breaking, but likely won't effect TOO many ppl... this will effect TONS of people.

justinedelson:
As I commented in #1626, I think the key thing missing here is a list of the breaking changes in the CHANGELOG. #1626 was represented by a single line - "Changing to osgi annotations" with no mention of any breaking changes.

badvision:
We really can't do a release on the old series without backporting/cherry picking at this point.  I introduced breaking changes much earlier than this one in the month.

Issue #1631, opened 2018-12-19
Handle OSGi R6 Breaking changes in a backward compatible manner


davidjgonzalez:
We should investigate if/how we can handle the legacy "-" params for OSGi services after the (currently) breaking changes to OSGi R6.

The  most feasible approach (thanks CZ) looks to be:

> There is no way to use the annotation based approach of defining properties if the property name contains a dash. That's a limitation to avoid complicated mapping rules of allowed java names to property names. As the dot is the most common separator, that one is supported.

> Defining a new property name with a fallback to the old one is only possible, but not so nice. The basic idea is that you define your annotation with property names not using a dash, then in your activate method (or if you're using constructor injection in your constructor) you have at least two parameters, one uses your annotation type, the other one is a map which then gets all configuration properties. Now, you need to check whether the old property is set in the map and use that value - but only if the new property is not set in the map. Otherwise, you can use the value from the annotation.

badvision:
Makes sense for maximum compatibility but this also adds a significant amount of additional technical debt.

davidjgonzalez:
lets try to design a solution to minimized additional technical debt then..   we made this bed, so..

badvision:
I'm all ears, dude. 😉

royteeuwen:
why do you say this does not work? Are you sure that you can't use dashes? did you try doing my$_$property() ? See following documentation:

https://osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#service.component-component.property.types

davidjgonzalez:
@royteeuwen see the discussion at: https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/1567

R7 is only available on 6.4 and ACS Commons supports the current and prior viersion of AEM - so at the time of writing, AEM 6.4 and 6.3.

heervisscher:
Also wondering why there is not an 'internalName' attribute in @ AttributeDefinition, to me this make it also more clear while reading the code what the prop-name is.

joerghoh:
in #1567  I evaluated 2 options how we could deal with it. 
* Switching to R7 annotations does not help, because the annotations write a metatype 1.4 requirement into the XML, which AEM 6.3 cannot handle (6.4 can).
* Sticking with a mixed approach (stay with the old SCR annotations where we would otherwise introduce changes into the property naming) breaks the jacoco/coveralls plugins.

From my point of view this leaves us with 2 options:

[...]

PR #1632, opened 2018-12-19
Revert "Revert "Moving to OSGi-annotations.""

davidjgonzalez:
Reverting this since it seems like the ship has sailed, and new PR's not resolving these merge conflicts will be difficult and error prone.

We should evaluate if #1631  is possible before cutting v4.0.0


badvision:
Honestly I think we just need to take the plunge.  We can't resolve the differences between SCR and OSGi ourselves.

badvision:
Also, if we revert back to SCR annotations, Coveralls will likely stop working again.

davidjgonzalez:
yup .. see "the ship has sailed" .. now we just have to try to mitigate the effects the best we can..

PR #1651, opened 2018-12-22
Asset Ingestor - fix target path if preserveFileName is unchecked

anasidorovich:
If preserveFileName option is unchecked, target path is cut off and Asset Ingestor doesn't work properly.
This is a fix of path validate method.


anasidorovich:
hi @badvision 
please review the fix

badvision:
@cylinder-y Do these changes work for what you intended also?

cylinder-y:
Hello @badvision, 

This pull request is a fix of minor issue that wasn't covered in #1627. So we need this changes

--Yaraslau

PR #1653, opened 2018-12-26
#1585 - Fix Editing with Comments in File

### LINKED ISSUE #1585: [Redirect Map Manager] - Editing a redirect entry causes the previous entry to be edited instead
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: __6.4.2.0 (probably affects earlier versions as well)_____
* [x] ACS AEM Commons Version: ___3.19.0__
* [x] Reproducible on Latest? *yes*

### Expected Behavior

When editing an entry in the redirect map manager, I expect the entry to be edited correctly.

### Actual Behavior

The entry is not edited, instead the _previous_ entry is.

### Steps to Reproduce

1. Follow the docs for: [Redirect Map Manager](https://adobe-consulting-services.github.io/acs-aem-commons/features/redirect-map-manager/index.html)
2. Upload a `.txt` file with the following content:

```
# my test redirect maps
/content/test /
/content/test2 https://google.com
/content/test3 https://bing.com
```
3. try to edit the third entry (`/content/test3 https://bing.com`) in the UI and update the `source` to be `/content/updated`
4. Notice the second entry is updated instead.

@klcodanr:
This issue occurs only if a comment is present in the txt file, it appears the code updates based on line number as an index and does not ignore the comment in the process. So looking at the txt file with line numbers added:

```
[0] # my test redirect maps
[1] /content/test /
[2] /content/test2 https://google.com
[3] /content/test3 https://bing.com
```

updating `[3]` using the UI actually updates `[2]`, because the UI ignores the comment, but the backend code probably does not?

klcodanr:
This fixes #1585 and has some general cleanup of the code.


klcodanr:
@badvision - There shouldn't the only change is an additional field. In fact, the models really aren't meant to be used outside the tool. I can look into moving them into an internal package if that'd be preferred.

badvision:
It is preferred to minimize the external API surface where possible as it makes evolutionary changes more possible within minor releases.

badvision:
Hey @klcodanr!  Just checking in on this one.  Should we keep this PR open while you minimize the API surface or should we consider it as-is for now and merge?

klcodanr:
Hey @badvision  -- So not sure why I thought you could put models in an internal package and have them still work in scripts, but oh well ¯\_(ツ)_/¯

So I just marked the models as `@ProviderType` at least.

PR #1660, opened 2019-01-03
Feature/support content type override

BillBeckerAtBounteous:
Updated EnpointServiceImpl to support the ability to specify a custom content-type header.


badvision:
Isn't this the same as #1657?

PR #1678, opened 2019-01-11
Created log and error output for Asset Ingestor when asset is null

maryia-tatarnia:


cylinder-y:
Hello @badvision,

Could you please take a look into this request and approve it

—Yaraslau

badvision:
I like the idea of this one but the junit tests have to pass also.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/AssetIngestor.java
```diff
@@ -296,11 +296,10 @@ private void createAsset(Source source, String assetPath, ResourceResolver r, bo
                 Asset asset = assetManager.createAsset(assetPath, source.getStream(), type, false);
 
                 if (asset == null) {
```
cylinder-y: In this case report wouldn't be updated. Error will be logged only

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/AssetIngestor.java
```diff
@@ -293,7 +295,9 @@ private void createAsset(Source source, String assetPath, ResourceResolver r, bo
                 }
                 Asset asset = assetManager.createAsset(assetPath, source.getStream(), type, false);
 
-                if (asset != null) {
+                if (asset == null) {
+                    Logger.getLogger(AssetIngestor.class.getName()).log(Level.SEVERE, null, new AssetIngestorException("Cannot create asset: asset is null on path  " + assetPath));
+                } else {
```
cylinder-y: In this case report wouldn't be updated. Error will be logged only

badvision: It could be very well the case that if the unit tests aren't asserting this behavior currently then the unit tests might be what needs adjusting.  I haven't looked to confirm this though.

badvision: Agreed 100%.  However if the unit tests are failing because of the new error reporting then we might have to adjust them to assert this new behavior is correct, that's what I'm saying.

cylinder-y:
Looks good from my point of view,

@badvision could you please confirm and merge PR?

PR #1683, opened 2019-01-13
Feature/httpcache request cookie extension

Sc0rpic0m:
Http cache config extension based on request cookie keys and values.

Allows you to create keys based on a cookie value. Useful for instance in the following scenario which is common:

Login to a middleware (AEM or otherwise) from the front-end 
Login provides the user with a user type or group and saves this to the cookies.
Now we want to see different content (for example on a header menu list of links).

Here we can leverage the cookie value to show different content and cache it.


Sc0rpic0m:
I made the approach more generic and extensible, so I can add a key/value key based on request headers, request parameters, value map values on the resource. Tested this locally and works fine.

Sc0rpic0m:
[https://i.imgur.com/EBgrfHB.png](url) example of usage (JMX)

PR #1686, opened 2019-01-14
Initial implementation of CloseableQueryBuilder.

adamcin:
The QueryBuilder isn't Closeable, the Queries are.

The idea for progressive conversion of usages of QueryBuilder to CloseableQueryBuilder is you start with changing:

```java
@Reference
private QueryBuilder queryBuilder;
```

to 

```java
@Reference
private CloseableQueryBuilder queryBuilder;
```

Since the point of the `CloseableQuery` is to simply add a `close()` method to the `Query` interface, you will probably want to call that method, and the right place to do it with existing logic is _immediately after you are finished using the last `Hit`'s `Resource`_.

```java
Query myQuery = queryBuilder.createQuery(predicates, session);
SearchResult result = myQuery.getResult();
for (Hit hit : result.getHits()) {
  /* do stuff with Hit resources within the try block */ 
}
```

becomes

```java
CloseableQuery myQuery = queryBuilder.createQuery(predicates, session);
SearchResult result = myQuery.getResult();
for (Hit hit : result.getHits()) {
  /* do stuff with Hit resources within the try block */ 
}

myQuery.close();
// don't use Hit Resources from this result anymore.
```

CloseableQueryBuilderImpl wraps a reference to QueryBuilder and passes through all the expected arguments to `createQuery()`, and then wraps the returned `Query` type with the `CloseableQuery` type, which also passes through all the Query interface methods. 

The difference is that `CloseableQuery` implements `java.io.Closeable`, so that the next step after conversion would be to encapsulate all the `Query` logic into a _try-with-resources_ block, like so:

```java
try (CloseableQuery query = queryBuilder.createQuery(predicates, session)) {
  SearchResult result = query.getResult();
  for (Hit hit : result.getHits()) {
    /* do stuff with Hit resources within the try block */
  }
}

/* once the try block exits, the query's internal ResourceResolver will be closed */
```

The `CloseableQueryBuilder` also adds convenient `ResourceResolver` overloads for all the `QueryBuilder` methods that take `Session` arguments, so that you don't have to adapt your `ResourceResolver` to a `Session`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 - #1654 - Added I18nProvider service to support injectors
 - #1648 - Add Smart Tags to XMP Metadata Node Workflow Process
 - #1670 - Added @JsonValueMapValue, @I18N, @HierarchicalPageProperty, and improved @AemObject and @SharedValueMapValue.
+- Added CloseableQueryBuilder service to deal with CQ QueryBuilder's shallow unclosed ResourceResolvers.
 
 ### Fixed
 - #1667 - Refactored the activate methods of all http cache services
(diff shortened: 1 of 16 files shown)


joerghoh:
Can you give an example how one can convert an existing usage of the QueryBuilder to the CloseableQueryBuilder?

adamcin:
@joerghoh I added a description for use. I'm still adding test coverage to prove it all out, but is the description sufficient?

joerghoh:
Wow, that's a documentation! Absolutely great! Thanks @adamcin

badvision:
This is full of so much win.   Keep up the good work @adamcin !!!! 😄

adamcin:
@badvision I've got a successful build and what should be 100% coverage for the included files. Please merge.

Issue #1539, opened 2018-11-01
Review use of QueryBuilder and prevent ResourceResolver leakage


davidjgonzalez:
Use of QueryBuilder APIs for querying often results in a leaking ResourceResolver.

The ACS Commons code-base should be reviewed for use of QueryBuilder and any leakage should be mitigated.

adamcin:
@davidjgonzalez I think we should reopen this issue after merging #1686 . The CloseableQueryBuilder should ease mitigation in most places.

davidjgonzalez:
@adamcin Do you recall where this was left off? I thought you actually ended up going through and updated the uses of QueryBuilder to use CloseableQueryBuilder. Do you recall if you only tackled a subset?

adamcin:
@davidjgonzalez I believe I converted all usages of QueryBuilder that were present at the time.

davidjgonzalez:
thanks @adamcin took a quick scan and looks like you did; there were a few unused references left but easy enough to clean up in #1794 - thanks!!!

davidjgonzalez:
This was fixed w/ CloseableQueryBuilder

PR #1698, opened 2019-01-15
Issue #1539 replace encapsulated usage of QB with CloseableQB.

adamcin:
See #1686 for background. 

TODO:

evaluate content/src/main/content/jcr_root/apps/acs-commons/granite/ui/components/form/queryautocomplete/datasource/datasource.jsp for similar treatment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 - #1667 - Refactored the activate methods of all http cache services
 - #1664 - OSGI annotations : fixed default values for various activate methods
 - #1607 - HttpCache: improved the write to response mechanism.
+- #1539 - Reviewed usages of QueryBuilder for ResourceResolver leakages and close leaks.
 - #1590 - Multifield component doesn't render non-composite at all (NPE error)
 - #1588 - Updated error handler JSP to use ModeUtils
 - #1583 - Asset Ingestor may try to create asset folders when they already exist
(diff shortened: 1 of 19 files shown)


badvision:
I like where this is headed.  Couldn't Query implement Closeable and therefore allow us to use try-with-resources (autoclose)?

adamcin:
@badvision Adobe could fix the issue in a deeper way that wouldn't require closing the Hit ResourceResolver, but only if it changed the QueryBuilder API to accept the original ResourceResolver in the first place instead of the underlying JCR Session. I added the CloseableQuery interface under the assumption that that kind of API refactoring likely won't happen quickly.

badvision:
Sure but I noticed a lot of cases where the results are used within a try/catch anyway, so using Closable might tidy things up a little bit.

adamcin:
@badvision Yes, in some cases, a type may not have access to a `@Reference QueryBuilder`, like `OnDeployScriptBase`, where instead the QueryBuilder is retrievable only by using `ResourceResolver.adaptTo(QueryBuilder.class)`. I didn't want to go so far as to implement an AdapterFactory just for this situation, because I don't like that pattern for retrieving a QueryBuilder to begin with (nothing about the ResourceResolver being adapted has anything to do with the QueryBuilder being returned). 

The other case is the `QueryBuilderViewQuery`, which accepts a query in its constructor. Since the type is provided in a constructor, I can't create it in a try-with-resources block with in the execute method without changing a bunch of other logic outside the class. It's easier just to use try/finally block and close the query explicitly.

badvision:
Well it's your option to take it further or leave as-is.  I think it's a step in a great direction either way.

badvision:
Nice.  👍 from me.  @justinedelson any thoughts on this one?

PR #1707, opened 2019-01-21
Feature/aem remote assets

HitmanInWis:
This is reopen of https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/1294 - the PR was previously approved, but was lacking junit tests and thus went stale and never got merged.

As a reminder, this is a PR for the "AEM Remote Assets" feature presented at 2018 Adobe Summit AEM Rockstar.

PR now has strong junit code coverage, and has been tested to work on AEM 6.3 and 6.4.

There are multiple use cases for this feature.
- Keeping Dev/QA/Stage servers up to date with Production assets
- Local servers used for development or other business functions
- Assets stored in a non-Adobe DAM or even a PIM (NOTE: this would require further development to implement "connectors" to those different systems)
- Standalone AEM Assets instance with separate AEM Sites instance (until Adobe releases a solution for this)

We currently have a client requesting this to be installed on their DEV servers to keep assets in sync w/o pulling the entire DAM, so we're looking forward to getting this officially merged and potentially enhanced further.  I've also had other companies reach out to me directly regarding this feature.

One change I am considering is renaming this feature to be more accurate to what it is.  Though the Summit Rock Star feature was termed "Remote Assets", perhaps this would better be termed "Asset Auto-Sync" or something along those lines.  That way if AEM were to ever support assets from a truly remote DAM to be used directly it would avoid confusion.


bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncScheduler.java
```diff
@@ -0,0 +1,75 @@
...
+import org.osgi.service.metatype.annotations.ObjectClassDefinition;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Scheduled Service that runs the Remote Assets node sync.
+ *
+ * This job will sync asset nodes based on OSGi configuration.
+ */
+@Component(
```
HitmanInWis: For the life of me I cannot figure out why I can't get this scheduler to run.  I've added a config via felix console (to satisfy `configurationPolicy=REQUIRE` and used a scheduler expression of `0 * * ? * *` (every minute, just for testing purposes) but it never runs.  It seems every other scheduler in ACS Commons has slightly different Annotations as well, so I'm not sure which pattern to follow.

Anyone have any ideas on what I have incorrect here?

HitmanInWis: Do I maybe need a reference to a `Scheduler` and then need to add/remove my object?  I dont see any of the other schedulers in ACS Commons doing that though...

HitmanInWis:
The failing test is confusing...
```
[ERROR] testGetResourceSyncsRemoteAsset(com.adobe.acs.commons.remoteassets.impl.RemoteAssetDecoratorTest)  Time elapsed: 1.776 s  <<< ERROR!
java.lang.RuntimeException: Unable to initialize JCR_MOCK resource resolver factory: Unable to inject mandatory reference 'resourceAccessSecurityTracker' for class org.apache.sling.resourceresolver.impl.ResourceResolverFactoryActivator : no matching services were found.
Caused by: org.apache.sling.testing.mock.osgi.ReferenceViolationException: Unable to inject mandatory reference 'resourceAccessSecurityTracker' for class org.apache.sling.resourceresolver.impl.ResourceResolverFactoryActivator : no matching services were found.
```

It appears to happen intermittently.  I wonder if there's some sort of issue being caused occasionally by PowerMockito acting on the test class?  I cannot reproduce this locally running the tests from IntelliJ, so Im guessing it has something to do with the order of the tests or something just a little bit different about how maven runs the test vs. IntelliJ...

dmitry-cherkas:
@HitmanInWis ,
a64461387 revision adds 'cq:tags' directory, which does not allow to checkout the branch on windows, since colon is not allowed in directory names:
`fatal: cannot create directory at 'content/src/main/content/jcr_root/content/cq:tags': Invalid argument`

HitmanInWis:
Sorry @dmitry-cherkas, this is now resolved.

badvision:
Are there any reasonable limits someone should know about when using this?  e.g. if you try to sync over 1000 assets you have to update the scheduler, etc?

HitmanInWis:
@badvision I imagine there might be an upper limit to making it feasible, but that's going to depend largely on connection speed between the two servers.  Running local instance to local instance I can sync the full node structure of a DAM with about 1000 assets in just over a minute.  The sync process saves incrementally after every 100 assets, so given enough time it should be able to chug thru a DAM of pretty large size before it becomes unfeasible.

content/src/main/content/jcr_root/etc/tags/_rep_policy.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
HitmanInWis: Is there a way to deploy this permission in a way that only works if the /etc/tags folder already exists?  When I deploy this to AEM 6.4, the next time I restart the AEM server it sees the /etc/tags folder and assumes my tags are there instead of under /content/cq:tags.

badvision:
> @badvision I imagine there might be an upper limit to making it feasible, but that's going to depend largely on connection speed between the two servers. Running local instance to local instance I can sync the full node structure of a DAM with about 1000 assets in just over a minute. The sync process saves incrementally after every 100 assets, so given enough time it should be able to chug thru a DAM of pretty large size before it becomes unfeasible.

I was able to cause issues in the previous version after ingesting 5000 assets in one server.  I'm happy to share my test data if it would help.

HitmanInWis:
Sure, any info you can give to help look deeper into it would be appreciated.  What types of "issues" were you seeing?

I simply tested with the we-retail folder (243 assets) copied into 4 folders to equal about 1000 assets.

HitmanInWis:
Also, please dont forget:

**For anyone testing:** Please make sure all of your asset workflow launchers have event-user-data:changedByWorkflowProcess in the "Exclude List" so that the sync does not trigger workflows. Unfortunately I couldn't find an easier way to avoid triggering workflows, so this will be part of configuration instructions.

If you didnt do this step, you may have triggered a bunch of asset workflows, which is unnecessary (we're copying the nodes verbatim) and ultimately beats on your server very unnecessarily.

badvision:
Actually the launcher bit might have been the issue.  My server (well, laptop) was rather unresponsive.  This makes a lot more sense that it could have been workflow-related.  Would you entertain the notion of a one-time health-check that runs to see if there are asset workflows that do not have the exclusionary flag set properly?  e.g. if this module is activated, then it runs a health-check so that people are somehow warned about this automatically?

HitmanInWis:
I certainly could look into it - I've never written a health-check before.  I also wonder if it would be ignored (do people regularly check those?), and if so then it's basically wasted effort.  Ultimately people will need to read the manual before implementing this feature, which I plan to mark clearly as "FOR NON-PRODUCTION USE" on the instructions page for this first release.

Out of curiosity, do you know of a better way to prevent this feature from triggering workflows?  I'd love it if I could somehow block all launchers from executing by the `acs-commons-remote-assets-service` user.

badvision:
Health-checks can trigger notifications if the server is properly configured, e.g. with external monitoring.  In a more production-like environment you'd certainly hope for that. ;)  As for preventing workflow, no I also use the same "started by workflow" flag in a lot of my stuff too since asset ingestion in terabyte-scale workloads will really hurt the server (badly) if you let it trigger workflows.

badvision:
Also the "Do not use in production" moniker resolves my other concerns.  For non-production use this is a very compelling feature indeed!  It certainly solves a common assets project concern about dealing with test data.

HitmanInWis:
Thanks @badvision.  Assuming the feature largely looks good to go for "non-production" use, any quick thoughts on why the Scheduler class is currently not working?  I'll try to find some more time to dig into it, but thought I'd ask in case it's something obvious :).  To be fair, I tried configuring other "Scheduler" classes in ACS Commons on my 6.4 instance but it didnt seem any of them were running, so maybe something is screwy on my server.

Also, instead of using rep:policy nodes for service user permissions, I wonder if I should instead use EnsureServiceUser to both create the user and assign permissions.  That way the deployment wont create a /content/cq:tags folder on AEM 6.3 instances and conversely create an /etc/tags folder on AEM 6.4 instances...

content/src/main/content/jcr_root/etc/tags/_rep_policy.xml
HitmanInWis: Maybe I just need to switch over to using `EnsureServiceUser`...

davidjgonzalez:
Closing to clean-up PR backlog, to help focus on v4.0.0 release. 

Non-4.0.0 release blocking PR's will not be merged until 4.0.0 release anyhow.

We can re-open post 4.0.0 release.

HitmanInWis:
@davidjgonzalez why is this being closed versus left open with the "Do NOT merge until 4.0.0 release" label like the other currently active PRs?

HitmanInWis:
@davidjgonzalez Now that 4.0.0 is released, can this be reopened?

HitmanInWis:
@badvision @davidjgonzalez FWIW I was trying to get this completed in time for Summit this year, as I know people will ask me about it.  One very large company even reached out to me with their team of engineers looking for this feature.  Even if 4.1.0 wont be officially released by Summit, it would be great if I could point people to ACS Commons and indicate it will be in the next release.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncScheduler.java
HitmanInWis: This is now resolved.

content/src/main/content/jcr_root/etc/tags/_rep_policy.xml
HitmanInWis: This is now resolved.

badvision:
I think this was closed only because we had a lot of small fires to put out.  Do we have a pr in the docs project to explain this one?  I think the dev/QA use cases were rather compelling!

HitmanInWis:
Sounds good @badvision, thanks.  I'm going to get onto the documentation task next - finally had time to sort out a couple final issues on this PR.  Will link to it when ready, hopefully in the next day or two.

HitmanInWis:
The single remaining codeclimate issue I dont know a good way to resolve without making the code more obscure - to me its most clear the way it is, but happy to consider refactoring.

I plan to have the documentation PR up this afternoon or sometime tomorrow.  Just waiting on creative team to create a thumbnail for the feature :)

HitmanInWis:
And yay...if this gets merged I get to lay claim to bringing ACS Commons code coverage over 50% 😛

HitmanInWis:
Documentation PR complete!

badvision:
@HitmanInWis -- any chance you can check into that switch statement complaint from Code Climate?

HitmanInWis:
@badvision The single remaining codeclimate issue I dont know a good way to resolve without making the code more obscure - to me its most clear the way it is, but if you insist it must be addressed I can take another look.

HitmanInWis:
For example, I could put a bunch of callback functions (or objects with a function) into a map, fetch them from the map by key, and then call the function, but I'm not sure how much more "clear" that makes the code.  In some ways it could be considered more obscure that way.

badvision:
I'm okay with adding a comment to disable that check for that one method. ;)

badvision:
Alright.  Here we go! 😄

HitmanInWis:
weeeee, thanks @badvision !  It's been a long time coming, just in time for Summit :)

PR #1712, opened 2019-01-22
Issues/1706

heervisscher:


badvision:
I think this was made on top of the master branch and reverses some of the changes in the OSGi7 branch, such as updated pom changes and so on.

Issue #1614, opened 2018-12-12
Disable all Clientlibs by default


davidjgonzalez:
## Required Information

* [X] AEM Version, including Service Packs, Cumulative Fix Packs, etc: All
* [X] ACS AEM Commons Version: All
* [X] Reproducible on Latest? yes

We have several features that inject JavaScript into OOTB client library categories by default (opposed to the opt-in method of all other ACS Commons) features. This is problematic in that this JS runs even if the consumer does not want the feature (unnecessary code execution) and this can also occasionally break or surface breaking behavior in OOTB features [#1602](https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/1602).

Ideally, all JS/Clientlibs are "off by default" and a consumer of ACS Commons must actively enable them. This is a breaking change, so this should be part of the 4.0.0 release.

## Possible solutions 

1) We have the DynamicTouchUIClientLibLoader [944](https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/944) and [1199](https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/1199) that can be used to disable ClientLiib inclusion, but this assumes the Clientlibs are ON by default. We could enable this by default, and allow users of the features to override the defaults, however, this now results in the DynamicTouchUIClientLibLoader feature being enabled by default. I

2) Do not include our clientlibs into default categories by default, and require a consumer to do this manually if they use the feature. (XML inclusion node definitions to be included in the docs).
     1) Reset our cq:clientLibraryFolder's to have an acs-commons specific category (`acs-commons.limited-parsys`)
    2)  Provide instructions that to enable this feature, you have to deploy your own cq:clientLibraryFolder definitions that looks like
  `/apps/my-project/clientlibs/vendor/acs-commons-limited-parsys` => `category: <the OOTB category>`, `embed: acs-commons.limited-parsys`

schoudry:
or just thinking out loud...

3.a. Remove the otb categories added in all UI extensions and may be add `categories=enable.this.extension.in.osgi`

b. Add an OSGI service named "ACS Commons Enable Touch UI extensions" with a list of all UI extensions added as boolean `org.apache.felix.scr.annotations.Property`  

c. Enabling an extension would add the specific categories in respective clientlib location under /apps/acs-commons...

Similarly add an OSGI service for Classic Extensions, test upgrades etc....

davidjgonzalez:
[...]

PR #1715, opened 2019-01-24
#1614 - (Breaking change) Disables all auto-on clientlibs by default,…

davidjgonzalez:
… requiring proxy clientlibs.

This disables ACS clientlibs from being "on by default" which has problematic for quite some time.  
This is a BREAKING change for 4.0.0, and requires clear documentation on which features are broken, and the required proxy clientlib definitions to fix. (the proxy clientlib definition XML is in the updated XML files in the commit for convenience)

* TouchUI Audio Asset Finder
* TouchUI DAM Copy Publish URL
* TouchUI MetadataEditor DAM Location
* TouchUI DAM Custom Component Activator
* TouchUI WCM Switch Layers
* TouchUI WCM Editor Styles for Audio (Sidepanel)
* TouchUI RTE Plugin - Dialog Plugin
* TouchUI Composite Multifield
* TouchUI Widgets
   * icon-picker.js / jquery.fonticonpicker.coral.less
   * touchui-limit-parsys.js
   * touchui-configure-parsys-placeholder.js
   * search-based-pathbrowser.js
   * draggablelist.js / draggablelist.css
   * dynamic-loader.js
* ClassicUI Widgets 
   * add_multifield_limit.js
   * create_namespace.js
   * predicates/AuthorizablePredicate.js
   * predicates/PathPropertyPredicate.js
   * predicates/SelectionPredicate.js
   * DDPathField.js
   * polyfill_composite_field_processParentRecord.js
   * multipanel.js
   * node-multi-panel.js
   * check_vanity_path.js
   * validate_responsive_column_control_dialog.js
   * add_dynamic_options_to_feed_importer.js
   * add_language_titles_to_item_dialog.js
   * icon-picker.js
   * dynamic-loader.js
   * cq-placeholders.less
   * icon-picker.css
* ClassicUI Widgets Themes (Contentfinder Audio)

@justinedelson, @schoudry and @suryalamech  could you take a quick pass and see if this look good. Are there any we are super comfortable leaving on by default; I think we'd have to be positive it won't break anything (ie. the CSS for the audio tab) and/or can we consolidate any of these into a single category and/or leverage dependencies?

The two areas that seem most likely candidates for consolidation are the Audio and DAM Widgetry that I think @justinedelson mainly worked on.

Ill document these changes in each feature doc page as well as an "Upgrading to v4.0.0 Readme" w/ easy to copy/paste XML proxy snippets.


badvision:
Works for me but it will require the requisite documentation to go with this change otherwise people will never know these things exist.

davidjgonzalez:
@kwin Sure I can add allowProxy=true to both the actual clientlib, as well as (and probably more importantly) the documented proxy XML snippet. I saw more importantly, because the proxy clientlib embeds the ACS clientlib, so that should never be directly accessed anyhow. EOD it will be up to the user to put the proxy in the correct location (which we can guide them to).


@badvision wrt to documntation, this information will be in 2 places:

1. A centralized v4.0.0 documentation page covering the impact of this release
2. Each feature doc page that is effected.

badvision:
I'm all for it.  Right now we have a release blocker where all the PRs are failing for the same reason.  We can merge this now if you want, but it might be more prudent to fix the build breaker bugs first.

davidjgonzalez:
I’ll update with allow proxy today or tomorrow - this shouldn’t result in merge conflicts anyhow so no rush on this

content/src/main/content/jcr_root/apps/acs-commons/authoring/audio-asset-finder/.content.xml
```diff
@@ -1,6 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<jcr:root xmlns:cq="http://www.day.com/jcr/cq/1.0" xmlns:jcr="http://www.jcp.org/jcr/1.0"
+          jcr:primaryType="cq:ClientLibraryFolder"
+          jsProcessor="[default:none,min:gcc]"
+          allowProxy="{Boolean}true"
+          categories="[acs-commons.authoring.audio-asset-finder]"
+          dependencies="[cq.authoring.editor]"/>
+<!--
+TODO Is this OK?
```
davidjgonzalez: @justinedelson  does this look ok to you? i don't recall off the top of my head how the "hook" clientlibs work.

justinedelson: @davidjgonzalez so the idea is that a "consumer" of ACS AEM Commons would create a "proxy" client library at the location of their choosing with the category 'cq.authoring.edit.hook.assetfinder` and embed `acs-commons.authoring.audio-asset-finder`?

If so, yes, that should work.

justinedelson:
This all looks OK to me. One documentation comment -- to avoid confusion, we shouldn't call the consumer-created client libraries "proxy client libraries" since that will be confused (at least _I_ was confused) with the `allowProxy` setting.

I would suggest referring to these as "Wrapper Client Libraries" (not the best term, but I don't have any other ideas).

davidjgonzalez:
@justinedelson works for me - updated the comment name to "Wrapper Client Library... " in the comments and will use same in docs.

schoudry:
@davidjgonzalez looks good to me (but this is going to break a LOT of functionality if users miss reading the documentation and directly go to github and raise tickets)

i guess we can leave a `console.log("READ DOCUMENTATION - <link>")` for existing acs commons clientlib categories with dependencies say `cq.authoring.editor`

davidjgonzalez:
@schoudry yup! on a more positive note, we'll stop breaking all those people that aren't using these features! This is a damned if we do, damned if we don't situation IMO. (generally, I loathe to break BC, but I loathe "accidentally" breaking ppl that don't even want a feature even more)

Historically (at one point i started tagging them in GH issues) turning these on by default has resulted in a massive number of issues, and most unfortunately for folks that don't even WANT to use these features.

Agree it's not ideal. 

My suspicion is these 2 are are the main ones in use:

* TouchUI Composite Multifield
* TouchUI Widgets

And w/ the Multifield we encourage ppl to move to the OOTB (understanding its hard for existing customers w content created by this impl)

schoudry:
yes totally agree half glass full guy :)

PR #1723, opened 2019-01-25
Issue/fix unclosed channel when non exising path provided

cylinder-y:
sessions leaks removed

PR #1729, opened 2019-01-28
Url Asset Ingestor | Support case sensitive properties

cylinder-y:
This is improvement of Url Asset Ingestor to support case sensitive properties. Now any header is converted to lowercase
If we provide the following excel header:

|Source | Target | Rendition | Original | dc:description | dc:title | test:aspectRatio
| -- | -- | -- | -- | -- | -- | -- |
|/some/source/test.jpg|/some/target||| some description | some title | 16:9 |

We will get image node with the following metadata properties:

|Property Name|Property Value|
| --|--|
|dc:description|some description|
|dc:title|some title|
|test:aspectRatio|16:9|


badvision:
This is something I had done originally but retracted: property names are inherently case sensitive in JCR and really we have to enforce that.  The only reason that source/destination are case insensitive is that they are more logical constructs of the import process, so we can make those case insensitive.  Property names, however, have to be case sensitive in order for users to enforce consistent data.  For example, if they're using all upper-case property names then they have to use all upper-case header names too.  It's not for this tool to dictate what the stored property names should be because it could cause a mismatch (and previously it had done exactly that, which is why I took the case conversion logic out.)

cylinder-y:
hi @badvision,
I have implemented check box, that is responsible for headers case sensitivity.

According to your comment, should I just use `Spreadsheet `constructor with `convertHeaderNames`= false?

`fileData = new Spreadsheet(false, importFile);`

In this case migration user will be always responsible for correct configuration of the headers

cylinder-y:
ok, now I got it.

cylinder-y:
Hi @badvision,
I have removed previously added checkbox field from ingestor.
New configuration with list of case insensitive headers was implemented

Build fails because of master branch

badvision:
Yeah I have build failures in my feature branch for the same reason.  Standby until we fix.  Sorry about that!   If you have any insight as to why the mocks/spys might not work in Travis-CI, I'm all ears!!

cylinder-y:
@badvision, mocks/spys issue is related to another PR #1723 :)

Issue #1730, opened 2019-01-28
MCP Multifield throws init error on array fields


badvision:
Multifield works on collection fields but init bombs with a class cast exception for arrays.

PR #1731, opened 2019-01-28
Feature/bugfix multifield support arrays

badvision:
Bugfix for #1730


justinedelson:
@badvision I think this needs to be rebased.

badvision:
boy howdy... yeah I'm still working on it but that's a good idea.

badvision:
That was one ugly rebase but I think we're in the clear on this branch.  I'll do some additional testing to be sure.  It also probably needs more test coverage before I can call it done because there was some additional logic added.

justinedelson:
don't feel great about merging this because of the decrease in coverage (despite my last ditch effort there to increase the coverage by testing the Select component), but it is a major API change so better to do it with 4.0 than wait.

PR #1735, opened 2019-01-30
Feature/1713 etag servlet filter

### LINKED ISSUE #1713: ETag servlet filter (generating the ETag response header based on MD5 checksum)
To better support [conditional requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Conditional_requests) it would be nice if there would be an optional servlet filter  provided which generates the `ETag` response header based on the calculated MD5 checksum of the whole response.
That way the ETag would be stable among several publishers in a server farm for the same request path (as long as they deliver the same response) and would change whenever something changes about its content. Leveraging only the local properties of the current node would not consider i.e. code changes in the markup, therefore generating a checksum on top of the actual response is the only reliable way to get different values for different responses.

This is also mentioned in https://helpx.adobe.com/experience-manager/kb/optimizing-aem-site-caches.html#AEMPublishInstances.

If you use a dispatcher you would need to enable ETag header caching in addition (compare with https://helpx.adobe.com/experience-manager/dispatcher/using/dispatcher-configuration.html#CachingHTTPResponseHeaders)

kwin:
This fixes #1713


kwin:
@justinedelson or @davidjgonzalez Any chance you can merge that for the upcoming 4.0?

joerghoh:
This filter has problems with large responses. As soon as the response is committed (thus the response headers have been sent) you cannot send the etag header. It depends on the size of the response buffer (I found a default of 32kbyte for jetty in some documentation). 

As the response size is highly dependent on the page being rendered, it is not predictable if the header can be sent or not. 

The only way to make this work reliably is the avoid the streaming of the response and buffer the complete response in memory, calculate the etag and then send everything at once. But that comes at the cost of an increased memory usage and overall slower responses.

Another way would be to have an etag already present at the begin of the response processing, maybe calculating it based on jcr:lastModified and other properties. But I haven't thought through all of its implications.

kwin:
> The only way to make this work reliably is the avoid the streaming of the response and buffer the complete response in memory, calculate the etag and then send everything at once. But that comes at the cost of an increased memory usage and overall slower responses.

This is already done in the PR with the help of https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/1735/files#diff-e7d454feb1a1f29bca3282857e7ba2b3R39. It also prevents `flushBuffer()` calls (https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/1735/files#diff-e7d454feb1a1f29bca3282857e7ba2b3R170). Is there any request where the ETag generation does not work for you?

joerghoh:
Ah, missed that. I was confused by https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/1735/files#diff-002f8fa44b07864d1938bc8f0265143eR129 which checks if the response is already comitted. Is this check required then?

kwin:
The check is there to help debugging issues in case another filter with higher priority already committed the response for some reason.

bundle/src/main/java/com/adobe/acs/commons/etag/impl/EtagMessageDigestServletFilter.java
```diff
@@ -0,0 +1,178 @@
...
+                try (BufferedSlingHttpServletResponse bufferedResponse = new BufferedSlingHttpServletResponse((SlingHttpServletResponse) response)) {
+                    chain.doFilter(request, bufferedResponse);
+                    if (!configuration.overwrite() && slingHttpServletResponse.containsHeader(HttpConstants.HEADER_ETAG)) {
+                        log.debug("Do not overwrite existing ETag header with value '{}'", slingHttpServletResponse.getHeader(HttpConstants.HEADER_ETAG));
+                    } else {
+                        if (slingHttpServletResponse.isCommitted()) {
+                            log.error("Can not send ETag header because response is already committed, try to give this filter a higher ranking!");
+                        } else {
+                          String digest = calculateDigestFromResponse(bufferedResponse);
+                          slingHttpServletResponse.setHeader(HttpConstants.HEADER_ETAG, digest);
```
joerghoh: Can you log this digest to the request progress tracker as well?

kwin: Done

badvision:
Is there anything you can do to raise the code coverage up just a tad bit @kwin?  Otherwise seems fine to me.

kwin:
@badvision I added some more tests, now coveralls is happy as well.

PR #1736, opened 2019-01-30
Catch ReplicationExceptions in the Dispatcher Flush Rules Preprocessor

### LINKED ISSUE #1733: DispatcherFlushRulesImpl (Replication Preprocessor) should not throw an exception in case no flush agent is found
## Required Information

* AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.4.1
* ACS AEM Commons Version: 3.19 
* Reproducible on Latest? yes

### Expected Behavior

In case a dispatcher flush rule is configured but there is no flush agent found other agents should continue to work.

### Actual Behavior

In case there is no flush agent found there is a `AgentNotFoundException` being  thrown  from https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/84f76f7faccca34062008ea61499e2dd201c7a90/bundle/src/main/java/com/adobe/acs/commons/replication/dispatcher/impl/DispatcherFlusherImpl.java#L104. That is not caught in the calling `DispatcherFlushRulesImpl` either, leading to the exception being thrown from the Preprocessor (i.e. from https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/84f76f7faccca34062008ea61499e2dd201c7a90/bundle/src/main/java/com/adobe/acs/commons/replication/dispatcher/impl/DispatcherFlushRulesImpl.java#L163). Any exceptions being thrown from a `Preprocessor` prevent any agents from being called (at least with AEM 6.4.1).

### Steps to Reproduce

Set up a standard (non flush) agent also on the AEM publish instance and disable any flush agents there. Then enable the [Dispatcher Flush Rules](https://adobe-consulting-services.github.io/acs-aem-commons/features/dispatcher-flush-rules/index.html). Now activate some path which is covered by the rules and observe that the other agent on the publish is no longer triggered due to the `AgentNotFoundException` from the `DispatcherFlusherImpl`

Although one could argue that in such a case the configuration of the Dispatcher Flush Rules should not be done, I think the Preprocessor should be more robust and actually catch that exception instead of letting it prevent other agents from being executed.

kwin:
This closes #1733

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 - #1700 - MCP Forms framework now tracks client libraries required for components as needed
 
 ### Fixed
+- #1733 - Do not throw ReplicationExceptions from Dispatcher Flush Rules Preprocessor
 - #1745 - Show/hide widgets: feature can now also show/hide complex fields like Image or FileUpload
 - #1724 - AemEnvironmentIndicatorFilterTest.testDisallowedWcmMode is failed because of caret in windows
 - #1699 - MCP UI doesn't work because of StackOverflowError exception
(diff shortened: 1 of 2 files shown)


justinedelson:
@kwin will you have a chance to fix the codeclimate issue?

kwin:
@justinedelson I just fixed it

justinedelson:
Thanks!

PR #1739, opened 2019-02-03
fixed unit tests

joerghoh:
I got this working on a branch created from master.

Regarding the root cause see https://issues.apache.org/jira/browse/SLING-8263, OsgiMocks cannot handle all r7 features properly.

### DIFF
--- a/bundle/pom.xml
+++ b/bundle/pom.xml
@@ -573,6 +573,16 @@
             <version>1.0.0</version>
             <scope>provided</scope>
         </dependency>
+        
+        <!--  We need to have this package on the build-path to avoid bnd complaining about
+              missing annotations (in the uber.jar these annotations are referenced)     
+         -->
+        <dependency>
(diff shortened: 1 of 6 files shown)


bundle/pom.xml
```diff
@@ -573,6 +573,16 @@
             <version>1.0.0</version>
             <scope>provided</scope>
         </dependency>
+        
+        <!--  We need to have this package on the build-path to avoid bnd complaining about
+              missing annotations (in the uber.jar these annotations are referenced)
```
kwin: maybe it should be clarified here, that this is only necessary due to the option `<_dsannotations-options>inherit</_dsannotations-options>` and the fact that we have OSGi components which inherit from legacy Felix SCR annotated classes.

davidjgonzalez:
Closing as we reverted back to SCR annotations: https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/1741#issuecomment-461836942

Issue #1742, opened 2019-02-05
acs-aem-commons not compatible with with AEM 6.5


gkolluri:
## Required Information

AEM Version: AEM 6.5 load 20
ACS AEM Commons Version:  3.19.0
Reproducible on Latest? yes

### Expected Behavior

com.adobe.acs.acs-aem-commons-bundle is active 

### Actual Behavior

com.adobe.acs.acs-aem-commons-bundle failed due to missing dependencies as mentioned below

```
...
com.adobe.cq.dialogconversion,version=[1.0,2) -- Cannot be resolved but is not required
com.amazonaws.auth,version=[1.10,2) -- Cannot be resolved but is not required
com.amazonaws.services.s3,version=[1.10,2) -- Cannot be resolved but is not required
com.amazonaws.services.s3.model,version=[1.10,2) -- Cannot be resolved but is not required
org.apache.felix.scr,version=[1.8,2) -- Cannot be resolved but is not required
org.apache.sling.xss,version=[1.1,2) -- Cannot be resolved
sun.misc -- Cannot be resolved but is not required and overwritten by Boot Delegation
```

### Steps to Reproduce

1. Install vanilla instance of AEM 65-load20 
2. Install latest version of acs-aem-commons v3.19.0
3. Inspect the acs aem commons bundle in felix

davidjgonzalez:
ASC ran into this issue as well, as we were able to work around by expanding the import range. [1]

ASC uses a single method call once, and this method call was not broken going to v3 of xss, so OMMV.

[1]  https://github.com/Adobe-Marketing-Cloud/asset-share-commons/commit/a98551bc68c5accddc9c3ea08514595252a7ca9e

jfmitchell:
https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/1743 open for this

justinedelson:
It appears the only problematic class in this package is JSONUtil.

justinedelson:
I've applied this change on the release/4.x branch

kelvinxzf:
thanks @justinedelson. do we have an estimate when the next release would come?

tarunprakash11:
I am facing this issue with AEM 6.5 SP3. Can anyone point me to the solution for this ?

joerghoh:
can you please raise a new issue for this? Please also clarify what you mean with "not compatible" in your specific aspect. Thanks!

PR #1743, opened 2019-02-06
#1742 Allowing wider range of XSS version support

jfmitchell:
This range of versions is required for future version support.


justinedelson:
already incorporated

Issue #1716, opened 2019-01-24
Set mandatory short-name element on tld files


kwin:
The latest master contains a lot of TLD files below https://github.com/Adobe-Consulting-Services/acs-aem-commons/tree/master/bundle/src/main/resources/META-INF. All of them reference the 2.1 taglib schema (http://java.sun.com/xml/ns/javaee/web-jsptaglibrary_2_1.xsd) in which the element `short-name` is mandatory.
Still the `short-name` is set in none of the TLD files, which leads to validation errors.

In the past all TLD files were generated only from annotations with the tld-generator annotation parser, but in commit https://github.com/Adobe-Consulting-Services/acs-aem-commons/commit/f3ba4f3dbeb23f8f8c92301ab3b2bb053c9fa275 all of those tld files were committed to the repo as well.

Please either fix the TLDs to make them have the mandatory `short-name` or preferably rely on TLD generation only.

kwin:
I think this should be easy fix and I would like to see this in the 4.0.0 milestone. But maybe @badvision can answer, why those TLDs are no longer only generated but really committed to git (this feels redundant as we still have the TLD annotations)

davidjgonzalez:
@badvision can you shed light on this?

badvision:
The TLD generation stopped working and the files were committed in order to
patch the behavior.  The thinking was that jsp development is less likely
to occur and we're maintaining legacy tag libraries rather than development
new ones.

kwin:
Ok, if it stopped working there is a lot more code (i.e. annotations and also plugins) you should get rid of.

PR #1748, opened 2019-02-08
add short names and remove tldgen (which is non-functional anyway). f…

justinedelson:
…ixes #1716

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/util/TemplateUtil.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/util/TemplateUtil.java
@@ -21,7 +21,6 @@ package com.adobe.acs.commons.util;
 
 import javax.annotation.CheckForNull;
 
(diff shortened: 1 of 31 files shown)


bundle/src/main/java/com/adobe/acs/commons/util/TemplateUtil.java
```diff
@@ -21,9 +21,9 @@
 
 import javax.annotation.CheckForNull;
 
+import jdk.nashorn.internal.objects.annotations.Function;
```
davidjgonzalez: @justinedelson accidental import?

justinedelson: good catch. fixed.

badvision:
Sucks that the "tld file workaround" became the winning solution.  Was hoping we could have a working TLD generator but I guess it's not terribly important with the direction of all things HTL these days.  +1 from me.

Issue #1753, opened 2019-02-13
Disable Dynamic*ClientLibraryServlets and dynamic-loader.js inclusion in 4.x


davidjgonzalez:
The `DynamicTouchUiClientLibraryServlet` and `DynamicClassicUiClientLibraryServlets` should be disabled by default and marked as deprecated in 4.x.

Likewise, the inclusion of the `dynamic-loader.js` in the widget client libraries should be removed.

This is in support for #1715 

/cc @kwin

PR #1754, opened 2019-02-13
#1753 - Disable Dynamic*ClientLibraryServlets and dynamic-loader.js i…

davidjgonzalez:
…nclusion (in support of #1715)

@kwin wdyt?

@justinedelson this looks right in terms of disabling this func, correct? (TBH, ive never used it before)

We can document turning it back on (OSGi config for servlet) and a ClientLibrary for dynamic-loader.js if someone really wants to use it over the wrapper. Though, i cant imagine why..

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/DynamicTouchUiClientLibraryServlet.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/DynamicTouchUiClientLibraryServlet.java
@@ -45,6 +45,9 @@ import java.util.Map;
         )
 })
 @Service(Servlet.class)
+/**
+ * @deprecated Using Wrapper Client Library definitions is the preferred method of using ACS Commons-provided Client Libraries as of ACS Commons 4.0.0.
+ */
 @Deprecated
 public class DynamicTouchUiClientLibraryServlet extends AbstractDynamicClientLibraryServlet {
 
--- a/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/DynamicTouchUiClientLibraryServlet.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/DynamicTouchUiClientLibraryServlet.java
@@ -21,17 +21,30 @@ package com.adobe.acs.commons.wcm.impl;
 
 import com.adobe.granite.ui.clientlibs.HtmlLibraryManager;
 import org.apache.felix.scr.annotations.Activate;
+import org.apache.felix.scr.annotations.Component;
+import org.apache.felix.scr.annotations.ConfigurationPolicy;
+import org.apache.felix.scr.annotations.Properties;
 import org.apache.felix.scr.annotations.Property;
 import org.apache.felix.scr.annotations.Reference;
-import org.apache.felix.scr.annotations.sling.SlingServlet;
+import org.apache.felix.scr.annotations.Service;
 import org.apache.sling.commons.osgi.PropertiesUtil;
 
+import javax.servlet.Servlet;
 import java.util.Map;
 
-@SlingServlet(paths = "/bin/acs-commons/dynamic-touchui-clientlibs.json",
-    label = "ACS AEM Commons - Dynamic Touch UI Client Library Loader",
-    description = "Allows for dynamic loading of optional Touch UI Client Libraries",
-    metatype = true)
+@Component(
+        label = "ACS AEM Commons - Dynamic Touch UI Client Library Loader",
+        description = "(DEPRECATD) Allows for dynamic loading of optional Touch UI Client Libraries",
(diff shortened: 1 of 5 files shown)


content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/js.txt
```diff
@@ -5,4 +5,6 @@ touchui-limit-parsys.js
 touchui-configure-parsys-placeholder.js
 search-based-pathbrowser.js
 draggablelist.js
-dynamic-loader.js
\ No newline at end of file
+
+# Remove automatic inclusion with updated approach using Wrapper Client Library approach (#16715)
```
kwin: typo in the ticket number

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/DynamicTouchUiClientLibraryServlet.java
```diff
@@ -21,17 +21,30 @@
...
+        metatype = true
+)
+@Properties({
+        @Property(
+                name = "sling.servlet.paths",
+                value = "/bin/acs-commons/dynamic-touchui-clientlibs.json"
+        )
+})
+@Service(Servlet.class)
 public class DynamicTouchUiClientLibraryServlet extends AbstractDynamicClientLibraryServlet {
```
kwin: Can you also add a deprecated annotation to that class so that it can be easier removed in upcoming versions?

content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/js.txt
davidjgonzalez: fixed

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/DynamicTouchUiClientLibraryServlet.java
davidjgonzalez: done - thanks for the quick review

content/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/js.txt
```diff
@@ -5,4 +5,6 @@ touchui-limit-parsys.js
 touchui-configure-parsys-placeholder.js
 search-based-pathbrowser.js
 draggablelist.js
-dynamic-loader.js
\ No newline at end of file
+
```
kwin: The whole file needs some rework I think because at least `touchui-limit-parsys.js` and `touchui-configure-parsys-placeholder.js` is no longer available in the `source` folder. I rather think the whole clientlib can be removed as this is incorrectly trying to aggregate stuff. Instead every aspect should be put in its own dedicated clientlib.

content/src/main/content/jcr_root/apps/acs-commons/widgets/js.txt
```diff
@@ -14,4 +14,6 @@ validate_responsive_column_control_dialog.js
 add_dynamic_options_to_feed_importer.js
```
kwin: This clientlib should be removed as it aggregates unrelated features. Rather each feature should be encapsulated in its own clientlib.

davidjgonzalez: @kwin youre saying create individual clientlibs for each logical set of functionality across these files? 

image

Same with w/ the ClassicUI widgets.

I tend to agree. You have any time to define the CL's?

kwin: Unfortunately not in the near future. But IMHO it is more important to fix the one for Touch UI though, as Classic is deprecated anyhow...

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/DynamicTouchUiClientLibraryServlet.java
```diff
@@ -21,17 +21,34 @@
...
+)
+@Properties({
+        @Property(
+                name = "sling.servlet.paths",
+                value = "/bin/acs-commons/dynamic-touchui-clientlibs.json"
+        )
+})
+@Service(Servlet.class)
+/**
+ * @deprecated Using Wrapper Client Library definitions is the preferred method of using ACS Commons-provided Client Libraries as of ACS Commons 4.0.0.
```
kwin: I think you need to add a description otherwise `@deprecated` is not correctly detected by codeclimate.

justinedelson:
@davidjgonzalez i'm a bit confused... if the recommended approach is to use the wrapper libraries and that is already a breaking change, shouldn't we just remove this entirely?

davidjgonzalez:
Sure!

PR #1757, opened 2019-02-18
Feature / OSGI Service Based Granite Render Condition

ahmed-musallam:
# OSGI Service Based Granite Render Condition

A [Granite Render Condition ](https://helpx.adobe.com/experience-manager/6-4/sites/developing/using/reference-materials/granite-ui/api/jcr_root/libs/granite/ui/docs/server/rendercondition.html) that allows consumers to implement Render Conditions as an OSGI service with a dead simple API. No more fiddling with JSPs!

### `granite:rendercondition` node definition

add a `granite:rendercondition` node under any Granite Widget as explained in the [docs](https://helpx.adobe.com/experience-manager/6-4/sites/developing/using/reference-materials/granite-ui/api/jcr_root/libs/granite/ui/docs/server/rendercondition.html. The node definition is as such:

```
granite:rendercondition: 
    - jcr:primaryType: "nt:unstructured"
    - sampleProperty: "some property value"
    - condition.name: "the-implemented-condition-service"
    - default: true
    - sling:resourceType: "/apps/acs-commons/granite/ui/components/renderconditions/service"
```

Where:

- `condition.name` is a mandatory property corresponding to the same property name on the implemented OSGI service (more on that below)
- `default` (optional) is the default render value if the implemented OSGI service fails and is also passed to the OSGI service (more on that below)
-  `sampleProperty` is just an example property. You can obtain this value in the OSGI service impl (more below)

### Implementing a render condition service

Create a new OSGI service that implements: `com.adobe.acs.commons.rendercondition.GraniteRenderCondition` and declare a property `condition.name` whose value is the name you'll use to reference this condition.

> for complete example, see: `com.adobe.acs.commons.rendercondition.Aem64TemplateBasedPagePropertiesCondition`

simple example:

```
@Component(
    immediate = true,
    property = {
        "label=Simple value-basedrender  condition,
        "description=A render condition that is evaluated based a passed boolean value called 'show'",
        GraniteRenderCondition.CONDITION_NAME + "=simple-value-based-condition"
    },
    service = GraniteRenderCondition.class)
public class SampleValueBasedRenderCondition implements GraniteRenderCondition {

    @Override
    public boolean evaluate(
        SlingHttpServletRequest slingHttpServletRequest,
        HttpServletRequest httpServletRequest,
        Config cfg,
        boolean defaultValue) {
             return cfg.get("show", true);
        }
}
```
 
You can then use that condition in a `granite:rendercondition` node such as:

```
granite:rendercondition:
    - jcr:primaryType: "nt:unstructured"
    - show: true
    - condition.name: "simple-value-based-condition"
    - sling:resourceType: "/apps/acs-commons/granite/ui/components/renderconditions/service"
```

## GraniteRenderCondition#evaluate Params:

the `GraniteRenderCondition#evaluate` method passes a few params:

-  `slingHttpServletRequest`: The sling request you know and love.
- `httpServletRequest`: The Servlet Request
   > notably used for the method: httpServletRequest#renderedgetPathInfo, which will obtain the path of the page where the Granite Widget being evaluated is rendered.
- `componentConfig`: The Component Config instance. Helpful in obtaining any properties defined on the `granite:rendercondition` node.
- `defaultValue`: The value of the `default` property on the `granite:rendercondition` node. `false` if left blank.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/rendercondition/Aem64TemplateBasedPagePropertiesCondition.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/rendercondition/Aem64TemplateBasedPagePropertiesCondition.java
@@ -1,0 +1,73 @@
+package com.adobe.acs.commons.rendercondition;
+
+import com.adobe.granite.ui.components.Config;
+import com.day.cq.wcm.api.Page;
+import com.day.cq.wcm.api.Template;
+import javax.servlet.http.HttpServletRequest;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.sling.api.SlingHttpServletRequest;
+import org.apache.sling.api.resource.Resource;
+import org.apache.sling.api.resource.ResourceResolver;
+import org.osgi.service.component.annotations.Component;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Renders page properties Granite Widget based on the current page template.
+ * Assumes that page properties console is under path: /wcm/core/content/sites/properties
+ * and that page path is passed via a request param called "item".
+ * For example, if you open page properties for page: /content/my-company/my-page, the URL would be:
+ * http://localhost:4502/mnt/overlay/wcm/core/content/sites/properties.html?item=/content/my-company/my-page
+ */
+
+@Component(
+    immediate = true,
+    property = {
+        "label=AEM 6.4 Template Based Page Properties Render Condition",
+        "description=Renders Granite Widgets in Page Properties if page template matches tha passed template path"
+    },
+    service = GraniteRenderCondition.class)
+public class Aem64TemplateBasedPagePropertiesCondition implements GraniteRenderCondition {
+
+  private static final Logger log = LoggerFactory.getLogger(Aem64TemplateBasedPagePropertiesCondition.class);
+  private static final String PAGE_PROPERTIES_CONSOLE = "wcm/core/content/sites/properties";
+
+  @Override
+  public boolean evaluate(
+      SlingHttpServletRequest slingHttpServletRequest,
+      HttpServletRequest httpServletRequest,
+      Config cfg,
+      boolean defaultValue) {
+
+    String expectedTemplatePath = cfg.get("templatePath", "");
+
+    if (slingHttpServletRequest == null
+        || httpServletRequest == null
+        || StringUtils.isBlank(expectedTemplatePath)) {
+      throw new IllegalArgumentException("One of the passed parameters is null.");
+    }
+
+    // the open console is the page properties console.
+    if (StringUtils.contains(httpServletRequest.getPathInfo(), PAGE_PROPERTIES_CONSOLE)) {
+
+      // get the actual page path
+      String pagePath = httpServletRequest.getParameter("item");
+
+      ResourceResolver resourceResolver = slingHttpServletRequest.getResourceResolver();
+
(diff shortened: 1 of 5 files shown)


badvision:
This looks really exciting!  Can't wait to see it in action.  Let me know if you have any questions regarding the code climate issues.

ahmed-musallam:
@badvision thank you! I'm excited about it as well!

is there a way I can run code climate checks locally? I don't want to push and wait every time if that's possible.

justinedelson:
@ahmed-musallam yes, you can install the [CodeClimate CLI](https://github.com/codeclimate/codeclimate)

and then run it like

    codeclimate analyze -f text

ahmed-musallam:
Ah, nice! Thank you, will do that later tonight!

badvision:
Are there some easy avenues you can explore to improve the testing coverage?  It would be so awesome to pass the 50% test coverage mark! 😄

justinedelson:
I'm not sure I understand the value of this. You can already write render conditions in Java. See https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/reports/internal/ReportsRenderCondition.java

Perhaps (although that example doesn't really indicate this is the case) doing so could be easier and/or more intuitive, but that could be done with a simple abstract base class or helper method.

ahmed-musallam:
@justinedelson yes you are right, you *could* use a resource-type based servlet. And I would agree that it would make things simpler. **However, that would mean missing two pieces:**

1. the `ComponentHelper` which only has one constructor whose parameter is the JSP `PageContext`. This is needed to obtain properties/datasources on the granite widget resource (JCR Node) 
2. The `HttpServletRequest` object from the JSP. That is, the original servlet request object created from requesting the page on which the granite widget is rendered.

If you know how I can obtain those two objects in a servlet, I can close this PR and write a blog post with your help instead.

justinedelson:
I'm not sure why you need the `ComponentHelper` for that. If you need the properties of the render condition resource, just call `request.getResource().getValueMap()`

Same thing with the request -- why would you need the `HttpServletRequest` instead of the `SlingHttpServletRequest`

I tried rewriting the template rendercondition you created and it seems to work:

```
    @Override
    protected void doGet(@Nonnull SlingHttpServletRequest request, @Nonnull SlingHttpServletResponse response) throws ServletException, IOException {
        ValueMap properties = request.getResource().getValueMap();
        String expectedTemplatePath = properties.get("templatePath", "");

        boolean result = false;

        if (request == null || StringUtils.isBlank(expectedTemplatePath)) {
            throw new IllegalArgumentException("One of the passed parameters is null.");
        }

        // make sure page path exists
        String pagePath = request.getParameter("item");
        if (StringUtils.isBlank(pagePath)) {
            log.warn("item parameter is blank, returning default value, {}", result);
        } else {
            result = Optional.ofNullable(request)
                .map(SlingHttpServletRequest::getResourceResolver)
                .map(resolver -> resolver.getResource(pagePath))
                .map(pageResource -> pageResource.adaptTo(Page.class))
                .map(Page::getTemplate)
                .map(Template::getPath)
                .map(templatePath -> StringUtils.contains(templatePath, expectedTemplatePath))
                .orElse(result);
        }

        request.setAttribute(RenderCondition.class.getName(), new SimpleRenderCondition(result));
    }
```

ahmed-musallam:
@justinedelson you make valid points! And actually, it's much easier and more expressive the way you've described it. Closing this.

PR #1776, opened 2019-03-08
fix possibly negative index

Nov11:
deal with situations that index overflows and returns negative value. this leads to IndexOutOfBoundException


davidjgonzalez:
@Nov11  - Could you add a Unit test for this? as well as a changelog entry?

Nov11:
Unit Tests and an entry in change log have bean added.

PR #1777, opened 2019-03-08
add acs-aem-commons prefix to pipeline.type value to prevent name

### LINKED ISSUE #1773: Name clash for pipeline.type in com.adobe.acs.commons.rewriter.impl.PlainXMLSerializerFactory
The serializer name (`xml-serializer`) being set in https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/e4b480edfada3e75a94d61ee809a4972ddd397c5/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/PlainXMLSerializerFactory.java#L34 clashes with the same name in https://github.com/apache/sling-org-apache-sling-rewriter/blob/669f2074ae550781cd7745422f8b3d17b1045892/src/main/java/org/apache/sling/rewriter/impl/components/XMLSerializerFactory.java#L35.

Also I am wondering in which case I should use the one from acs-aem-commons over the default one.
Update: I guess i figured it out:https://issues.apache.org/jira/browse/SLING-8305

Also I would recommend to use some prefix (like acs-aem-commons) for all properties named `pipeline.type`.

kwin:
clashes

This fixes #1773


kwin:
I only fixed two factories related to xml. But there are some more with potential name clashes. 
Maybe one should prefix all of them with acs-aem-commons. But that change is backwards incompatible, i.e. all consumers would need to fix their pipeline configs.

srikanth010:
Can anyone help me with Column Component?

badvision:
There seems to be a build issue on this PR @kwin.  Let me know if you need any assistance.

kwin:
@badvision The IT failure is fixed now.

PR #1783, opened 2019-03-13
Added the possibility to replace the existing host in an attribute

nick-vanvynckt:
In our current project we have the need to rewrite img src attributes where there is already a host present. In the current setup it does not work and it just prepends the desired host before the matched pattern, resulting in two subsequent hosts. This is not correct so we needed the possibility to replace the existing host in the attribute with the desired host.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StaticReferenceRewriteTransformerFactory.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StaticReferenceRewriteTransformerFactory.java
@@ -267,6 +267,10 @@ public final class StaticReferenceRewriteTransformerFactory implements Transform
         this.staticHostPattern = PropertiesUtil.toStringArray(properties.get(PROP_HOST_NAME_PATTERN), null);
         this.staticHostCount = PropertiesUtil.toInteger(properties.get(PROP_HOST_COUNT), DEFAULT_HOST_COUNT);
         this.replaceHost = PropertiesUtil.toBoolean(properties.get(PROP_REPLACE_HOST), false);
+
+        if (!this.replaceHost && !matchingPatterns.values().stream().anyMatch(str -> str.toString().startsWith("^"))) {
+            log.warn("BEWARE! Replace host is false and your regex is not anchored to the start of the string, this may result in a double host.");
+        }
     }
 
     private static Map<String, Pattern> initializeMatchingPatterns(String[] matchingPatternsProp) {
--- a/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StaticReferenceRewriteTransformerFactory.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StaticReferenceRewriteTransformerFactory.java
@@ -107,7 +107,7 @@ public final class StaticReferenceRewriteTransformerFactory implements Transform
             description = "Path prefixes to rewrite.")
     private static final String PROP_PREFIXES = "prefixes";
 
-    @Property(label = "Override existing host", description = "This property allows you to override the existing host in the attribute that has to be rewritten")
+    @Property(label = "Override existing host", description = "This property allows you to override the existing host in the attribute that has to be rewritten", boolValue = false)
     private static final String PROP_REPLACE_HOST = "replaceHost";
 
     private Map<String, String[]> attributes;
@@ -120,7 +120,7 @@ public final class StaticReferenceRewriteTransformerFactory implements Transform
 
     private String[] staticHostPattern;
 
-    private Boolean replaceHost;
+    private boolean replaceHost;
 
     public Transformer createTransformer() {
         return new StaticReferenceRewriteTransformer();
@@ -231,7 +231,7 @@ public final class StaticReferenceRewriteTransformerFactory implements Transform
                     // prepend host
                     url = prependHostName(url);
                     // Added check to determine whether the existing host has to be replaced
-                    if (BooleanUtils.isTrue(this.replaceHost)) {
+                    if (this.replaceHost) {
                         int index = attrValue.indexOf("://");
                         sb.setLength(0);
                         sb.append(attrValue,0, index + 1);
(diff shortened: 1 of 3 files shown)


bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StaticReferenceRewriteTransformerFactory.java
```diff
@@ -225,13 +231,21 @@ private String handleMatchingPatternAttribute(Pattern pattern, String attrValue)
                     // prepend host
                     url = prependHostName(url);
                     m.appendReplacement(sb, Matcher.quoteReplacement(url));
+                    // Added check to determine whether the existing host has to be replaced
+                    if (BooleanUtils.isTrue(this.replaceHost)) {
+                        int index = attrValue.contains("://") ? attrValue.indexOf("://") + 1 : 0;
```
royteeuwen: what about url's in the form of script src="//assets.adobedtm.com/..." ? This won't be covered by the logic you created

bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StaticReferenceRewriteTransformerFactory.java
```diff
@@ -116,6 +120,8 @@ public void startElement(String namespaceURI, String localName, String qName, At
 
     private String[] staticHostPattern;
 
+    private Boolean replaceHost;
```
justinedelson: Can you just use `boolean` here? It doesn't look like there is a code path where this is null, so the primitive would make more sense.

bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StaticReferenceRewriteTransformerFactory.java
```diff
@@ -224,14 +230,24 @@ private String handleMatchingPatternAttribute(Pattern pattern, String attrValue)
                 if (url.startsWith(prefix)) {
                     // prepend host
                     url = prependHostName(url);
-                    m.appendReplacement(sb, Matcher.quoteReplacement(url));
-                    // First prefix match wins
+                    // Added check to determine whether the existing host has to be replaced
+                    if (BooleanUtils.isTrue(this.replaceHost)) {
```
justinedelson: Same thing -- could just be `if (this.replaceHost)` if `replaceHost` is a primitive.

bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StaticReferenceRewriteTransformerFactory.java
```diff
@@ -224,14 +230,24 @@ private String handleMatchingPatternAttribute(Pattern pattern, String attrValue)
...
+                    } else {
+                        m.appendReplacement(sb, Matcher.quoteReplacement(url));
+                        // First prefix match wins
+                    }
                     break;
                 }
             }
+        }
 
+        if (BooleanUtils.isFalse(this.replaceHost)) {
```
justinedelson: ditto

bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/StaticReferenceRewriteTransformerFactory.java
```diff
@@ -106,6 +107,9 @@ public void startElement(String namespaceURI, String localName, String qName, At
             description = "Path prefixes to rewrite.")
     private static final String PROP_PREFIXES = "prefixes";
 
+    @Property(label = "Override existing host", description = "This property allows you to override the existing host in the attribute that has to be rewritten")
```
justinedelson: Please add a default (of false) here so this gets recognized as a `boolean` property.

bundle/src/test/java/com/adobe/acs/commons/rewriter/impl/StaticReferenceRewriteTransformerFactoryTest.java
```diff
@@ -121,6 +121,58 @@ public void test_with_prefix_and_matching_pattern_and_single_host() throws Excep
...
 
+    @Test
+    public void test_with_prefix_and_matching_pattern_and_single_host_and_replace_host() throws Exception {
+        MockBundle bundle = new MockBundle(-1);
+        MockComponentContext ctx = new MockComponentContext(bundle);
+        ctx.setProperty("prefixes", new String[] { "/content/dam" });
+        ctx.setProperty("attributes", new String[] { "img:src" });
+        ctx.setProperty("host.pattern", "static.host.com");
+        ctx.setProperty("matchingPatterns", "img:src;(\\/content\\/dam\\/.+?\\.(png|jpg))");
+        ctx.setProperty("replaceHost", "true");
```
justinedelson: I don't think it really matters since `PropertiesUtil` handles this case, but this should probably just be a boolean `true` not a string.

bundle/src/test/java/com/adobe/acs/commons/rewriter/impl/StaticReferenceRewriteTransformerFactoryTest.java
```diff
@@ -121,6 +121,58 @@ public void test_with_prefix_and_matching_pattern_and_single_host() throws Excep
...
+
+    @Test
+    public void test_with_prefix_and_matching_pattern_and_single_host_and_replace_host_without_protocol() throws Exception {
+        MockBundle bundle = new MockBundle(-1);
+        MockComponentContext ctx = new MockComponentContext(bundle);
+        ctx.setProperty("prefixes", new String[] { "/content/dam" });
+        ctx.setProperty("attributes", new String[] { "img:src" });
+        ctx.setProperty("host.pattern", "static.host.com");
+        ctx.setProperty("matchingPatterns", "img:src;(\\/content\\/dam\\/.+?\\.(png|jpg))");
+        ctx.setProperty("replaceHost", "true");
```
justinedelson: ditto

nick-vanvynckt: @justinedelson, thanks for your comments. I updated the PR with the new code. Any idea when this can be merged?

PR #1795, opened 2019-03-19
Feature/asset packager

id-keenan:
I was recently tasked with moving a fair bit of content (and the associated assets) from Server A to Server B and was unhappy with the bulk packages necessary to do so. I came up with a tool to iterate over properties and generate a package filter I could then use to create a content package and move the content without needing to resort to larger packages. I figured it could be of use to other people so the ACS AEM Commons would be a great place for it to live - especially considering the other content packagers already present in it. Note: also really handy for porting subsections of content and assets to a local instance when trying to resolve issues present on a remote server. It saves time and space on your machine to just bring what is needed!

**Example Task:** 
Move /content/we-retail/en/subsection and the associated assets to a new server.
**Current way to do so:**
Package up /content/we-retail/en/subsection and /content/dam/we-retail (since you can't know what was used you usually resort to bringing more than necessary).
**New way:**
Create a content packager definition providing the page path "/content/we-retail/en/subsection", define any restrictions wanted on either the pages or assets packaged, and create the package!

Note: This is my first ACS AEM Commons contribution so I'm not positive if I did everything correctly - happy to amend anything necessary!


HitmanInWis:
@davidjgonzalez @badvision are you okay with me reviewing this?  Or should I avoid reviewing/merging PRs from my own colleagues?

davidjgonzalez:
@hitmaninwis as long as you weren’t super involved in writing it (ie dev bias) - then review away!  We trust you will review all code equally, regardless of the author! :)

badvision:
I think that @HitmanInWis has proven his commitment to quality and substance time and time again.  No complaints from me.

content/src/main/content/jcr_root/apps/acs-commons/components/utilities/packager/asset-packager/configuration/dialog.xml
```diff
@@ -0,0 +1,140 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
HitmanInWis: I know all of the other packagers are done in Classic UI dialogs - do we think it would be a lot of effort for this one to be the first one using a Touch UI dialog?  Eventually all of the other packagers are going to need to convert.

content/src/main/content/jcr_root/apps/acs-commons/components/utilities/packager/asset-packager/configuration/dialog.xml
```diff
@@ -0,0 +1,140 @@
...
+            <asset-prefix
+                jcr:primaryType="cq:Widget"
+                fieldDescription="The path that the asset reference has to start with."
+                fieldLabel="Asset Prefix"
+                name="./assetPrefix"
+                rootPath="/content/dam"
+                xtype="pathfield"/>
+            <page-exclusions
+                jcr:primaryType="cq:Widget"
+                fieldDescription="Click the '+' to add a new exclude pattern. The pattern will both exclude the pages from the package and exclude them from the asset search.&lt;br/>By default all pages are searched and included in the package."
```
HitmanInWis: Might be more clear to say "... pattern will exclude both the pages and their assets from the package.<br/>By default..."

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImpl.java
```diff
@@ -0,0 +1,281 @@
...
+
+        try {
+            final String pagePath = properties.get(PN_PAGE_PATH, String.class);
+            final List<PathFilterSet> paths;
+            if (StringUtils.isBlank(pagePath)) {
+                paths = Collections.emptyList();
+            } else {
+                paths = this.findAssetPaths(resourceResolver, pagePath);
+            }
+            if (!properties.get(PN_EXCLUDE_PAGES, false) && paths.size() > 0) {
```
HitmanInWis: For code consistency and readability, would it make sense to put lines 111-119 into a `findPagePaths` function similar to `findAssetPaths`?  seems odd that we delegate for one and not the other.

content/src/main/content/jcr_root/apps/acs-commons/components/utilities/packager/asset-packager/configuration/dialog.xml
id-keenan: I thought about this and my initial desire was to do it in Touch UI, but I actually opted for Classic UI because I think if this packager moves to Touch UI then all the others will have to as well for consistency and I honestly didn't have the time to rework the whole section - it's all structured around the supertype of "acs-commons/components/utilities/packager".
I could raise an issue to have all these converted to Touch UI so the desire is tracked at least if that works.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 - #1778 - Added folder support to system notifications
 - #1780 - Added a new version of the XSS Taglib to support the sling XSSAPI.
 - #1783 - Added the possibility to replace the existing host in an attribute
+- #1795 - Added the Asset Content Packager and unit tests
```
HitmanInWis: Comment should probably omit "and unit tests" - no other features mention that and it's more just a prerequisite of adding the feature in the first place.

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImpl.java
```diff
@@ -0,0 +1,281 @@
...
+    private static final String DEFAULT_PACKAGE_GROUP_NAME = "Assets";
+    private static final String DEFAULT_PACKAGE_DESCRIPTION = "Asset Package initially defined by a ACS AEM Commons - "
+            + "Asset Packager configuration.";
+    private static final String THUMBNAIL_RESOURCE_PATH =
+            "/apps/acs-commons/components/utilities/packager/asset-packager/definition/package-thumbnail.png";
+
+    @Reference
+    private PackageHelper packageHelper;
+
+    private String customPrefix;
```
HitmanInWis: Servlets are not threadsafe, and should therefore not use instance variables.  Though it's a low risk on a servlet like this that have very little chance of being executed simultaneously by two different users, out of principle we should update this to avoid the bad practice.

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImpl.java
```diff
@@ -0,0 +1,281 @@
...
+    private List<Pattern> pageExclusionPatterns;
+    private List<Pattern> assetExclusionPatterns;
+    private final List<String> excludedPages = new ArrayList<>();
```
HitmanInWis: With this being `final` and thus only every added to in the code, I actually wonder if this consistently accrues pages across multiple requests.  This issue goes away if we remove this from being an instance var, as mentioned above.

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImpl.java
```diff
@@ -0,0 +1,281 @@
...
+     */
+    private List<PathFilterSet> findAssetPaths(final ResourceResolver resourceResolver,
+                                               final String pagePath) {
+
+        final List<PathFilterSet> filters = new ArrayList<>();
+        final Resource parentResource = resourceResolver.resolve(pagePath);
+        if (isExcluded(this.pageExclusionPatterns, pagePath)) {
+            final PageManager pageManager = resourceResolver.adaptTo(PageManager.class);
+            Page page = pageManager.getContainingPage(parentResource);
+            if (!excludedPages.contains(page.getPath())) {
```
HitmanInWis: Might simplify all of the code to use `Set` (even `LinkedHashSet` to keep order if you wish) in order to more easily avoid duplicates in all of the different collections we're playing with in this code.

HitmanInWis: For example, if `excludedPages` is a Set, these 3 lines can be reduced to 1

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImpl.java
```diff
@@ -0,0 +1,281 @@
...
+     *
+     * @param request The request obj
+     * @return A ValueMap representing the properties
+     */
+    private ValueMap getProperties(final SlingHttpServletRequest request) {
+        if (request.getResource().getChild("configuration") == null) {
+            log.warn("Asset Packager Configuration node could not be found for: {}", request.getResource());
+            return new ValueMapDecorator(new HashMap<>());
+        } else {
+            return request.getResource().getChild("configuration").adaptTo(ValueMap.class);
```
HitmanInWis: `.adaptTo(ValueMap.class` can instead be `getValueMap()` (optional refactor)

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImpl.java
```diff
@@ -0,0 +1,281 @@
...
+     * @param pagePath         The page path
+     * @return The full list of filter sets
+     */
+    private List<PathFilterSet> findAssetPaths(final ResourceResolver resourceResolver,
+                                               final String pagePath) {
+
+        final List<PathFilterSet> filters = new ArrayList<>();
+        final Resource parentResource = resourceResolver.resolve(pagePath);
+        if (isExcluded(this.pageExclusionPatterns, pagePath)) {
+            final PageManager pageManager = resourceResolver.adaptTo(PageManager.class);
```
HitmanInWis: Is this logic to get the containing page path instead of the current `pagePath` required for some sort of edge case?  Struggling to think of the case where it's needed.

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImpl.java
```diff
@@ -0,0 +1,281 @@
...
+
+        final ValueMap properties = this.getProperties(request);
+
+        this.customPrefix = properties.get(PN_ASSET_PREFIX, String.class);
+        this.pageExclusionPatterns = generatePatterns(properties.get(PN_PAGE_EXCLUSIONS, new String[]{}));
+        this.assetExclusionPatterns = generatePatterns(properties.get(PN_ASSET_EXCLUSIONS, new String[]{}));
+
+        try {
+            final String pagePath = properties.get(PN_PAGE_PATH, String.class);
+            final List<PathFilterSet> paths;
```
HitmanInWis: Can we rename this to `pathFilters` to be more clear?

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImpl.java
HitmanInWis: After reading the code I have a better understanding of what's happening here...this is more `addPageExcludes`

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImpl.java
HitmanInWis: or even `packagePathFilters`

content/src/main/content/jcr_root/apps/acs-commons/components/utilities/packager/asset-packager/configuration/dialog.xml
HitmanInWis: Ok, I think that's fair.

HitmanInWis:
Also be sure @id-keenan to submit a documentation update PR to https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io to go along with this new feature.  LMK if you need a hand getting set up locally.

bundle/src/test/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImplTest.java
```diff
@@ -0,0 +1,207 @@
...
+@RunWith(MockitoJUnitRunner.class)
+public class AssetPackagerServletImplTest {
+
+    @Rule
+    public final AemContext context = new AemContext(ResourceResolverType.JCR_OAK);
+
+    @Mock
+    Packaging packaging;
+
+    @InjectMocks
```
HitmanInWis: I'm not really familiar with `@InjectMocks` but it appears that it is not needed on this var.

bundle/src/test/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImplTest.java
```diff
@@ -0,0 +1,207 @@
...
+        assertEquals(responseJSON.getString(STATUS), PREVIEW);
+        JSONArray filterSets = responseJSON.getJSONArray(FILTER_OBJECT);
+        assertEquals(filterSets.length(), previewResults.length);
+        for (int i = 0; i < previewResults.length; i++) {
+            assertEquals(filterSets.getJSONObject(i).getString(ROOT_PATH), previewResults[i]);
+        }
+    }
+
+    @Test
+    public void testAssetExclusion() throws Exception {
```
HitmanInWis: Would it make sense for these tests to validate that multiple exclusion paths are honored?

bundle/src/test/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImplTest.java
```diff
@@ -0,0 +1,207 @@
...
+    public final AemContext context = new AemContext(ResourceResolverType.JCR_OAK);
+
+    @Mock
+    Packaging packaging;
+
+    @InjectMocks
+    @Spy
+    final AssetPackagerServletImpl assetPackagerServlet = new AssetPackagerServletImpl();
+
+    @InjectMocks
```
HitmanInWis: Im a bit confused by this `@InjectMocks` mostly b/c I've never used it, but docs at https://static.javadoc.io/org.mockito/mockito-core/2.27.0/org/mockito/InjectMocks.html seem to indicate it would be used to create a mock itself rather than act on an instance you've created.  That said, if I remove it the tests break, so I'm not entirely sure :D.

id-keenan: So I have the `@InjectMocks` here because from what I understand it replaces a value _inside_ the object with a mocked object above. So what I'm doing is mocking the `packaging` object above and then the injection replaces the local 'packaging' object (which is null in my mocked objects but necessary to create a package) with this mocked object that has (line 134) a defined return for a package manager that lets me actually fake the package objects.

content/src/main/content/jcr_root/apps/acs-commons/components/utilities/packager/asset-packager/configuration/dialog.xml
id-keenan: Addressed!

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImpl.java
id-keenan: It's more of a check to see if the user wants the page path included in the content package, and if they do then we add the page + any exclusions they added. I'll separate it out into another method

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImpl.java
davidjgonzalez: +1 - generally speaking, `.getValueMap()` is preferred as it guarantees a non-null (Empty value map) return value, whereas with `.adaptTo(ValueMap.class)` requires a null check before accessing.

bundle/src/test/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImplTest.java
HitmanInWis: ahhh ok, thanks for connecting the dots for me :D

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImpl.java
id-keenan: Addressed this!

CHANGELOG.md
id-keenan: Updated Changelog

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImpl.java
id-keenan: I did this to avoid passing each of these 4 objects around in every method call but I see the point in changing that. Not sure I have a ton of options other than just passing them around either - it feels like I could do that better. Any suggestions?

HitmanInWis: Probably best suggestion would be to modularize to a separate class that is instantiated on each request and executes the logic.

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImpl.java
id-keenan: So the `findAssetPaths` method recurses over all the resources underneath a given path, so the `parentResource` defined here can be something like `/content/we-retail/language-masters/en/experience/arctic-surfing-in-lofoten/jcr:content/root/hero_image`. The point of the 'page exclusions' is to exclude content from certain pages, so it's getting the page of each resource that's being iterated over and checking if it's excluded. Essentially the point is to compare the page exclusions against the page currently being looked at (from any level).

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImpl.java
id-keenan: Renamed!

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImpl.java
id-keenan: Addressed!

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImpl.java
HitmanInWis: Correct me if I'm wrong, but if `/content/we-retail/language-masters/en/experience/arctic-surfing-in-lofoten` is already in the exclusion paths then we will never recurse down to `/content/we-retail/language-masters/en/experience/arctic-surfing-in-lofoten/jcr:content/root/hero_image` in the first place, will we?

HitmanInWis:
With the cut of v4.1.0 today, please be sure to merge latest master and update the CHANGELOG.md file accordingly to move this feature to the next release.

bundle/src/test/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImplTest.java
id-keenan: I believe it is since it will inject the above mocked object

bundle/src/test/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImplTest.java
id-keenan: I've updated the tests to include 2 tests for multiple exclusions. Something I would like to test but can't figure out a way to do is to test if the filter in the package _actually has_ the exclusion. I can assume that it does because the assets on that page are not present in the JSON response, but I can't verify the actual filters in it. See screens:
JSON response from servlet:
Screen Shot 2019-05-13 at 4 57 21 PM
Actual filters of the package:
Screen Shot 2019-05-13 at 4 57 37 PM

I don't feel _that_ bad about testing this though because technically it should be part of the PackageHelperImplTest class :)

id-keenan:
@HitmanInWis The PR updates have been made, feel free to take a second pass!

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImpl.java
badvision: Don't feel too bad.  I've been saying the same thing for 20 years.  It's a journey.  😄

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ after the 3.9.0 release. All changes up until the 3.9.0 release can be found in
 The format is based on [Keep a Changelog](http://keepachangelog.com)
 
 ## [Unreleased]
+- #1795 - Added the Asset Content Packager
```
HitmanInWis: Should be a `### Added` category here.

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/AssetPackagerServletImpl.java
HitmanInWis: ok, just seemed like we were running extra unnecessary code here, but I dont see any real harm in the way it is written

CHANGELOG.md
id-keenan: Added the added. Sorry :)

id-keenan:
@HitmanInWis updated to remove duplicates. I added a Collections.reverse call on the translated Set object to keep the page path entry at the top - I feel like that's more author friendly than having it at the bottom of a potentially giant list. Probably unimportant but I liked it at the top as opposed to the bottom :)

bundle/src/main/java/com/adobe/acs/commons/packaging/util/AssetPackageUtil.java
```diff
@@ -0,0 +1,225 @@
...
+     */
+    private Set<PathFilterSet> findAssetPaths(final ResourceResolver resourceResolver,
+                                               final String pagePath) {
+
+        final Set<PathFilterSet> filters = new LinkedHashSet<>();
+        final Resource parentResource = resourceResolver.resolve(pagePath);
+        if (isExcluded(this.pageExclusionPatterns, pagePath)) {
+            final PageManager pageManager = resourceResolver.adaptTo(PageManager.class);
+            Page page = pageManager.getContainingPage(parentResource);
+            excludedPages.add(page.getPath());
```
HitmanInWis: One last item in testing.

The **page** exclude patterns are AEM **paths**, but the **asset** exclude patterns are **regexp** expressions.  This is confusing from a usability perspective because the fields appear to be 100% the same (just one functions on pages, the other on assets).  I had to figure this out via trial and error when testing.

Perhaps we can solve with additional instructions and differentiated naming.  Or maybe you have a better idea.

Screen Shot 2019-05-14 at 3 40 16 PM

id-keenan: The page exclusion patterns accept regex as well. They are just represented in the filters as the matched page paths. Are you seeing the page exclusion regex not working? It's all run through `isExcluded` in `AssetPackageUtil` - line 217

id-keenan: Addressed in the latest commit

bundle/src/main/java/com/adobe/acs/commons/packaging/util/AssetPackageUtil.java
```diff
@@ -0,0 +1,242 @@
...
+
+        final Set<PathFilterSet> filters = new LinkedHashSet<>();
+        final Resource parentResource = resourceResolver.resolve(pagePath);
+        if (isExcluded(this.pageExclusionPatterns, pagePath)) {
+            if (pagePath.endsWith("/" + NameConstants.NN_CONTENT)) {
+                return filters;
+            }
+            final PageManager pageManager = resourceResolver.adaptTo(PageManager.class);
+            Page page = pageManager.getContainingPage(parentResource);
+            excludedPages.add(page.getPath());
```
HitmanInWis: A bit of an edge case, but when I test a page exclude path of `.*men.*` I get a NullPointerException on this line that blows up the package creation.

The reason is it finds `/content/we-retail/us/en/community/configuration/functions/assignments` where there is no "containing page".  The same error happens with full paths of `/content/we-retail/us/en/community/configuration` or `/content/we-retail/us/en/community/configuration/functions`.

As such:
- There is currently no way to exclude a non-page path - folders are coming along for the ride, whether desired or not
- If an exclusion pattern matches a folder path, the package creation errors out w/o any notification to the user.

id-keenan: Addressed in the latest commit.

PR #1797, opened 2019-03-20
Adding OakPal check to validate that all imported packages are export…

justinedelson:
…ed by AEM 6.3 and 6.4

Basically the way this works is that you can build a JSON representation of all of the exported packages in a "stock" AEM instance (ideally run in the nosamplecontent runmode). Included here are extracts from 6.3 and 6.4, but conceptually this could support other versions.

Then at check time, any bundles are scanned -- any exported packages are added to the stock list. And then all of the imported packages are compared to see if they are satisfied by the exported list.

This is really an idea more than a full implementation. Specifically, I'm not sure the best way to test this. Also I hate the parsing code.

But I'm creating this PR so that @adamcin (and anyone interested) can give their feedback.


davidjgonzalez:
@justinedelson for posterity, are there (will there be) instructions on how to generate the JSON extracts for future versions of AEM?

justinedelson:
@davidjgonzalez good point. it should probably go in the README or something like that. FWIW, the command is

```
cd oakpal-checks
mvn -Pextract-bundle-info -DaemVersion=<aemVersion> -Dcrx.port=<port number, defaults to 4502>
```

adamcin:
@justinedelson This is cool. I think it makes sense to keep this in the internal checklist with limited configurability. 

I noticed you triggered the violation reporting in `finishedScan` rather than in `afterExtract`, which I assume you did to avoid multiple expensive loops over the list of exports, so I think that is a reasonable tradeoff for having to store an additional map of packageIds. 

For manifest parsing in oakpal-core, I may have gone overkill by embedding a few classes from bnd to avoid having to deal with the quoting rules and possible value parameters, but it did clean up the actual parsing code quite a bit: https://github.com/adamcin/oakpal/blob/master/core/src/main/java/net/adamcin/oakpal/core/Util.java#L85

justinedelson:
> I noticed you triggered the violation reporting in finishedScan rather than in afterExtract, which I assume you did to avoid multiple expensive loops over the list of exports, so I think that is a reasonable tradeoff for having to store an additional map of packageIds.

The reason for this is that in theory (although not the case today with ACS AEM Commons), the set of tested content packages could contain multiple bundles and you need to collect all of the exported packages first (in `importedPath`) and only once that is done evaluate the set of imported packages.

I suppose in theory it would make sense to do the import evaluation multiple times -- if the imports aren't satisfied when the bundle is initially installed, but then gets satisfied later, maybe that's a minor violation since it will get satisfied eventually, but only after another Sling installer cycle. But mostly I was just trying to get something working :)

Good point about possibly embedding bnd.

badvision:
👍 from me but I'll leave it up to you folks to decide when to merge this one in. 😉

Issue #1778, opened 2019-03-10
[Feature Improvement] System notifications to groups


josephrignanese:
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.3.3.0
* [x] ACS AEM Commons Version: 3.17.2
* [x] Reproducible on Latest? yes

### Expected Behavior

The system notifications feature should allow for messages to be sent to particular users.

### Actual Behavior

Currently there is no easily maintainable way to allow for system notification messages to be targeted to specific users or groups

### Steps to Reproduce

1. Create a system notification. There is no configuration to allow this to be limited to groups.
2. Attempt it will display for all users

### Proposed approach

Modify the system notifications code to walk the folder structure under `/etc/acs-commons/notifications`.

This will allow for ACLs to be set up on subfolders which can be used to control which users/groups get the notifications.

### Links

https://adobe-consulting-services.github.io/acs-aem-commons/features/system-notifications/index.html

josephrignanese:
@badvision this feature was actually merged in and is part of release 4.1.0, can you update the issue appropriately?

badvision:
Thanks for letting me know @josephrignanese !

PR #1804, opened 2019-03-25
#1778 Enhanced system notifications to walk tree

josephrignanese:
PR for #1778


badvision:
Good idea!  Also, if you haven't seen there are some visitor implementations that can help simplify things too: https://github.com/Adobe-Consulting-Services/acs-aem-commons/tree/master/bundle/src/main/java/com/adobe/acs/commons/util/visitors

The SimpleFilteringResourceVisitor and TreeFilteringResourceVisitor allow you to provide lambda expressions.  Not that you need them here necessarily but just in case you were wondering if there were other ways.

PR #1805, opened 2019-03-25
The unit tests of Variant could fail with unusual OS locale settings.

csaboka:
My OS is set up to use the en_US locale on the UI but the hu_HU locale for date and time formatting. In this setup, a unit test of Variant failed because it tried to parse Hungarian dates with a US English format. This change should make the test work regardless of date format settings.


badvision:
🙇 Good find!  Thanks for fixing this one! 👍

PR #1809, opened 2019-03-29
Fixes #1807

### LINKED ISSUE #1807: Unable to deploy, get bundle dependency errors
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.4.3
* [x] ACS AEM Commons Version: 4.0.1-SNAPSHOT
* [X] Reproducible on Latest? yes

### Expected Behavior

Deploying the bundle should work

### Actual Behavior

Deploying the bundle does not work because dependencies are not met, mainly com.google.common and io.jsonwebtoken

Not sure but could this be related to recent changes @niekraaijmakers ?

badvision:


badvision:
@justinedelson not sure why I had to include google.commons all of a sudden to get the bundle to start in 6.4 (sp3).  Thoughts?

niekraaijmakers:
This is weird indeed. We use this for ages.

com.adobe.acs.commons.* as private package? Won't we block off all references then?

justinedelson:
My guess is that this is a regression introduced by the shade plugin. Would probably be better to revert that instead of introducing a workaround on top of it. I can look more on Monday.

justinedelson:
I think #1815 is a more idiomatic solution to this.

PR #1812, opened 2019-03-30
adding scenario of uncheckedvalue property present in checkbox value

Aashay-Chapatwala:
adding scenario of uncheckedvalue property present in checkbox value


badvision:
It's worth noting that the multifield in ACS Commons is effectively deprecated in favor of the product-provided version now.

Aashay-Chapatwala:
> It's worth noting that the multifield in ACS Commons is effectively deprecated in favor of the product-provided version now.

So you are suggesting that ACS commons 4. 0.0 & onward will not be supporting multifield feature & it's functionalities?

PR #1820, opened 2019-04-07
Added service rankings for http cache extensions, so that the combine…

niekraaijmakers:
Added service rankings for http cache extensions, so that the combiner extension through ranked services, uses the cheap checks first and expensive checks later to save performance.


badvision:
@niekraaijmakers looks like a small merge conflict and I'm not sure how to proceed with resolving it.  Care to take a look real fast?

niekraaijmakers:
done !

PR #1827, opened 2019-04-09
Added locale object, injected from resource page, with sling request …

niekraaijmakers:
Slight extension that adds Locale to the AemObject injector, injecting Locale based on resource Page, with slingrequest as fallback.


badvision:
Can you add a unit test method for the resolveLocale logic @niekraaijmakers ?   Thanks so much! :D

niekraaijmakers:
too difficult for this one, because aem context / sling context don't support getContainingPage on the resourceResolver :( not sure how to test this?

badvision:
You can just use mockito to make a fake object that returns something... really you just want to trigger expected behaviors in the method to ensure all the conditional logic works like it should.

niekraaijmakers:
ok sir :) because you asked

PR #1850, opened 2019-04-23
Prevent ISE when calling getBufferedString

### LINKED ISSUE #1848: AEM Environment Indicator breaks 6.5 interface
* [ ] AEM Version: `6.5`
* [ ] ACS AEM Commons Version: `4.0.1-SNAPSHOT`

### Expected Behavior

[AEM Environment Indicator](https://adobe-consulting-services.github.io/acs-aem-commons/features/environment-indicator/index..html) does not break anything when configured with any data.

### Actual Behavior

[How to Use](https://adobe-consulting-services.github.io/acs-aem-commons/features/environment-indicator/index..html#how-to-use) example breaks the instance interface:

crxdelite

start

### Steps to Reproduce

Implement the indicator just as described in [How to Use](https://adobe-consulting-services.github.io/acs-aem-commons/features/environment-indicator/index..html#how-to-use).

kwin:
Add guard to check if we really act on a writer-type response.

This fixes #1848

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/AemEnvironmentIndicatorFilter.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/AemEnvironmentIndicatorFilter.java
@@ -53,6 +53,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.adobe.acs.commons.util.BufferedHttpServletResponse;
+import com.adobe.acs.commons.util.BufferedServletOutput.ResponseWriteMethod;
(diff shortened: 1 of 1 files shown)


kwin:
@badvision As this regression was never released I don't think we need to extend the changelog, right?

badvision:
Yeah that's fine.  Thanks @kwin ! 👍

davidjgonzalez:
@kwin are there any other effected features that need similar treatment? Iirc the buffered response was used elsewhere.

kwin:
@davidjgonzalez Indeed there are some other places which need guards, I opened #1853 to treat those.

PR #1852, opened 2019-04-24
Remove event based listeners

csaboka:
Observing resource changes via the OSGi event mechanism is deprecated in favor of using the ResourceChangeListener API. This PR aims to get rid of all deprecated observation code, which should ideally allow Sling to stop generating the legacy OSGi events entirely and improve performance.

There is one case where the internals of the event mechanism are exposed in a component configuration so the switch cannot be cleanly made without breaking backwards compatibility. In this case, I've added a new configuration setting that enables the clean use of the ResourceChangeListener API, and fall back to registering an event handler if only an event filter is provided.


csaboka:
The build is failing on an unrelated unit test. The test also seems to be flaky since it only failed in one of the five builds. I think this is still OK to merge like this, but please let me know if I need to make further changes.

PR #1863, opened 2019-04-28
Feature/codeclimate fixes equals error

niekraaijmakers:


bundle/src/main/java/com/adobe/acs/commons/httpcache/config/impl/RequestHeaderHttpCacheConfigExtension.java
```diff
@@ -37,6 +37,9 @@
 import java.util.Map;
 import java.util.UUID;
 
+import static org.apache.commons.lang3.ArrayUtils.contains;
```
davidjgonzalez: static import of methods is unclear and makes it appear these methods are implemented by this class. better to clearly qualify where they come from in code.

bundle/src/main/java/com/adobe/acs/commons/httpcache/config/impl/RequestParameterHttpCacheConfigExtension.java
```diff
@@ -39,6 +39,9 @@
 import java.util.Map;
 import java.util.UUID;
 
+import static org.apache.commons.collections.CollectionUtils.containsAny;
```
davidjgonzalez: static import of methods is unclear and makes it appear these methods are implemented by this class. better to clearly qualify where they come from in code.

bundle/src/main/java/com/adobe/acs/commons/httpcache/config/impl/ResourcePropertiesHttpCacheConfigExtension.java
```diff
@@ -35,7 +35,14 @@
 import org.osgi.service.metatype.annotations.Designate;
 import org.osgi.service.metatype.annotations.ObjectClassDefinition;
 
-import java.util.*;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+import static org.apache.commons.collections.CollectionUtils.containsAny;
```
davidjgonzalez: static import of methods is unclear and makes it appear these methods are implemented by this class. better to clearly qualify where they come from in code.

niekraaijmakers:
Good question regarding the map to hashmap. Because map isnt serializeable it generates a codeclimate warning. I could also suppress the warning alternatively?

justinedelson:
> Good question regarding the map to hashmap. Because map isnt serializeable it generates a codeclimate warning. I could also suppress the warning alternatively?

If I understand the warning emitted by CodeClimate, it has more to do with the serialization.

My suggestion would be to leave `AbstractKeyValueExtension` as is (with `Map`) and then create a new HashMap when serializing.

niekraaijmakers:
I guess that can work as well, so wrap the map with a hashmap on serialization. I'll do that one. Give me some time as I'm quite busy with 2 projects :)

niekraaijmakers:
thanks !! @davidjgonzalez  do you approve sir?

Issue #1856, opened 2019-04-24
SpreadsheetTest fails on non-English OS settings


csaboka:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: (irrelevant)
* [ ] ACS AEM Commons Version: 4.0.1-SNAPSHOT (latest)
* [ ] Reproducible on Latest? yes

### Expected Behavior

When running unit tests using e.g. "mvn clean package", all unit tests complete.

### Actual Behavior

If your OS locale is Hungarian, the SpreadsheetTest unit test fails with the following error:
```[ERROR] Failures:
[ERROR]   SpreadsheetTest.testSheetTypesAsStrings:176 expected:<123[.]456> but was:<123[,]456>
```

I'm sure the issue will also happen on other non-English locales that use something other than the dot for decimal separator.

### Steps to Reproduce

1. Set your OS locale to something that uses decimal comma (German or French can work). How to do this will depend on your OS.
2. Execute `mvn test` in the root source folder.

The issue is coming from the fact that the Variant class implicitly depends on the default Java locale in multiple places. This behavior makes CompositeVariant and Spreadsheet also depend on the default Java locale, so the assertions in the unit test can either succeed or fail based on an external factor.

How to fix this will depend on whether the locale dependency of Variant is considered a feature or a bug. If it's a feature, the unit test should either make its string-based assertions with dynamically-formatted expected values, or not make those assertions at all. If it's a bug, then Variant should get a way to specify its locale, and this change should be also applied to CompositeVariant and Spreadsheet; then the unit test could specify that it wants an en_US spreadsheet and freely make its US-based assertions.

csaboka:
I'd like to help with fixing this issue, but I'm really not sure which of the two options is better (i.e. embracing dependency on the default locale vs. removing it).

badvision:
Actually it might make sense to enforce the locale detected in the spreadsheet if possible.  Another point to consider is that we're relying on a library and need to at least isolate if it's a bug in the library or in our code.

csaboka:
> Actually it might make sense to enforce the locale detected in the spreadsheet if possible.

I don't think Excel files have any locale saved with them. If I open the xlsx file from the test folder in Excel, I'm just seeing the numbers with my locale's proper decimal separator without any indication that the spreadsheet was created with another separator. I can't see any locale either on the Excel UI, or the public API of POI (the library used to parse the xlsx file).

[...]

PR #1864, opened 2019-04-28
Proposed fix for #1856: Allow specifying Spreadsheet locale.

csaboka:
Now the unit test can explicitly set US as the locale and assert safely about numbers and dates.


bundle/src/main/java/com/adobe/acs/commons/data/Variant.java
```diff
@@ -64,9 +66,13 @@ public Variant() {
     public <T> Variant(T src) {
         setValue(src);
     }
-
+    
     public Variant(Cell src) {
-        setValue(src);
+        this(src, Locale.getDefault());
```
kwin: Wouldn't it make more sense to take the Locale.ROOT as a language/country agnostic default?

csaboka: It could be, but it would change current behavior. This is a public constructor, so clients may already be using it and relying on the current default of Locale.getDecault().

davidjgonzalez:
@kwin i lost track of the review of this one - did it look good to you?

bundle/src/main/java/com/adobe/acs/commons/data/Spreadsheet.java
```diff
@@ -124,6 +126,25 @@ public Spreadsheet(RequestParameter file, List<String> caseInsensitiveHeaders, S
         this(true, file, required);
         this.caseInsensitiveHeaders = caseInsensitiveHeaders;
     }
+    
+    /**
+     * Sets the locale to be used for numeric and date/time conversions.
+     * @param locale The new locale to use.
+     * @return This Spreadsheet instance for method chaining.
+     */
+    public Spreadsheet setLocale(Locale locale) {
```
kwin: Instead of making this an explicit setter wouldn't it be better to just overload the method `buildSpreadsheet` with another version which takes the Locale? I don't think you actually need to store the Locale in a field and expose a getter for that. WDYT?

csaboka: Passing the locale through two layers of helper methods felt like a pain, but you're correct that the public API is cleaner if this is passed as a parameter instead of being set in a field.

Issue #1834, opened 2019-04-15
Audit Log Search does not work?


reggie7:
* [ ] AEM Version: `6.3`, `6.5`
* [ ] ACS AEM Commons Version: `4.0.0`
* [ ] Reproducible on Latest? yes

### Expected Behavior

*Audit Log Search* works OotB. I would expect to have it work as per the [docs](https://adobe-consulting-services.github.io/acs-aem-commons/features/audit-log-search/index.html) simply. Even better - just by interacting with the console and having it display all the warnings/errors in response to my actions.

### Actual Behavior

*Audit Log Search* does not work OotB at all. Or am I doing something wrong?
I have not created any new indices since the warning says:

> No index found for the type cq:AuditEvent, this will result in very slow performance.

I could live with slower performance for now though.

### Steps to Reproduce

Having *AEM* running with *ACS Commons* installed:

1. Open the [Audit Log Search](http://localhost:4502/etc/acs-commons/audit-log-search.html) console,
2. Enter any suitable parameters,
3. Click *Search Audit Log*,
4. No results show up.

Moreover, an `undefined` label shows up:
![undefined](https://user-images.githubusercontent.com/5563546/56124819-cc02e700-5f77-11e9-85d5-848260ac278c.png)

I've additionally tested *Day CQ DAM Event Recorder*

![confinMgr](https://user-images.githubusercontent.com/5563546/56125011-3d429a00-5f78-11e9-8d04-2526ae19cceb.png)

and having some `cq:AuditEvent` nodes existing

![crxde](https://user-images.githubusercontent.com/5563546/56125055-56e3e180-5f78-11e9-9ae8-c22698298bc1.png)

I had no luck either

![audit-log-empty](https://user-images.githubusercontent.com/5563546/56125230-c1951d00-5f78-11e9-86e0-d180ead1913c.png)

klcodanr:
@reggie7 The problem seems to be that the query is being blocked due to node traversal and for whatever reason, it returns a 200 response. I added a check to make sure that it was retuning a valid response and added a bit more warning to the message. Could you possibly check this branch and make sure it resolves this issue:

https://github.com/Adobe-Consulting-Services/acs-aem-commons/tree/issue/audit-log-1834

reggie7:
@klcodanr - it's given me proper information back on the error inside your branch, thank you!

PR #1866, opened 2019-04-29
Resolving #1834 - Index Audit Log Search Failure

klcodanr:
Added clearer messaging around the effects of not having the index in place and adding a check to see if the execution was actually successful


davidjgonzalez:
thanks dan!

PR #1874, opened 2019-05-02
Improvement - Reports Cleanup

klcodanr:
Updating to use triple brackets for the examples, doing some code cleanup and updating servlets to use OSGi R6 Annotations


bundle/src/main/java/com/adobe/acs/commons/reports/models/ContainingPageReportCellCSVExporter.java
```diff
@@ -37,7 +39,7 @@
   public String getValue(Object obj) {
     Resource result = (Resource) obj;
     PageManager pageMgr = result.getResourceResolver().adaptTo(PageManager.class);
-    Page page = pageMgr.getContainingPage(result);
+    Page page = Optional.ofNullable(pageMgr).map(p -> p.getContainingPage(result)).orElse(null);
```
badvision: Instead of using orElse on line 42 you could combine the next line as another map call and then to orElse(null) in the return statement

PR #1879, opened 2019-05-06
Fix ReflectionUtil typo

reggie7:


badvision:
Can you just use #1879 in the changelog?  I'm ok with using PR numbers if there's no defect number.

kwin:
I think there is an overlap with #1883 and I am inclined to rather merge #1883 over this one as this also adds some unit tests.

PR #1882, opened 2019-05-07
Add common LOCALE in SpreadsheetTest

reggie7:
This way different machines will react in the same way to the tests.

I've had test failures, when running from a machine with Polish locale.


badvision:
Sorry, I only tested in reverse-Polish

reggie7:
@badvision - is it ok for merge like that?

csaboka:
#1864 is trying to solve the same issue, but by allowing a locale to be specified for Spreadsheet instead of forcing a specific global locale during test execution.

reggie7:
@csaboka I'm all cool with #1864, but it's been a while with no fix and I think my quick workaround would suffice in short run before the full solution gets in place finally. Meanwhile it's pretty annoying to work with this test failing :(

kwin:
@reggie7 I merged #1864 which should make this PR obsolete. Can you quickly cross-check and then close this PR?

reggie7:
@kwin - this particular test succeeds finall @ my machine, but the `master` does not compile properly as a whole anymore for me neither in `J8` nor in `J11`. But no worries, let me check tomorrow more deeply maybe it's some local issue.

reggie7:
Yup, works fine on a clean `master`. Thanks!

Issue #1881, opened 2019-05-06
i18n Sling Models Injector causes very poor performance for AEM Forms Adaptive Forms


blackfor:
## Required Information

* AEM Version : 6.4.4.0
* ACS AEM Commons Version: 4.0.0

### Actual Behavior

The New i18n injector (see [1]) causes very poor performance for AEM Forms Adaptive Forms. 

When this is present, the peformance of POST requests to FORMs decreases massively - with the CPU time being spent apparently in logging. 

As soon as the component 'com.adobe.acs.commons.models.injectors.impl.I18nInjector' is disabled, normal performance returns.

When the component is enabled, thread dumps show all POSTs to adaptive forms queueing up in the method, like so: 

`"10.43.0.96 [1557172281638] POST /content/forms/af/tg-claims/direct/us/status/jcr:content/guideContainer.af.dermis HTTP/1.1" #47559 prio=5 os_prio=0 tid=0x00007f1a3010c000 nid=0x7746 waiting on condition [0x00007f1a04590000]
   java.lang.Thread.State: WAITING (parking)
	at sun.misc.Unsafe.park(Native Method)
	- parking to wait for  <0x000000072527f518> (a java.util.concurrent.locks.ReentrantLock$NonfairSync)
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:836)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireQueued(AbstractQueuedSynchronizer.java:870)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:1199)
	at java.util.concurrent.locks.ReentrantLock$NonfairSync.lock(ReentrantLock.java:209)
	at java.util.concurrent.locks.ReentrantLock.lock(ReentrantLock.java:285)
	at ch.qos.logback.core.OutputStreamAppender.writeBytes(OutputStreamAppender.java:197)
	at ch.qos.logback.core.OutputStreamAppender.subAppend(OutputStreamAppender.java:231)
	at ch.qos.logback.core.rolling.RollingFileAppender.subAppend(RollingFileAppender.java:235)
	at ch.qos.logback.core.OutputStreamAppender.append(OutputStreamAppender.java:102)
[...]

PR #1883, opened 2019-05-07
#1881 - ReflectionUtil.isAssignableFrom always returns true

justinedelson:
This results in excess invocations of i18nProvider.i18n


kwin:
@justinedelson Isn't this covered already with #1879? Although your PR seems to have the benefit of also adding unit tests.

justinedelson:
good point @kwin I missed #1879. I would also rather merge this than #1879 since it includes verification that the issue described in #1881 is addressed.

PR #1884, opened 2019-05-07
Made the annotation check for i18n injector apply for both conditions.

niekraaijmakers:


justinedelson:
This would seem to be breaking backwards compatibility. No?

Issue #1888, opened 2019-05-09
Stylesheet Inliner breaks AEM Start Page and does not respect the selector


reggie7:
* [ ] AEM Version: `6.5`
* [ ] ACS AEM Commons Version: `4.0.0`
* [ ] Reproducible on Latest? yes

### Expected Behavior

[Stylesheet Inliner](https://adobe-consulting-services.github.io/acs-aem-commons/features/e-mail/stylesheet-inliner/index.html) works as described in the [docs](https://adobe-consulting-services.github.io/acs-aem-commons/features/e-mail/stylesheet-inliner/index.html#how-to-use).

### Actual Behavior

[Stylesheet Inliner](https://adobe-consulting-services.github.io/acs-aem-commons/features/e-mail/stylesheet-inliner/index.html) breaks the [AEM Start](http://localhost:4502/aem/start.html) page:

![aem-start](https://user-images.githubusercontent.com/5563546/57451481-ae1c5f80-7261-11e9-99b4-200df1807793.png)

(the screenshot comes from `6.3`, behavior in `6.5` is the same). What is more - the [docs](https://adobe-consulting-services.github.io/acs-aem-commons/features/e-mail/stylesheet-inliner/index.html#purpose) mention `inline-css` selector, but it's nowhere to be seen - the feature, when [turned on](https://adobe-consulting-services.github.io/acs-aem-commons/features/e-mail/stylesheet-inliner/index.html#how-to-use) - works on all pages.

### Steps to Reproduce

1. Follow the [docs](https://adobe-consulting-services.github.io/acs-aem-commons/features/e-mail/stylesheet-inliner/index.html#how-to-use),
2. Observe.

reggie7:
I've posted a [PR](https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io/pull/162) for the docs typos already.

reggie7:
The [PR](https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/1890) will fix the issues.

PR #1890, opened 2019-05-10
Fix (?) Stylesheet Inliner: respect the selector + do not break author

reggie7:
Fix for [Stylesheet Inliner breaks AEM Start Page and does not respect the selector](https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/1888) on the [Stylesheet Inliner](https://adobe-consulting-services.github.io/acs-aem-commons/features/e-mail/stylesheet-inliner/index.html) feature:

1. Sometimes the inliner is unable to get the content of a `css` file for the inline. In the previous version it would simply abandon these styles. I'm including them as `link` just as it would be without the feature enabled. This way - what can be inlined - indeed is. If we can't inline something - we leave it as is.
2. The `inline-css` selector was not respected before even though the [docs](https://adobe-consulting-services.github.io/acs-aem-commons/features/e-mail/stylesheet-inliner/index.html#how-to-use) explicitly mention it.

In a different [PR](https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io/pull/162) I'm fixing typos in the docs.


reggie7:
@badvision - I've updated the JUnit tests, please check.

badvision:
Thanks so much! 😃

PR #1893, opened 2019-05-14
add javax.annotation dependency to pom as it was removed in java 11

ahmed-musallam:
see: https://stackoverflow.com/questions/48204141/replacements-for-deprecated-jpms-modules-with-java-ee-apis/48204154#48204154 for more details.

With this addition, ACS commons can be compiled with Java 11.


ahmed-musallam:
@davidjgonzalez I always forget the changelog entry! Done!

bundle/pom.xml
```diff
@@ -730,6 +731,9 @@
             <version>6.0.0</version>
             <scope>provided</scope>
         </dependency>
-        
+        <dependency>
+            <groupId>javax.annotation</groupId>
+            <artifactId>javax.annotation-api</artifactId>
```
kwin: Shouldn't this rather have scope "provided" because in fact the necessary packages are provided by AEM? Also you would then be able to remove the additional include below `bannedDependencies`.

pom.xml
```diff
@@ -468,6 +468,11 @@
                 <version>0.9.1</version>
                 <scope>provided</scope>
             </dependency>
+            <dependency>
+                <groupId>javax.annotation</groupId>
+                <artifactId>javax.annotation-api</artifactId>
+                <version>1.3.2</version>
```
davidjgonzalez:
```suggestion
                <version>1.3.2</version>
                <scope>provided</scope>
```

davidjgonzalez:
```suggestion
                <version>1.3.2</version>
                <scope>provided</scope>
```

bundle/pom.xml
davidjgonzalez:
```suggestion
            <artifactId>javax.annotation-api</artifactId>
            <scope>provided</scope>
```

bundle/pom.xml
```diff
@@ -214,6 +214,7 @@
                                         <include>io.jsonwebtoken:jjwt</include>
                                         <include>com.jcraft:jsch</include>
                                         <include>com.jcraft:jzlib</include>
+                                        <include>javax.annotation:javax.annotation-api</include>
```
davidjgonzalez:
```suggestion
```

bundle/pom.xml
ahmed-musallam: @kwin good catch! I'm kicking myself for not catching that!

ahmed-musallam:
@davidjgonzalez sorry for the delay, that's done!

adamcin:
I think this now requires an `Import-Package: javax.annotation;version=0.0.0` override in the bundle manifest.

Issue #1868, opened 2019-04-30
MCP @FormField doesn't support namespace as property


gauravjasiwal:
## Required Information

* [6.4.3 ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: _______
* [4.0.0 ] ACS AEM Commons Version: _____
* [Yes ] Reproducible on Latest? yes/no

### Expected Behavior

FormField should support sling `@named` to provide namespace as the property name "xyz:myField". 

### Actual Behavior

We should be an option to provide a different property name with a namespace "xyz:myField". And FormField should support `@named `sling annotation.
Example:
  ```
    @Inject
    @Named(value = "viacom:requestLabel")
    @FormField(name = "Request Label", category = "General")
    private String requestLabel;
```

Note: Right now it just takes the name of the variable as the name of the jcr properties.

badvision:
@gauravjasiwal let me know if this patch works for you, you can find it in the feature branch.

gauravjasiwal:
@badvision Yes it's working, I tried in my local.

PR #1896, opened 2019-05-15
Added support for Named annotations in MCP Form Processor, fixing #1868

badvision:


badvision:
@klcodanr, @HitmanInWis -- FYI I'm expanding the capabilities of the MCP Forms framework so these can generate cq:dialog replacements in the near future.  First step is fixing all the little bugs, but it's getting there slowly. 😃

davidjgonzalez:
IIUC this correctly, dialogs will be generated from Java code; is this inline w/ AEM best practices for component development?  — ie. Is the aem core component team going to adopt this? Or is there some other use case being targeted?

badvision:
AEM Best Practices require you to jump through a lot of hoops.  I'm optimizing that.  Will the core component team adopt this?  I don't know, I really can't speak for them.  I think that if the people writing the product had to use it as much as our consultants did, they might have made some things easier to do with fewer moving parts.

davidjgonzalez:
We should be careful about introducing functionality that replaces Productized/Adobe Supported functionality, especially when its a DSL/API that is intrusive. We can wait to discuss the viability in more detail in the feature review GH issue though.

badvision:
I don't know of any productized AEM features that generate forms out of sling models directly.  IIRC the only way to generate dialogs in AEM today is to hand-write them as XML in content projects.  This allows alternative solutions to that.

klcodanr:
@badvision just to make sure I'm getting this, the idea is that you'd point this at a model and it would automatically generate out a dialog? 

IMO -- wow, I have a lot of nerd love for that idea. I do see a couple of problems:

- It tightly couples 1 model <-> 1 component
- Build cycles for Java code is / can be longer than updating a component
- Previously mentioned concerns about overlaying product, even if it makes development 10x easier, product doesn't align that well with how people use it, so keeping this up to date / not in conflict will be a challenge

I wonder what if we approached this from a different direction. The problem IMO is that generating complex JCR structures sucks. As a developer you have to write correct XML, keeping it formatted is a PITA, there's not really a good way to syntactically validate that the XML is correct... it's just a pain.

What if instead, you could have some sort of fluent DSL which would execute at build time to generate the XML for you? Ideally with a pluggable interface so you could do more than just dialogs. Imagine if you could do something like:

```
/dialog.jcr
dialog {
  title 'Simple Dialog'
  items {
    textfield {
      label 'Title'
      name 'title'
    }
    pathfield {
      base '/content/test'
      label 'Link'
      name 'link'
    }
  }
}```

Then when your project builds in Maven it validates that your code is correct and then generates out the XML for the dialogs. You could probably (and this is truly crazy-pants) have it generate out a model to use.

badvision:
For the purpose of this PR, which just fixes one specific bug to support the @named annotation, that's not relevant.  The actual bit you're asking about is not submitted yet.  That's an entirely separate PR I haven't started on just yet, but I do have it working locally.

There's a OSGi service which watches for sling models as they're registered, and if they extend the GeneratedDialog class then it sets up a ResourceProvider instance, using the resourceType of the sling model itself such that the resourceType + "/cq:dialog" is a generated form dialog.  All of the product UI functionality works otherwise like normal.  Does it support every little thing people want?  No, probably not but they can register additional classes in their own projects for custom components.  It's designed to handle the majority of cases where you just want a standard form with perhaps a few extras like tabs and/or accordions to break things up.

badvision:
You can break free of the 1 model = 1 component using subclasses which provide different resource types.  The general idea here is that instead of doing things by convention (with all the chaos that ensues) the tradeoff is the framework forces you into a contract, which is less flexible but is also much more predictable.

HitmanInWis:
FWIW regarding @klcodanr's recommendations above...

We literally just this last month at Bounteous built a component code generator that reads a JSON configuration file and kicks out a full dialog (including shared/global dialogs if applicable), a fully coded sling model (impl + interface), HTL file (importing the model) and empty JS/CSS files in a clientlib.  The output follows all current standards.

Is that what's being considered here?

badvision:
Not in this PR.  Separate concern, but essentially using annotations on the model bean as the generator for other stuff such as APis (later) and dialogs (also later, but kind of working now)

badvision:
All that being said (and that the actual dialog generation via resource provider is a separate feature that isn't added yet) are we ok to merge in this bugfix for the code that's already there?

PR #1900, opened 2019-05-17
New Feature: Tag Report MCP

klcodanr:
Adding a new MCP to create a report of the tags under a specified path. The report will contain all of the tag IDs, tag titles, reference count and optionally a list of the references.


klcodanr:
Thanks @badvision! So the reason for the "duplicate" functionality is that TagManager.find only searches the `cq:tags` property of `cq:Taggable` nodes so if you've created custom properties which have tags, this won't find them. Instead the `ReferenceFinder` class does a full text search for the term and then evaluates the results to throw out false positives and de-duplicate the results.

It also looks like my "minor change" broke the test... 🤦 So fixing that and the code climate issues

klcodanr:
Thanks @badvision -- so related question, what's the correct place to also get this into the 3.X release? I have some clients on older versions of AEM that would love this for pre-6.5 upgrade cleanup.

badvision:
Hmm, I don't know if we have a branch for that.  @justinedelson, @davidjgonzalez, thoughts?

PR #1905, opened 2019-05-23
Feature/httpcache responseheader exclusion

niekraaijmakers:
Make it possible to exclude certain response headers (like login cookies etc) .

Throttle the storing of responses (todo item)

Cleaned up the HttpCacheEngine a little.


niekraaijmakers:
1 out of 5 builds on travis push failed :/

badvision:
Can you open an issue for the random build fails?  I think it's one of the fam tests having a race condition.

badvision:
In the meantime I restarted the build.

bundle/src/main/java/com/adobe/acs/commons/httpcache/engine/package-info.java
```diff
@@ -18,6 +18,6 @@
  * #L%
  */
 
-@org.osgi.annotation.versioning.Version("3.3.0")
+@org.osgi.annotation.versioning.Version("4.0.0")
```
badvision: Is it entirely necessary to have such a large API version increase?  @justinedelson, thoughts on this?  I'm not adverse but I would rather refrain if possible.  That would mean that either this PR would have to wait for the next minor version bump potentially, or we would have to find a way to adapt these changes so that such a large increase isn't necessary.

niekraaijmakers: This plugin forced me to bump it up with this high number :( I am not sure of the rules or how this thing works exactly.. some insight would be good indeed

bundle/src/main/java/com/adobe/acs/commons/httpcache/engine/impl/HttpCacheEngineImpl.java
```diff
@@ -409,42 +278,82 @@ public HttpCacheServletResponseWrapper wrapResponse(SlingHttpServletRequest requ
...
+
+    }
+
+    private Map<String, List<String>> extractHeaders(SlingHttpServletResponse response, HttpCacheConfig cacheConfig) {
+
+        Map<String, List<String>> headers = new HashMap<>();
+        List<Pattern> excludedHeaders = Stream.concat(globalHeaderExclusions.stream(), cacheConfig.getExcludedResponseHeaderPatterns().stream())
+                .collect(Collectors.toList());
+
+        List<String> headerNames = new ArrayList<>();
```
davidjgonzalez: Arrays.stream(response.getHeaderNames())
.filter(headerName -> !isResponseHeaderExcluded(headerName, excludedHeaders))
.collect(Collectors.toMap(headerName -> headerName, headerName -> response.getHeaders(headerName)));

?

bundle/src/main/java/com/adobe/acs/commons/httpcache/engine/impl/HttpCacheEngineImpl.java
```diff
@@ -409,42 +278,82 @@ public HttpCacheServletResponseWrapper wrapResponse(SlingHttpServletRequest requ
...
             }
         }
 
+        return headers;
+    }
+
+
+    private boolean isResponseHeaderExcluded(String headerName, List<Pattern> responseHeaderExclusions) {
+
+        if (responseHeaderExclusions
```
davidjgonzalez: just return value of this expression vs conditional?

bundle/src/main/java/com/adobe/acs/commons/httpcache/engine/impl/HttpCacheEngineImpl.java
```diff
@@ -409,42 +278,82 @@ public HttpCacheServletResponseWrapper wrapResponse(SlingHttpServletRequest requ
...
+            throw new AssertionError("Programming error.");
+        }
+        final Map<String, List<String>> extractedHeaders = extractHeaders(responseWrapper, cacheConfig);
+        final int status = responseWrapper.getStatus();
+        final String charEncoding = responseWrapper.getCharacterEncoding();
+        final String contentType = responseWrapper.getContentType();
 
-        CacheContent cacheContent = null;
-        try {
+        throttledTaskRunner.scheduleWork(() -> {
```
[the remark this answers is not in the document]
niekraaijmakers: This was a TODO item, not sure who put it there, but mainly for JCR / Disk storage. True with mem stores it's not needed.

niekraaijmakers:
I implemented the review comments (thanks!), please revisit :)

bundle/src/main/java/com/adobe/acs/commons/httpcache/engine/CacheContent.java
```diff
@@ -132,8 +132,10 @@ public CacheContent() {
      * Construct from the custom servlet response wrapper..
      *
      * @param responseWrapper
+     * @deprecated
```
davidjgonzalez:
```suggestion
     * @deprecated Use build(HttpCacheServletResponseWrapper responseWrapper,int status, String charEncoding, String contentType, Map<String, List<String>> headers) throws HttpCacheDataStreamException
```

bundle/src/main/java/com/adobe/acs/commons/httpcache/config/HttpCacheConfig.java
```diff
@@ -28,6 +28,8 @@
 import java.util.List;
 import java.util.regex.Pattern;
 
+import static java.util.Collections.emptyList;
```
davidjgonzalez:
```suggestion
import java.util.Collections;
```

bundle/src/main/java/com/adobe/acs/commons/httpcache/config/HttpCacheConfig.java
```diff
@@ -77,6 +79,15 @@
      */
     List<Pattern> getJCRInvalidationPathPatterns();
 
+    /**
+     * Get a list of headers (as regex pattern) that should NOT be put in the cached response, to be served to the output.
+     * This is useful for example with systems that put a login cookie in each response.
+     * @return
+     */
+    default List<Pattern> getExcludedResponseHeaderPatterns() {
+        return emptyList();
```
davidjgonzalez: Generally better to qualify than making static import; this makes me go look at where it's coming from (ie. did you implement your own for some reason?) I tend to use static imports for constants or in Unit tests where its very obvious what is not provided by custom code (ie. assertXXX(..), mock(..), etc.)
```suggestion
        return Collections.emptyList();
```

bundle/src/main/java/com/adobe/acs/commons/httpcache/engine/CacheContent.java
```diff
@@ -132,8 +132,10 @@ public CacheContent() {
...
      * @return
      */
+    @Deprecated
     public CacheContent build(HttpCacheServletResponseWrapper responseWrapper) throws HttpCacheDataStreamException {
         this.status = responseWrapper.getStatus();
```
davidjgonzalez: Should just call non-deprecated `build(..)` method, rather than duplicating work.

```suggestion
          // Extracting HTTP Response Header Names and Values
         Map<String, List<String>> headers = new HashMap<>();
        List<String> headerNames = new ArrayList<>();

        headerNames.addAll(responseWrapper.getHeaderNames());
        for (String headerName: headerNames) {
            List<String> values = new ArrayList<>();
            values.addAll(responseWrapper.getHeaders(headerName));
            headers.put(headerName, values);
        }

        return build(responseWrapper, responseWrapper.getStatus(), responseWrapper.getCharacterEncoding(), responseWrapper. getContentType(), headerNames);
```

bundle/src/main/java/com/adobe/acs/commons/httpcache/config/HttpCacheConfig.java
```diff
@@ -37,7 +39,7 @@
 @ProviderType
 public interface HttpCacheConfig {
 
-    public enum FilterScope {
```
davidjgonzalez: Why did this drop in scope? This is a public package, and this go dropped to default scope; if someone were to use this enum, wouldnt this break the code?

niekraaijmakers: This is redundant.. since it's interface it's marked public by default since JDK 7.

niekraaijmakers:
updated.. :) please review again

Issue #1835, opened 2019-04-15
'Paint Connections` break the view in 'Version Compare'


reggie7:
* [ ] AEM Version: `6.5`
* [ ] ACS AEM Commons Version: `4.0.0`
* [ ] Reproducible on Latest? yes

### Expected Behavior

Using **Paint Connections** option with other checkboxes leaves the connections pointing from source to target properly.

### Actual Behavior

When one uses **Paint Connections** option followed by other checkboxes the first connection starts in top-left corner of the console page.
Combining it with **Hide Unchanged** leaves connections pointing to empty space on the right side.

### Steps to Reproduce

1. Choose a page to work on, e.g. [Arctic Surfing in Lofoten](http://localhost:4502/editor.html/content/we-retail/language-masters/en/experience/arctic-surfing-in-lofoten.html),
2. Use its [Timeline](https://helpx.adobe.com/experience-manager/6-5/sites/authoring/using/working-with-page-versions.html) to create a version,
3. Add some random changes (I've removed a paragraph) under [Arctic Surfing in Lofoten](http://localhost:4502/editor.html/content/we-retail/language-masters/en/experience/arctic-surfing-in-lofoten.html),
4. Open [Version Compare](http://localhost:4502/etc/acs-commons/version-compare.html?path=/content/we-retail/language-masters/en/experience/arctic-surfing-in-lofoten),
5. Check **Paint Connections**,
6. Check **Hide Unchanged**,
7. Check **Hide Versions** (choose random ones).

**Hide Unchanged** will produse such artifacts:

![hide-unchanged-2](https://user-images.githubusercontent.com/5563546/56127589-d0320300-5f7d-11e9-80ce-d5a09a83c411.png)

**Hide Versions will yield:

![version-compare](https://user-images.githubusercontent.com/5563546/56127736-20a96080-5f7e-11e9-8fd3-2a82ff48097a.png)

Their combination:

![hide-unchanged](https://user-images.githubusercontent.com/5563546/56127752-2a32c880-5f7e-11e9-89e2-758c7c523df0.png)

reggie7:
I'm on it BTW :)

reggie7:
Fixed in [1908](https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/1908).

davidjgonzalez:
Fixed via #1933

PR #1908, opened 2019-05-24
Cleanups: Fix broken paint connections in Version Compare

reggie7:
Fix for [1835](https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/1835). Actually the fix is within frontend files. Java cleanup is a side-effect, it turned out that the BE output was enough already. But I'm leaving those in as e.g. code duplication was reduced, etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 - #1881 - Fixed issue where ReflectionUtil.isAssignableFrom() returned false positive result.
 - #1888 - Fixed issues with Stylesheet Inliner.
 - #1836 - Allow uniform download links in JCR Compare
+- #1835 - all options work together now and do not break the connections placement anymore
 
 ## [4.0.0] - 2019-02-20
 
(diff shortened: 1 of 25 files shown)


davidjgonzalez:
@reggie7 thanks for fixing! This PR's changes are really hard to review; Any ideas how/why this might be happening? If not, I'll take your word it's fully functional, as trying to truly grok what's changed/how is pretty hard. :)

kwin:
I guess this is due to some IDE autoformatting. Can we get a cleaner version of a diff somehow?

reggie7:
@davidjgonzalez well yeah, true, I've made some refactors in BE that after all do not affect the solution so much. In the first attempt I needed them, but then a much simpler FE-only approach was enough. Though following the [boy scout rule](https://deviq.com/boy-scout-rule/) - I've left those BE adjustments in place anyway. The gist is in the non-`java` files.
There's an option to split this PR into smaller chunks or drop the refactor, if needed.

reggie7:
[1933](https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/1933) has the fix, let's first focus on the other one, then this one will cover only the refactors.

PR #1920, opened 2019-05-29
New @ChildRequest model injector

HitmanInWis:
Added new @ChildRequest annotation to substitute for @ChildResource when a child model object requires a SlingHttpServletRequest to adapt from.


HitmanInWis:
Included a refactor of the tests for `@SharedValueMapValue` to bring in line with the structure of other annotation tests.

kwin:
Can we rather aim to fix this in Sling?

HitmanInWis:
https://issues.apache.org/jira/browse/SLING-8279 is another sling issue that could be resolved to address the issue that the new `@ChildRequest` annotation is solving, but that would be much more far reaching.  I'm not opposed to this being fixed in Sling, but given that SLING-5726 has been open for years and SLING-8279 has far reaching implications, my thought is there is no harm in this annotation being made available in ACS Commons in the meantime, and deprecating it later if Sling eventually has a solution.

And I know `@ChildRequest` may seem a bit of an odd name, but it's named as such since it is analogous to `@ChildResource` but instead acts on a `Request` instead of a `Resource`.

HitmanInWis:
BTW, the reason Im pushing for this in ACS Commons is:
- This is necessary for any codebase currently using WCM Core components where models have children to inject as models.
- This is leveraged by an AEM component code generator that Bounteous has open sourced at https://github.com/Bounteous-Inc/bounteous-aem-commons/tree/master/aem-component-generator

niekraaijmakers:
Can we not use the @ChildResource annotation and implement another Injector that would support sling request? With a different service.ranking ? Just thinking out loud here :)

HitmanInWis:
@niekraaijmakers perhaps - the logic is already all in the `ChildRequestInjector` class which could potentially be coded to handle cases of `@ChildResource` but to me it seems more safe to have a dedicated annotation since it is adapting a request instead of a resource.  However, Im open to discussion on it if we see a fundamental problem with this new annotation/injector.

niekraaijmakers:
@justinedelson @davidjgonzalez WDYT?

davidjgonzalez:
I'm leery about "replacing" the `@ChildResource` annotation. If this is changed in the future then there is an onus on this project to update, and it also creates confusion since it's not obvious to developers which impl is being used (ie. a new developer on project googles `@ChildResource`, they find Sling's docs, but it's not behaving the way its documented... confusing)

FWIW name `@ChildRequest` was unclear to me... it made it sounds like there are multiple, HTTP requests, which doesn't make sense :)..  IIUC what this is really doing is more `@ChildResourceWithRequest` ... where that Resource is adaptable from either a Request or Resource? .. ultimately, the intent here is to return a collection of Sling Models, adapted from either a Resource or Request, right? so maybe `@ChildResourceAsSlingModel` or `@ChildResourceAsRequestAdaptableSlingModel` for the verbose...

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/ChildRequestInjector.java
```diff
@@ -0,0 +1,149 @@
...
+                }
+            }
+        } else if (adaptable instanceof Resource) {
+            logger.trace("Injecting '{}' from resource (request not available)", name);
+            return getValueForResource((Resource) adaptable, name, declaredType);
+        }
+        return null;
+    }
+
+    private Object getValueForRequest(SlingHttpServletRequest request, String name, Type declaredType) {
```
davidjgonzalez: This and `getValueForResource` looks identical except for 1 line; could this be abstracted to reduce code dupe?

bundle/src/main/java/com/adobe/acs/commons/util/OverridePathSlingRequestWrapper.java
```diff
@@ -0,0 +1,103 @@
...
+    @Override
+    public Resource getResource() {
+        return this.resource;
+    }
+
+    /**
+     * Overriding `adaptTo` to avoid using the original request as the adaptable.
+     */
+    @Override
+    public <AdapterType> AdapterType adaptTo(Class<AdapterType> type) {
```
davidjgonzalez: Per docs when null is returned, a message should be logged.

https://sling.apache.org/apidocs/sling9/org/apache/sling/api/adapter/AdapterFactory.html#getAdapter-java.lang.Object-java.lang.Class-

bundle/src/main/java/com/adobe/acs/commons/util/OverridePathSlingRequestWrapper.java
```diff
@@ -0,0 +1,103 @@
...
+     * @param wrappedRequest SlingHttpServletRequest
+     * @param relPath        String
+     */
+    public OverridePathSlingRequestWrapper(final SlingHttpServletRequest wrappedRequest, final String relPath) {
+        super(wrappedRequest);
+
+        SlingBindings slingBindings = (SlingBindings) getSlingRequest().getAttribute(ATTR_SLING_BINDINGS);
+
+        this.adapterManager = slingBindings.getSling().getService(AdapterManager.class);
+        this.resource = getSlingRequest().getResourceResolver().resolve(getSlingRequest(),
```
davidjgonzalez: Could this be simplified to: 

`getSlingRequest().getResource().getChild(relPath);` ?

The use of `resolve(..)` and `Text.fullFilePath(..)` make it seem there is more complexity in resolving the relative path, but im not clear what that is (or if this isnt around about way of doing the above).

https://sling.apache.org/apidocs/sling7/org/apache/sling/api/resource/Resource.html#getChild-java.lang.String-

bundle/src/main/java/com/adobe/acs/commons/models/injectors/annotation/ChildRequest.java
```diff
@@ -0,0 +1,56 @@
...
+ * require a `SlingHttpServletRequest`, the loss of the `SlingHttpServletRequest` makes the
+ * `@ChildResource` injector an insufficient solution.
+ *
+ * See https://issues.apache.org/jira/browse/SLING-8279 for more context on the issue that this
+ * via provider solves.
+ */
+@Target({ElementType.METHOD, ElementType.FIELD, ElementType.PARAMETER})
+@Retention(RetentionPolicy.RUNTIME)
+@InjectAnnotation
+@Source("child-requests")
```
davidjgonzalez: As noted in comments, I don't find the annotation name clear - My personal preference is a more descriptive name at the cost of being more verbose.

bundle/src/main/java/com/adobe/acs/commons/util/OverridePathSlingRequestWrapper.java
```diff
@@ -0,0 +1,103 @@
...
+                Text.fullFilePath(getSlingRequest().getRequestPathInfo().getResourcePath(), relPath));
+        this.myBindings.putAll(slingBindings);
+        this.myBindings.put(WCMBindings.PROPERTIES, this.resource.getValueMap());
```
davidjgonzalez: This could be all that's needed, but double checking that you're sure this is all that's needed. If it is, cool!

HitmanInWis:
Totally cool with @ChildResourseWithRequest.  Will update and address other review comments when back in the office next week.

bundle/src/main/java/com/adobe/acs/commons/util/OverridePathSlingRequestWrapper.java
HitmanInWis: I believe this would actually be the responsibility of the `AdapterFactory` implementation itself, since only it would have the info as to why the adaptation failed.  This function is simply delegating to the adapter factory implementations, similar to `SlingAdaptable` which also does not log a `null` return from `adaptTo`.

bundle/src/main/java/com/adobe/acs/commons/util/OverridePathSlingRequestWrapper.java
HitmanInWis: I believe @adamcin did it this way to ensure null safety of the retrieved resource in the case that a path does not exist, since `resolve` will return a synthetic resource in that case.

HitmanInWis: Also, `Text.fullFilePath(..)` is used to support both relative and absolute paths.  I'll update the var name and docs accordingly.

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/ChildRequestInjector.java
HitmanInWis: through this magics of java 8, this is resolved

bundle/src/main/java/com/adobe/acs/commons/models/injectors/annotation/ChildRequest.java
HitmanInWis: This is resolved.

bundle/src/main/java/com/adobe/acs/commons/util/OverridePathSlingRequestWrapper.java
HitmanInWis: It's all we've needed thus far :D.

I did add a binding for current page now as well.

bundle/src/main/java/com/adobe/acs/commons/util/OverridePathSlingRequestWrapper.java
davidjgonzalez: so returning a non existing resource is a valid use case? When would requestPathInfo.getResourcePath() be relative? 

I think adding some comments to the code would be helpful clarifying these seemingly? corner cases.

bundle/src/main/java/com/adobe/acs/commons/util/OverridePathSlingRequestWrapper.java
davidjgonzalez: Aren’t you returning null here if adaptermanger is null? So it seems it should log.warn(“unable to adapt x to y because AdapterManager is null”); ?
Of course if that’s null that’s a different, worse problem, but this should log or per docs?

davidjgonzalez:
Looks good - I’m curious more than anything about when non existing resources would be used, and/or when requestPathInfo resourcePath is relative. Would be good to make a note of why In the comments so someone in the future doesn’t try to “clean it up” on accident.

The log.warn in adaptto still seems applicable since your code is deciding to return null in 1 condition. (All others are delegated as you note). Code looks good tho!

bundle/src/main/java/com/adobe/acs/commons/util/OverridePathSlingRequestWrapper.java
HitmanInWis: Updated OverridePathSlingRequestWrapper to no longer support relative paths, as it was an unnecessary feature and more likely to cause bugs then be truly helpful.

Added test for OverridePathSlingRequestWrapper to ensure support for requests to non-existent paths, since sling seems to generally support this concept and the only other option would be to throw an exception from the constructor.

bundle/src/main/java/com/adobe/acs/commons/util/OverridePathSlingRequestWrapper.java
HitmanInWis: I guess its an edge case possibility, so I covered it now.

HitmanInWis:
Thank you!

davidjgonzalez:
Thank YOU!

kwin:
Just for the record: I opened https://issues.apache.org/jira/browse/SLING-9674 for having such a capability in Sling some day.

Issue #1904, opened 2019-05-22
Feature Review: Fully automatic component dialogs


badvision:
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.4+

### Feature Brief

Building on the synthetic form generation in MCP, expand to let developers create component dialogs using just the sling models.  Basically, no more need to hand-edit cq:dialog nodes.

### How this would work

![image](https://user-images.githubusercontent.com/5144198/58193524-2f77e700-7c91-11e9-8162-f2c56b8dbff5.png)

This requires the introduction of an OSGi service factory, DialogResourceProviderFactory, which binds to all declared sling models (via AdapterFactory registered instances.)  If an adapter factory is an exported class and is a subclass of GeneratedDialog, then it is considered a candidate for this feature.

Candidate sling models are then inspected for their provided resource type.  This can be provided in the class-level annotation which is part of sling models.   This can also be provided in a getter method or field annotated accordingly as the resource type.

The resource type is then pre-pended with "/apps/" and appended with "cq:dialog" to construct the target path of the resource provider root location.   A resource provider service is constructed and registered at runtime with this location, such that any requests to this resource are relative to the generated synthetic form resource.  

An additional optional feature is to disable sling merger behaviors so that there are no additional form elements (to be decided on later) -- this is likely not a big deal for components but if using this feature on pages it will be necessary to prevent additional garbage from showing in the generated form.

The result is that a developer only need create a sling model with proper annotations and the dialog is then always correct and always relevant to the model.  There is no need to update a secondary location every time the component needs to be updated or revised.

![image](https://user-images.githubusercontent.com/5144198/58193888-1a4f8800-7c92-11e9-94eb-3436a0963ef1.png)

badvision:
@klcodanr, @HitmanInWis , @davidjgonzalez  this is the not-yet-added feature I was describing in the other recent PR's.  Those PR's do not include this feature but do support its eventual addition, but those PRs mostly just fix current bugs which would otherwise impede the success of a feature such as this.

joerghoh:
Sounds good and definitely something we should add. Might not solve all issues with generating dialogs, but for the majority of the simple component dialogs it's probably sufficient.

Some questions:
* you register a ResourceProvider for each of these models? Have you already tested if this has an performance impact on general resource lookup? Or would it be possible to register a single ResourceProvider to handle all of these models?
[...]

PR #1922, opened 2019-05-30
Feature/dialog resource provider #1904

badvision:
As discussed here is the dialog resource provider feature.  It incorporates the disable switch suggested by @bstopp and also is more flexible with annotation support recommended by @justinedelson

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/DialogResourceProviderFactoryImpl.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/DialogResourceProviderFactoryImpl.java
@@ -26,6 +26,15 @@ public class DialogResourceProviderFactoryImpl implements DialogResourceProvider
     private final Map<Class, ServiceRegistration<ResourceProvider>> resourceProviders = Collections.synchronizedMap(new HashMap<>());
     private DialogResourceProviderConfiguration config;
     private final Set<String> allKnownModels = Collections.synchronizedSet(new HashSet<>());
+    private String[] ignoredPackages = new String[]{
+        "com.adobe.cq.",
+        "com.adobe.aemds.",
+        "com.adobe.fd.ccm.",
+        "com.adobe.forms.",
+        "com.adobe.granite.",
+        "com.day.cq.",
+        "we.retail."
+    };
 
     @Reference(
             service = AdapterFactory.class,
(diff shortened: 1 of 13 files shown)


badvision:
Let me know if you see any other anti-patterns that need resolving @justinedelson.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/DialogResourceProviderFactoryImpl.java
```diff
@@ -0,0 +1,208 @@
...
+import org.osgi.service.component.annotations.Reference;
+import org.osgi.service.component.annotations.ReferenceCardinality;
+import org.osgi.service.component.annotations.ReferencePolicy;
+import org.osgi.service.component.annotations.ReferencePolicyOption;
+import org.osgi.service.metatype.annotations.Designate;
+import org.slf4j.LoggerFactory;
+
+@Component(
+        service = DialogResourceProviderFactory.class,
+        immediate = true
```
shsteimer: wouldn't it be better to make this as configurationPolicy=required rather than using an explicit enabled flag?  It sorta does the same thing, but that would disable it at an osgi framework level instead of a code level.

badvision: Well, no in this case I need the service aware of things that are registered so I need the service "running" but when it's not enabled it just doesn't create or register the resource providers.  This kind of operation also allows it to work properly when the OSGi service is reconfigured, such as if it is enabled later in the lifecycle via configuration after several sling models have already been registered as AdaperFactory instances.   Since catching registration of sling models is already really a weird thing, I'm more comfortable leaving this one as it is.  For what it's worth, there is a unit test that ensures disable/enable of the service works properly.

badvision: Good call though @shsteimer  -- in a normal OSGi service I'd be happy with that, but this one is a slight step into "weird" territory and there's only so much that OSGi mocks let me test, so this way is probably safer.

shsteimer: That makes sense.  I am now of course curious and want to find time to play with it to see if I can make it work without that need, but if you say it's required that's good enough for me.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/DialogResourceProviderFactoryImpl.java
```diff
@@ -0,0 +1,208 @@
...
+)
+@Designate(ocd = DialogResourceProviderConfiguration.class)
+public class DialogResourceProviderFactoryImpl implements DialogResourceProviderFactory {
+
+    private static final org.slf4j.Logger LOG = LoggerFactory.getLogger(DialogResourceProviderFactoryImpl.class);
+    private static final String IMPLEMENTATION_CLASS = "models.adapter.implementationClass";
+    private final Map<String, ServiceRegistration<ResourceProvider>> resourceProviderRegistrations
+            = Collections.synchronizedMap(new HashMap<>());
+    private final Set<String> allKnownModels = Collections.synchronizedSet(new HashSet<>());
+    private final String[] ignoredPackages = new String[]{
```
shsteimer: probably not for version 1, but I'd be inclined to make this ignored packages setting part of the `DialogResourceProviderConfiguration`. Would make it a bit more flexible overall.

badvision: It could but honestly it might not matter too much.  This just helps avoid scanning the hundred or so adapter factories that trigger from product stuff.  For customer bundles there would have to be a lot of additional model packages for additional ignore rules to do anything useful, IMHO.

badvision:
@shsteimer, let me know if you have any other concerns otherwise I'm good to merge I think.

PR #1925, opened 2019-06-04
Adjust PowerMockErrorPageCacheImplTest to Java 11

reggie7:


bundle/src/main/java/com/adobe/acs/commons/util/ResourceDataUtil.java
```diff
@@ -21,49 +21,22 @@
...
-import com.adobe.acs.commons.util.BufferedServletOutput.ResponseWriteMethod;
 import com.day.cq.commons.jcr.JcrConstants;
 
 @ProviderType
 @SuppressWarnings({"checkstyle:abbreviationaswordinname", "squid:S1118"})
-public class ResourceDataUtil {
-    public static final String ENCODING_UTF_8 = "UTF-8";
-    private static final Logger log = LoggerFactory.getLogger(ResourceDataUtil.class);
-
-    public static String getIncludeAsString(final String path, final SlingHttpServletRequest slingRequest,
```
davidjgonzalez: @reggie7 we cannot remove methods like this; this is an exported API -- removing it would break all code that might depend on it.

reggie7: Oh, sorry, I was thinking locally (there was just the one usage), forgot that someone might use it as a library method simply :) Restored it. Thanks for noticing it!

reggie7:
@davidjgonzalez - I've seen https://stackoverflow.com/questions/52966897/powermock-java-11 indeed, but went for the option to remove `PowerMock` dependency altogether in the test. It's not perfect either way, but I guess still a step in better direction.

reggie7:
@davidjgonzalez, regarding cleanups - sure, let me adjust my future PRs then.
BTW though, at the moment all changes relate to the fixed tests already, there is no additional cleanup, or is it, from your PoV?

PR #1927, opened 2019-06-05
Feature/httpcache response cookie exclusion

niekraaijmakers:
Making it possible to also exclude specific cookie keys in the http cache.


bundle/src/main/java/com/adobe/acs/commons/httpcache/config/impl/HttpCacheConfigImpl.java
```diff
@@ -218,6 +218,13 @@
     private List<String> cacheHandlingRulesPid;
 
 
+    @Property(label = "Config-specific HttpCacheHandlingRules",
```
davidjgonzalez: The label and description looks incorrect.

PR #1932, opened 2019-06-07
JDK11 support

davidjgonzalez:
testing..

PR #1940, opened 2019-06-10
Adjust EntryNodeMapVisitorTest to Java 11

reggie7:


davidjgonzalez:
@reggie7 I recently went through a similar exercise on ASC, and needed to update a handful of dependencies to get this to work. Any idea if this is possible here rather than touching all these tests?
I updated deps and find/replaced some deprecated packages and removed some unnecessary stubbing and seemed to get pretty far just doing that.

I added the open failures in the comments on the PR below; the biggest worries seem to the MCP stuff since that relies on 3rd party classes, and the JUnitProvider issue -- not sure what the deal w that is (haven looked it up).

https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/1932

PR #1949, opened 2019-06-12
Adjust BrandPortalSyncProcessTest to Java 11

reggie7:
Dropping the `PowerMockito` in favour of `protected` method `resolveToAsset` - not a perfect solution, but well, at least it's working...


bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/BrandPortalSyncProcess.java
```diff
@@ -97,28 +100,43 @@ public final void execute(WorkItem workItem, WorkflowSession workflowSession, Me
...
+        }
+
+        if (StringUtils.equalsIgnoreCase(processArgs, ReplicationActionType.DEACTIVATE.getName())) {
             return ReplicationActionType.DEACTIVATE;
         }
 
         return null;
     }
+
+    protected Asset resolveToAsset(final Resource resource) {
```
davidjgonzalez:
```suggestion
   /* This is broken out into its own method to allow for easier unit testing */
    protected Asset resolveToAsset(final Resource resource) {
```

PR #1957, opened 2019-06-14
Adjust MultiFieldPanelWCMUseTest to Java 11

reggie7:


bundle/src/main/java/com/adobe/acs/commons/widgets/package-info.java
```diff
@@ -20,5 +20,5 @@
 /**
  * JSP Functions for working with custom widgets.
  */
-@org.osgi.annotation.versioning.Version("1.1.1")
+@org.osgi.annotation.versioning.Version("1.2.0")
```
davidjgonzalez: A test update shouldnt be changing package export versions! :(

davidjgonzalez:
@reggie7 a general note for all these tests - im not going to have the time to look through all the test changes and grok the changes to the actual Impls (tbh, the *only* changes to the impls should be things that make testing materially easier; like wrapping calls to the AEM static methods in mockable methods).

There's just too many changes to Impl's for me to take the time to understand and sanity check that they aren't breaking something (while we have unit tests, they absolutely aren't complete - so i don't trust the tests themselves to ensure uncovered cases aren't breaking).

If the changes were ONLY to the test files, then I wouldn't have any problems merging.

bundle/src/main/java/com/adobe/acs/commons/widgets/package-info.java
reggie7: I agree, but otherwise we get `[ERROR] com.adobe.acs.commons.widgets: Version increase required; detected 1.1.1, suggested 1.2.0` on build.

reggie7:
@davidjgonzalez let me revert the non-test non-essential changes then simply.

PR #1963, opened 2019-06-19
Adjust RemoteAssetDecoratorTest to Java 11

reggie7:


HitmanInWis:
Hi Reggie,

Can I get some background on why the refactors?  Were they needed to make it work on Java 11 or satisfy new coding standards?

At a glance the changes look fine - haven't yet tested or done a full review, but I will.

reggie7:
@HitmanInWis - the changes are all to make the tests run in `J11`. I found it much more difficult to make them work with the new Java as-is :( so I've concluded it's easier to rewrite it to keep the coverage.

I'll appreciate any input, if changes are required.

HitmanInWis:
Will try to get to this next week.

bundle/src/test/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetDecoratorTest.java
```diff
@@ -19,318 +19,248 @@
  */
```
HitmanInWis: I'm amazed at your ability to modularize code and tests, but I question why the complete rewrite of the work I did here to make this work with an actual sling implementation - i.e. as "real" as possible, using sling mock to have an in-memory JCR.  Though the new test do test functionality on an individual function level, the old tests I feel had an additional value of testing things truly working.  For example, a call to `resourceResolver.getResource()` was doing a real fetch of a real Resource, which was getting decorated by the RemoateAssetDecorator in the natural way it happens in sling.  The new tests have lost all of that.

Were this an original code submission, I'd be completely happy with these tests.  However, given that these tests were already written and in some ways perceivably more valuable, I struggle a bit with these changes.

If all we needed to do was remove PowerMockito, I think your wrapper function around `AccessControlUtil.getUserManager` was already sufficient, as the `remoteAssetDecorator` in the old tests was already a `spy` that can have the new `getUserManager` function mocked.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetDecorator.java
```diff
@@ -85,7 +89,7 @@ public Resource decorate(final Resource resource) {
             if (!this.accepts(resource)) {
                 return resource;
             }
-        } catch (Exception e) {
+        } catch (final RepositoryException e) {
```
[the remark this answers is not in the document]
reggie7: Ok, restored.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetDecorator.java
```diff
@@ -135,19 +142,24 @@ public Resource decorate(final Resource resource, final HttpServletRequest reque
...
         }
 
-        Calendar lastFailure = props.get(RemoteAssets.REMOTE_SYNC_FAILED, (Calendar) null);
-        if (lastFailure != null && System.currentTimeMillis() < (lastFailure.getTimeInMillis() + (this.config.getRetryDelay() * 60000))) {
+        final Calendar lastFailure = props.get(RemoteAssets.REMOTE_SYNC_FAILED, (Calendar) null);
+        if (lastFailure != null && System.currentTimeMillis() < (lastFailure.getTimeInMillis() + (config.getRetryDelay() * 60000))) {
+            return false;
+        }
+
+        final boolean isAllowedUser = isAllowedUser(resource);
```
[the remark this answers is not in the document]
reggie7: Great catch, fixed, thanks for noticing it!

bundle/src/test/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetDecoratorTest.java
reggie7: Well, my intention was to just make the build work under `J11`, nothing more. Usually, while reading new code unknown to me, I also apply [The Boy Scout Rule](https://deviq.com/boy-scout-rule/), but after a couple of PRs guys here convinced me to drop it :) I am not sure atm whether it was before or after this PR though.

I've obviously tried the fast and simple approach that you mention, but since it did not work just like that and since it still failed as I had to digg deeper and deeper - I've concluded that a rewrite with dropping the deeper checks approach will simply be much faster. It was simple time economics :) 

So actually I don't care whether the `PowerMockRunner` stays or not and whether my testing approach will replace the existing one or not, as long as the build will pass under `Java 11`...
Do you think you could try and check yourself how to fix your tests as you suggest above? Let me know how I can help.

HitmanInWis: Ok, I'll take a look at this when I get a chance.  How urgent are these java 11 updates?  Will have to squeeze this in somewhere.

HitmanInWis:
Closing in favor of https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/1976, which fixes the test errors much less invasively.

@reggie7 feel free to submit a PR cleaning up other cosmetic issues if you would like.

reggie7:
@HitmanInWis great job, simple as [that](https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/1976/files#diff-2eecae8dc1f9c1c2a8f823ee618ace4cL554)...

I'll skip any cleaning, there's  no need for it, no time for it :)

HitmanInWis:
Ok, thanks again for bringing this to my attention @reggie7 and for your willingness to contribute!

Issue #1956, opened 2019-06-14
New feature: OnDeployScripts support for external calls & logging to historic node


geertsmichael:
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: AEM 6.4.4
* [x] ACS AEM Commons Version: latest
* [x] Reproducible on Latest? yes

### Expected Behavior

I want to use the OnDeployScriptExecutor to call from my own code to trigger a script again or run the failed scripts again. Do you approve that I move this outside the impl package and export this class for external reference:

https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutor.java

I also want to add support to the OnDeployScriptBase to support logging to the runnedNode in /var/acs-commons/on-deploy-scripts-status. So the OnDeployScript can output, and can be checked instead of checking the log files.

### Actual Behavior

The class is hidden for referencing from my OSGI bundle. No logging to /var/acs-commons/on-deploy-scripts-status is available.

### Steps to Reproduce

I just want to have push access and approval for these features. I will modify the code myself and open a Pull Request.

### Links

https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutor.java

geertsmichael:
Can you grant me access for contribution to this project?

davidjgonzalez:
Just PR your contribution .. all contributors follow this process:

https://adobe-consulting-services.github.io/acs-aem-commons/pages/contribute.html

PR #1967, opened 2019-06-20
Expose OnDeployExecutor Interface and store exception message on result node

geertsmichael:
Expose OnDeployExecutor Interface: so scripts can be triggered externally
Store exception message on result node: so it can be read


HitmanInWis:
Will try to get to this next week.

bundle/src/main/java/com/adobe/acs/commons/ondeploy/OnDeployExecutor.java
```diff
@@ -17,7 +17,7 @@
  * limitations under the License.
  * #L%
  */
-package com.adobe.acs.commons.ondeploy.impl;
+package com.adobe.acs.commons.ondeploy;
```
shsteimer: Since I wrote the MBean support for this, I feel qualified to comment here.  I love the idea of having a better public API for executing these that fully supports updating status nodes, etc.  But I don't personally think the current internal mbean api makes for a suitable public API.  It is "stringly" typed due to the constraints of JMX, and it exposes the JMX TabularDataSupport object which feels wrong to me.

I would rather leave the Mbean interface internal only (maybe rename it to OnDeployExecutorMBean?), and then create a new interface for public usage, that uses the `OnDeployScript` and `OnDeployScriptProvider` interfaces rather than passing around class names/packages as strings.  More or less it could expose the existing runScript and runScript methods from OnDeployExecutorImpl, but also add support for a force parameter, to force re-execution of already completed scripts.

Will defer to @HitmanInWis and others for any decisions, but just sorta my $.02.

bundle/pom.xml
```diff
@@ -7,7 +7,7 @@
     <parent>
         <groupId>com.adobe.acs</groupId>
         <artifactId>acs-aem-commons</artifactId>
-        <version>4.2.1-SNAPSHOT</version>
```
HitmanInWis: Individual PRs should not modify the version in the pom files.  The version is already SNAPSHOT so there is no need to up it for testing.

bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutorImpl.java
```diff
@@ -226,11 +229,14 @@ protected void runScripts(ResourceResolver resourceResolver, List<OnDeployScript
         }
     }
 
-    protected void trackScriptEnd(Resource statusResource, String status) {
+    protected void trackScriptEnd(Resource statusResource, String status, String output) {
         try {
             ModifiableValueMap properties = statusResource.adaptTo(ModifiableValueMap.class);
             properties.put(SCRIPT_STATUS, status);
             properties.put(SCRIPT_DATE_END, Calendar.getInstance());
+            if (output != null) {
```
HitmanInWis: This should just be `properties.put(SCRIPT_OUTPUT, output)` even if the `output` is null because you would want a successful re-run of a script to overwrite the field if it currently has a value in it.

bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutorImpl.java
```diff
@@ -197,12 +200,12 @@ protected boolean runScript(ResourceResolver resourceResolver, OnDeployScript sc
             try {
                 script.execute(resourceResolver);
                 logger.info("On-deploy script completed successfully: {}", statusResource.getPath());
-                trackScriptEnd(statusResource, SCRIPT_STATUS_SUCCESS);
+                trackScriptEnd(statusResource, SCRIPT_STATUS_SUCCESS, null);
```
HitmanInWis: Should `trackScriptStart` be updated to null out the `output` property on the status node as well?

content/pom.xml
```diff
@@ -7,7 +7,7 @@
     <parent>
         <groupId>com.adobe.acs</groupId>
         <artifactId>acs-aem-commons</artifactId>
-        <version>4.2.1-SNAPSHOT</version>
+        <version>4.3.0-SNAPSHOT</version>
```
HitmanInWis: Same comment here regarding the versions.

HitmanInWis:
Can you please help me understand the use case for `Expose OnDeployExecutor Interface: so scripts can be triggered externally`?

davidjgonzalez:
@HitmanInWis I believe @geertsmichael discussed the use-case here: https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/1956

HitmanInWis:
I'm not really seeing the use case, just the request to be able to call the OnDeployExecutor from other code.  i.e. I see "what" is being requested (ability to re-trigger a script from code), but don't understand "why" it's being requested especially since the JMX console already allows re-running a script I believe AND a script can always be re-run by re-deploying the bundle (simply need to delete the status node if you are re-running a previously successful script).

OnDeployScripts were always kinda meant to be one-time things, but I imagine folks have come up with other use cases that I'm not thinking of.  Seeking to understand.

bundle/src/main/java/com/adobe/acs/commons/ondeploy/OnDeployExecutor.java
geertsmichael: Thank you. I agree with the suggested approach.

geertsmichael:
The use-case is about a tool that I wrote, which will visualize the script execution (success, running, fail). This tool allows to do a re-run through my Touch UI tool.

bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutorImpl.java
geertsmichael: I only expect output when completing the script (successful or failure).

HitmanInWis: Right, but if you dont clear the status out at the beginning of the run then this scenario exists:

- Script A fails and writes output stacktrace of "it broke"
- Script A sits with status "failure" and output "it broke" for awhile
- Code (or deployment) triggers Script A to run again, setting status back to "running"
- _Script A has a status of "running" and output of "it broke" until it finishes_
- Script A finishes successfully and clears out the output property

It's item 4 that I'm concerned with, since the old output remains while the new script instance is running.  Since there is only one status node, the status node should remain consistent with a single run - thus when it's running the "output" should be empty.

bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutorImpl.java
geertsmichael: Applied fix

content/pom.xml
geertsmichael: Restored versions

bundle/pom.xml
geertsmichael: Restored version

bundle/src/main/java/com/adobe/acs/commons/ondeploy/OnDeployExecutor.java
geertsmichael: Implemented like suggested. Please verify again

bundle/src/main/java/com/adobe/acs/commons/ondeploy/impl/OnDeployExecutorImpl.java
geertsmichael: That's indeed a valid use case that I missed.
I implemented the suggested approach. Please review.

CHANGELOG.md
```diff
@@ -32,7 +34,6 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 - #1897 - Fixed an NPE with removing a group w/ Ensure Authorizable when the group was already removed
 - #1934 - add explicit javax.annotation version to maven-bundle-plugin after #1893
 - #1202 - fix overflow handling in looping iterator
-- Adjust PageCompareDataImplTest to Java 11
```
HitmanInWis: Is this line removal intended?

bundle/src/main/java/com/adobe/acs/commons/ondeploy/OnDeployExecutor.java
HitmanInWis: 👍

HitmanInWis:
One quick question remains about `CHANGELOG.md` - other than that this looks good and I'll test it out later today.

CHANGELOG.md
geertsmichael: I've re-added this line

HitmanInWis:
Everything tests out, merged.  Thanks @michael-geerts for the contribution and @shsteimer for the recommended approach.

PR #1973, opened 2019-06-27
Adding vanity URL and URL rewrite support to SiteMap

leonbe02:
Summary of changes:
- Vanity URLs specified on a Page resource will be used over the actual path
- New property added that allows users to specify multiple URL rewrites that change `<loc>`'s to match basic apache URL rewrites used on a dispatcher. Format used: `<path>:<replacement>` i.e. `content/mysite/en_US/:mysite/`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 - #1806 - Http Cache: Added RequestPath extension
 - #1825 - Added sql2scorer JSON servlet to provide oak:scoreExplanation details for JCR-SQL2 queries.
 - #1899 - Added page inheritance respected in Named Transform Image Servlet for cq:Page
+- #1973 - Added Vanity URL support and the ability to specify URL rewrites so the output matches dispatcher
 
 ### Changed
 - #1539 - Removed unused references to the QueryBuilder API.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 - 1927 - HTTP cache: added cookie exclusion
 - 1905 - HTTP cache: added response header exclusion
 
+### Changed
+- #1945 - Added support for jcr:content creation and update to the Data Importer
+
 ## [4.2.0] - 2019-06-18
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 - #1806 - Http Cache: Added RequestPath extension
 - #1825 - Added sql2scorer JSON servlet to provide oak:scoreExplanation details for JCR-SQL2 queries.
 - #1899 - Added page inheritance respected in Named Transform Image Servlet for cq:Page
-- #1973 - Added Vanity URL support and the ability to specify URL rewrites so the output matches dispatcher
+- #1973 - Added Vanity URL support to SiteMap and the ability to specify URL rewrites so the output matches dispatcher
 
 ### Changed
 - #1539 - Removed unused references to the QueryBuilder API.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 - #1806 - Http Cache: Added RequestPath extension
 - #1825 - Added sql2scorer JSON servlet to provide oak:scoreExplanation details for JCR-SQL2 queries.
 - #1899 - Added page inheritance respected in Named Transform Image Servlet for cq:Page
+- #1973 - Added Vanity URL support and the ability to specify URL rewrites so the output matches dispatcher
```
davidjgonzalez:
```suggestion
- #1973 - Added Vanity URL support to SiteMap and the ability to specify URL rewrites so the output matches dispatcher
```

leonbe02:
@davidjgonzalez Is there anything else I need to change for approval?

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -233,6 +239,22 @@ protected void doGet(SlingHttpServletRequest request, SlingHttpServletResponse r
         return allAssetFolders;
     }
 
+    private String applyUrlRewrites(String url) {
+        for (String rewrite : urlRewrites) {
```
davidjgonzalez: it might be more efficient (or atleast clear) if, in activate(..) you parse the params out into a Map using [1]

And in this you just check the parsed map values?

[1] https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/util/ParameterUtil.java#L102

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -233,6 +239,22 @@ protected void doGet(SlingHttpServletRequest request, SlingHttpServletResponse r
...
+            if (!rewrite.contains(":") || rewrite.startsWith(":")) {
+                continue;
+            }
+            String[] tokens = rewrite.split(":");
+            String path = tokens[0];
+            String replacement = tokens.length > 1 ? tokens[1] : "";
+            if (url.contains(path)) {
```
davidjgonzalez: Should this be "contains" or "startsWith" ?  .. It might be an outlier case, but couldn't `contains` result in false positives?

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
leonbe02: Addressed in the latest commit

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
leonbe02: Addressed in the latest commit

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -240,16 +243,10 @@ protected void doGet(SlingHttpServletRequest request, SlingHttpServletResponse r
...
-            if (!rewrite.contains(":") || rewrite.startsWith(":")) {
-                continue;
-            }
-            String[] tokens = rewrite.split(":");
-            String path = tokens[0];
-            String replacement = tokens.length > 1 ? tokens[1] : "";
-            if (url.contains(path)) {
-                url = url.replaceFirst(path, replacement);
-                break;
+        String path = URI.create(url).getPath();
```
davidjgonzalez: @leonbe02 LGTM - just want to double check that url will always be a valid param for `URI.create(..)`? I'd think if it isnt, and an IllegalArgumentException is thrown, it would be best just to use the param's original value?

leonbe02: Not sure if the externalizer guarantees a valid URL so I just wrapped it in a try/catch

Issue #1914, opened 2019-05-28
java.lang.ClassNotFoundException: com.fasterxml.jackson.core.JsonProcessingException not found by com.adobe.acs.acs-aem-commons-bundle


suryakand:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.5
* [ ] ACS AEM Commons Version: 4.1.0
* [ ] Reproducible on Latest? DID NOT TRY

### Expected Behavior

The following exception should not be thrown on AEM 6.5:
`java.lang.ClassNotFoundException: com.fasterxml.jackson.core.JsonProcessingException not found by com.adobe.acs.acs-aem-commons-bundle [593]`

_Please describe briefly the expected behavior, i.e. when I do X, Y should happen._

### Actual Behavior
While using Adobe I/O integration from acs commons, getting following exception:

`java.lang.ClassNotFoundException: com.fasterxml.jackson.core.JsonProcessingException not found by com.adobe.acs.acs-aem-commons-bundle [593]`

Looks like the AEM 6.5 comes with Jackson 2.9.5, but acs commons is still pointing to 2.8.4

_Please describe briefly the actual behavior, i.e. when I do X, Z happens._

### Steps to Reproduce
Install ACS commons on AEM 6.5 and try to invoke Adobe I/O integration (e.g. Analytics)

### Links

ACS Commons Adobe I/O integration guide: https://adobe-consulting-services.github.io/acs-aem-commons/features/adobe-io-apis/index.html

avinashtalamanchi:
- We encountered same issue while using `acs-aem-commons 4.1.0`, found that `acs-aem-commons` bundle is not importing `com.fasterxml.jackson*.` which is why bundle is not finding `com.fasterxml.jackson*.`


- _I have created a Pull request https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/2007 for the above issue as a permanent fix._

---------------------------------------------------------------------------------------------

- **FYI** -  As a temporary solution, we added missing imports to`acs-aem-commons budle` by 
creating `bundle-fragment` to our project. Below is MANIFEST.MF

```
Manifest-Version: 1.0
Bundle-SymbolicName: com.XXXXXX.XXXX.jackson-core-fragment
Bundle-Name: XXXXX Jackson-core fragment for ACS Commons
Bundle-Version: 1.97.0.201907292300
Built-By: XXXXXXXXXX
[...]

PR #1788, opened 2019-03-14
shade embedded libraries. fixes #1786

justinedelson:
This is a potential fix for the issue reported in #1786. With this change, rather than simply embeddeding guava, jsch, etc. we shade (relocate) the embedded packages.

This will need some testing before merging.

PR #2007, opened 2019-07-29
BugFix #1914 : Fix to com.fasterxml.jackson* ClassNotFoundException

avinashtalamanchi:
Issue : https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/1914

- When verified we noticed that`acs-aem-commons` bundle is not importing `com.fasterxml.jackson*`. which is why bundle is not finding `com.fasterxml.jackson*`. 

- As fix, we added all missing `<Import-Package>` to the `maven-bundle-plugin` configuration.


davidjgonzalez:
What version of AEM was this effecting? Aren’t these exported by all supported versions of AEM to support sling mode exporter? 

Of these are optional what features are impacted and when we’re they introduced?

It seems strange that this is only popping up now.

adamcin:
@davidjgonzalez I believe this was discovered when installed on AEM 6.5, but the issue isn't that the exports aren't there, but that the imports weren't added in the acs-aem-commons-bundle manifest. I actually think this might have been introduced by #1788 (v4.1.0). By changing the means of embedding `jjwt` from the maven-bundle-plugin to the maven-shade-plugin, we lost the implicit import-package processing for the embed step that was previously identifying the following java package imports in `io.jsonwebtoken.impl.DefaultJwtBuilder`: 

```
import com.fasterxml.jackson.core.JsonProcessingException;
import com.fasterxml.jackson.databind.ObjectMapper;
```

davidjgonzalez:
@adamcin thanks for the explanation! Makes more sense now!

Issue #2022, opened 2019-08-20
Export CSV functionality is not taking the custom executor defined


highsoftx980:
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: AEM 6.1, AEM 6.2, AEM 6.3, AEM 6.3 
* [x] ACS AEM Commons Version: 2.14.x, 3.19.x, 4.x.x
* [x] Reproducible on Latest? yes

### Expected Behavior

According to documentation on Extending Reports it should be possible to create a custom Report Executor, use the CSV export option without errors and get the CSV generated.

### Actual Behavior

> Caused by: com.adobe.acs.commons.reports.api.ReportException: Failed to get report executor

### Steps to Reproduce

Follow the guide for extending the report builder.

Also, I created a component which uses TestReportExecutor
1-  Create a TestReportExecutor.java class with @Model annotation and implemented ReportExecutor interface.
2- Configure it in the report config dialog:
`<reportExecutor
                            jcr:primaryType="nt:unstructured"
                            sling:resourceType="granite/ui/components/foundation/form/hidden"
                            name="./reportExecutor"
                            value="com.co.test.TestReportExecutor"/>
`
3- Configure the rest of the report.
4- Open the report and execute it -> when it's working.
5- Click 'Download Report'.
6- Error popup is shown and check the logs looking for the error:
> Caused by: com.adobe.acs.commons.reports.api.ReportException: Failed to get report executor

### Links

Links
Documentation on Extending Reports
https://adobe-consulting-services.github.io/acs-aem-commons/features/report-builder/extending.html#creating-a-report-executor
Check the 'Creating a Report Column' section

PR #2023, opened 2019-08-20
#2022 - Adding logic for getting the custom report executor

highsoftx980:
#2022 - Adding logic for getting the custom report executor for exporting the reports CSV file (option -> Download Report)


klcodanr:
@davidjgonzalez - could you merge this? It does have a negative impact on test coverage, but this code does not have coverage to begin so I don't think that's reasonable to expect to add at this point.

PR #2032, opened 2019-09-01
Fixing /var/acs-commons filter.xml

geertsmichael:
Fixing /var/acs-commons filter.xml (removed double rep:policy, moved /var/workflow filter, added on-deploy-scripts-status)


ui.content/src/main/content/META-INF/vault/filter.xml
```diff
@@ -95,7 +95,7 @@
             pattern="/etc/acs-commons/reports(.*)?"
             mode="merge"/>
         <include pattern="/var/acs-commons/httpcache"/>
-        <include pattern="/var/acs-commons/rep:policy"/>
-        <include pattern="/var/workflow/instances/rep:policy"/>
+        <include pattern="/var/acs-commons/on-deploy-scripts-status"/>
     </filter>
+    <include pattern="/var/workflow/instances/rep:policy"/>
```
kwin: This must be a dedicated filter rule with only root. No include/exclude is necessary

geertsmichael: Fixed

geertsmichael:
Could you merge this Pull Request please? Thank you.

ui.content/src/main/content/META-INF/vault/filter.xml
```diff
@@ -95,7 +95,7 @@
             pattern="/etc/acs-commons/reports(.*)?"
             mode="merge"/>
         <include pattern="/var/acs-commons/httpcache"/>
```
kwin: /var/acs-commons/mcp is missing as well, right?

ui.content/src/main/content/META-INF/vault/filter.xml
```diff
@@ -95,7 +95,7 @@
             pattern="/etc/acs-commons/reports(.*)?"
             mode="merge"/>
```
kwin: also mode=merge is only supported on filter elements,  I think line 94ff can be removed

HitmanInWis:
FWIW I believe the issues here were caused by commit 65fafaaea151ef30819a83f7660d75c4bc95f663 back in June - this breaks On-Deploy Scripts for new installs, so thanks for this PR.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 
 ## [Unreleased]
 
+- #2032 - Fixed filter.xml on /var/acs-commons
```
kwin: The idea is that you reference a real issue report here, not the actual PR id. AFAIK there is no issue report yet for this incorrect filter, right?

kwin:
I came up with a new PR for this in #2045 . @geertsmichael Could you cross-check?

badvision:
Should I close this PR in favor of #2045?  Thanks to all of you folks working on this problem, by the way!

HitmanInWis:
Let's hold off for just a bit.  Seems #2045 is having some issues currently (I'll take a look tomorrow) - want to confirm that one is good (it covers this and other changes, but the other changes are what is currently holding it up) before closing this.

badvision:
Fantastic.  Happy to wait, just didn't want you fine folks thinking I was ignoring you. :)

ui.content/src/main/content/META-INF/vault/filter.xml
HitmanInWis: This will be added by #2045

ui.content/src/main/content/META-INF/vault/filter.xml
HitmanInWis: Unrelated to this card.

CHANGELOG.md
HitmanInWis: Unless there's a hard and fast rule about this, I dont think this is a big deal.  Pretty sure this has been done before.

Issue #1970, opened 2019-06-24
Rich text controls in touch UI multifield fail to populate


awferg:
## Required Information

* AEM Version 6.5.0
* ACS AEM Commons Version: 4.0.0
* Reproducible on Latest? yes

### Expected Behavior

Rich Text controls in Touch UI Multifield should be populated. 

### Actual Behavior

Rich texts controls are not populated.

### Steps to Reproduce

Using a dialog configuration such as this:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<jcr:root xmlns:sling="http://sling.apache.org/jcr/sling/1.0" 
  xmlns:cq="http://www.day.com/jcr/cq/1.0" xmlns:jcr="http://www.jcp.org/jcr/1.0" xmlns:nt="http://www.jcp.org/jcr/nt/1.0"
  jcr:primaryType="nt:unstructured"
  jcr:title="Accordion"
  sling:resourceType="cq/gui/components/authoring/dialog">
  <content
    jcr:primaryType="nt:unstructured"
    sling:resourceType="granite/ui/components/foundation/container">
    <items jcr:primaryType="nt:unstructured">
      <tabs
        jcr:primaryType="nt:unstructured"
        sling:resourceType="granite/ui/components/foundation/container">
        <layout
          jcr:primaryType="nt:unstructured"
          sling:resourceType="granite/ui/components/foundation/layouts/tabs"
          type="nav"/>
        <items jcr:primaryType="nt:unstructured">
          <tab1
            jcr:primaryType="nt:unstructured"
            jcr:title="Accordion Configuration"
            sling:resourceType="granite/ui/components/coral/foundation/fixedcolumns">
            <items jcr:primaryType="nt:unstructured">
              <column
                jcr:primaryType="nt:unstructured"
                sling:resourceType="granite/ui/components/foundation/container">
                <items jcr:primaryType="nt:unstructured">
                  <accordionData
                    jcr:primaryType="nt:unstructured"
                    sling:resourceType="granite/ui/components/coral/foundation/form/multifield"
                    fieldLabel="Accordion Data"
                    maxItems="8">
                    <field
                      jcr:primaryType="nt:unstructured"
                      sling:resourceType="granite/ui/components/foundation/form/fieldset"
                      acs-commons-nested=""
                      name="./accordionDataTouch">
                      <items jcr:primaryType="nt:unstructured">
                        <accordionHeader
[...]

PR #2050, opened 2019-09-19
Update touchui-composite-multifield.js

SylvesterAbreu:
This is to address the issue reported here https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/1970. Credit to https://github.com/awferg


badvision:
Why not remove the previous line?  Seems we're just setting the same variable twice in a row.  Also the previous line of code is less specific (doesn't indicate 'input' in the query selector).  Is the fact it is being too specific why the bug is occurring?

ui.apps/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/composite-multifield/source/touchui-composite-multifield.js
```diff
@@ -189,6 +189,10 @@
 
                         _.each(record, function (rValue, rKey) {
                             $field = $($fieldSets[i]).find("[name='./" + rKey + "']").last();
+																												
+																												if ($field.hasClass("coral-RichText-editable")) {
+	                               $field = $($fieldSets[i]).find("input[name='./" + rKey + "']").first();
```
davidjgonzalez: Is this not overwriting the value set on line 191?  What is the rationale for the first vs the last?

@schoudry is this a safe change?

schoudry: i think its not overwriting... indentation is so poor,` if ($field.hasClass("coral-RichText-editable")) {` is not visible

davidjgonzalez: omg....

ui.apps/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/composite-multifield/source/touchui-composite-multifield.js
```diff
@@ -189,6 +189,10 @@
 
                         _.each(record, function (rValue, rKey) {
                             $field = $($fieldSets[i]).find("[name='./" + rKey + "']").last();
+																												
+																												if ($field.hasClass("coral-RichText-editable")) {
```
davidjgonzalez:
```suggestion																										if ($field.hasClass("coral-RichText-editable")) {
```

PR #2064, opened 2019-10-01
Adding a Marketo Form Component

klcodanr:
Adding a simple component to add a Marketo Form embed onto any AEM page. Includes some basic customization features such as being able to set the redirect URL, set field values from various sources, adding hidden fields and a custom script hook.


klcodanr:
@davidjgonzalez @badvision -- any thoughts here?

shsteimer:
@klcodanr I'm curious if you've looked at the core embed component https://opensource.adobe.com/aem-core-wcm-components/library/embed.html and thought about using that as the basis for this (either as an OOB available embeddable in core components, or in some form as an acs commons extension of that core component) rather than a standalone component?  Just food for thought.

klcodanr:
@shsteimer I had not, I wasn't aware of the embed feature, which makes sense as it's pretty new. 

I've looked at it and while it does seem like it would also work, it also seems clunky from a user interface perspective. My intent is to make a component that is very easy to author. Adding in the step of adding an embed, then the user selecting embedded and then actually configuring the form seems unnecessary. 

I'm not _opposed_ but I'd be concerned that it's making a worse user experience for the sake of development purity. That being said if the intent is to move all "embed like" things into this embed component, well I don't want this to be the odd one out. If that is the case though, I'd suggest some refactoring to the embed component to make it easier to author.

shsteimer:
That makes sense, mostly wanted to make you aware of it for consideration
because it is relatively new.

I don’t personally think all embed like things should/must use the embed
component.  I do however feel there is value to using core components as
much as possible for consistency & maintainability reasons.  How that
squares with whether this particular thing belongs in acs commons is
probably not for me to judge.

One way or the other, I like having an easily available option for a
Marketo form embed, so it’s a great contribution, imho!

klcodanr:
Thanks @shsteimer! Whom would be the person to make that decision? It'd be a fairly minor effort to refactor this to be a core component instead. @gabrielwalt ?

klcodanr:
As an update, after talking with Gabriel Walt & Cedric Huesler, I submitted this to Core Components, but they feel it's out of scope for the Core Components project. As @shsteimer indicated I did end up making this a sub-component of the embed component.

ui.apps/src/main/content/jcr_root/apps/acs-commons/content/cloudservice/create/.content.xml
```diff
@@ -0,0 +1,161 @@
...
+        </head>
+        <body
+            jcr:primaryType="nt:unstructured"
+            sling:resourceType="granite/ui/components/coral/foundation/page/body">
+            <items jcr:primaryType="nt:unstructured">
+                <form
+                    granite:id="acs-commons-cloudservice-create"
+                    jcr:primaryType="nt:unstructured"
+                    sling:resourceType="granite/ui/components/coral/foundation/form"
+                    action="/bin/acscommons/createcloudconfig"
```
davidjgonzalez: @klcodanr is there any natural place you could expose a resource with a resourceType that this servlet could be bound to, rather than mounting it at the path `/bin/...`?

klcodanr:
Thanks @davidjgonzalez - I attached the servlet to the cloudconfiglist component so it's no longer attached to /bin.

Issue #2067, opened 2019-10-04
Rewrite workflow remover as MCP task


badvision:
### Proposal

In order to remove extra weight from the package, it makes sense to convert lesser-used utilities to run from the MCP console.   This requires less overall maintenance of the UI (especially in lieu of UI changes in the product over time), as well as provide a more consistent way to provide monitoring and error reporting from these utilities.

davidjgonzalez:
This is a good idea. I created a workflow model data source for the Bulk Workflow process you probably could leverage.

Keep in mind Workflow remover has a scheduled service component to it.

badvision:
I think we're going to keep most of it as-is (including the service component) and just focus on the MCP-based front end in this release.  A more thorough solution would to let MCP manage more of the removal queue and so on, but that's a bit extra.

PR #2068, opened 2019-10-04
Feature/update wf remover mcp

badvision:
Implementation for #2067 for review


shsteimer:
@badvision there is one remaining code climate issue around throwing generic `Exception` but I checked other processes and those seem to do the same, so my assumption is that is OK?  

I have 2 empty test classes I'm still working on, but functionally everything is there and appears to me to be working.  Will push the updated tests when they are ready, will leave it up to you if you want to merge this now and merge the tests in a separate PR or wait for the tests and merge it as a single PR, given that enough without those tests there is no decrease in code coverage.

Issue #2071, opened 2019-10-08
Export tags as CSV file compliant with Tag Maker parsers


Gatar:
There is useful tool in ACS AEM Tools which gives possibility to import tags into AEM - Tag Maker, but there are no possibility to export tag from AEM to CSV file. 
New functionality should be created, which gives possibility to export tags, both localized and non-localized. Functionality should be created as new option of ACS AEM Commons Exporter to provide consistency and good user experience.

Gatar:
@klcodanr could You please look at https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io/pull/175 ?

Gatar:
@joerghoh could You please look at Adobe-Consulting-Services/adobe-consulting-services.github.io#175 ?

PR #2072, opened 2019-10-08
Tag Export functionality as new option in Exports #2071

Gatar:


klcodanr:
@Gatar how would you compare this with the functionality of the Tag Report MCP? https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/TagReporter.java

It seems like similar functionality to me.

Gatar:
@klcodanr those functionalities are similar, but not same. Tag Report MCP is used to create customizable reports about tags. Tag Export is one-click tool to export tag structure as CSV in format, where You can directly import those tags back to AEM by Tag Maker.

klcodanr:
Thanks @Gatar, so follow-up question then, should this be in the MCP framework vs Exporters? Given Adobe's move away from /etc and general consolidation on the MCP framework, I would think this makes more sense.

Gatar:
@klcodanr I think that this is good idea as next step for Tag Export tool: move tool to MCP framework and support MCP Tag Creator formats (xlsx files). In this moment Tag Export exports only CSV in format supported by Tag Maker from ACS Tools, but it is in ready to use state including tests and uses in living project.

Gatar:
@klcodanr are You agree with my approach?

klcodanr:
@Gatar How much effort would it be to convert to a MCP? If that's the intended direction, I would recommend going in that approach directly rather than introducing a feature and then deprecating it almost immediately.

badvision:
MCP wouldn't be difficult but the output format of MCP is rather limited to what the Generic Report can do.  If the output format isn't terribly complicated then MCP is very possible, and Generic Report can provide the XLSX output.  If we also need CSV output that is very easy to implement as a bonus MCP feature.

Gatar:
Moving this functionality to be a part of MCP means, generally, that in need to be written from scratch, cause of:  
- change UI
- align with MCP API
- support all converters from Tag Creator functionality (they are not same like those from Tag Maker)
- tests check Tag Maker tags schema, so they need to be rewritten

ACS AEM Tools - Tag Maker is not deprecated. Is it going to be marked as deprecated in nearest future? In my opinion it is better, in this moment, to had exporter compatible at least with Tag Maker than none.

klcodanr:
Thanks for the reminder @Gatar - I think we're good. I just approved and merged.

PR #2073, opened 2019-10-10
fixes #1966 additional ignore properties are ignored

### LINKED ISSUE #1966: Oak Ensure index doesn't use properties.ignore on AEM restart
## Required Information

* [ ] AEM Version: 6.4.4.0
* [ ] ACS AEM Commons Version: 3.19.0
* [ ] Reproducible on Latest? untested

### Expected Behavior

When ignore.properties is set on com.adobe.acs.commons.oak.impl.EnsureOakIndexManagerImpl OSGi config, this should be used when ensuring the indexes.

### Actual Behavior

All extra ignore.properties are skipped when ensuring the indexes on AEM restart.

### Steps to Reproduce

- Create a Lucene type index and don't include the 'seed ' value in the Oak Ensure Index definition, but do include 'forceReindex={Boolean}true'
- Create an OSGi config for com.adobe.acs.commons.oak.impl.EnsureOakIndexManagerImpl and set ignore.properties to '[seed]'.
- Restart AEM
- Watch your index being re-indexed (normal because the index doesn't exist yet)
- Restart AEM
- Watch your index being re-indexed again (not normal because 'seed' should be ignored)

### Extra info

When enabling debug logging on com.adobe.acs.commons.analysis.jcrchecksum you see that the 'seed' property is used and not excluded for calculating the checksum, thus causing a difference to be detected and the re-index being executed.

As of 4.0.0 the issue for 'seed' will not be visible because it is added to the defaults.

### Work-around

Include the 'seed' property in your Oak Ensure Index definition

### Plausible cause

Most likely is that the activate method in EnsureOakIndexManagerImpl reading the property isn't called yet before the AppliableEnsureOakIndex is bound to the service.

wimsymons:
Fix for #1966

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 - #2065 - Upgraded oakpal to 1.5.1; use expectPaths and expectAces checks to verify rep:policy nodes instead of inlineScript
 
 ### Fixed
+- #1966 - Fixed properties.ignore not being honored on Ensure Oak Index
 - #2032 - Fixed filter.xml on /var/acs-commons
 - #2048 - Fixed ui.apps ACE import by setting acHandling property to "merge"
 - #2053 - ETag header not properly delivered from the servlet (missing quotes)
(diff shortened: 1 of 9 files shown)


wimsymons:
Updated with a CHANGELOG.md entry.

wimsymons:
Needed to update com.adobe.acs.commons.oak OSGi version to 2.0.0.

wimsymons:
Please read https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/1966#issuecomment-540587569

wimsymons:
Fixed codeclimate error.

wimsymons:
We should go with @joerghoh 's solution: https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/2074

Issue #2019, opened 2019-08-13
WorkflowPackageManagerImpl requires that the user creating workflow package has read access to /var/workflow/packages


gicig:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.5.1
* [ ] ACS AEM Commons Version: 4.2.0
* [ ] Reproducible on Latest? yes

### Expected Behavior

If user creating workflow package has read&write permissions for `/var/workflow/packages/project1/tenant1/`, it should be able to create packages there, even though it doesn't have any permissions for `/var/workflow/packages`

### Actual Behavior

The user with permissions described above can not create packages for `/var/workflow/packages/project1/tenant1/`, unless it is granted read permissions for `/var/workflow/packages/` - check [here](https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/workflow/impl/WorkflowPackageManagerImpl.java#L265)

### Steps to Reproduce

Create user with permissions above and try to create a workflow package using [WorkflowPackageManagerImpl#create(ResourceResolver , String, String, String...)](https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/workflow/impl/WorkflowPackageManagerImpl.java#L110)

### Links

https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/workflow/impl/WorkflowPackageManagerImpl.java

joerghoh:
Seems to be caused by https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/workflow/impl/WorkflowPackageManagerImpl.java#L265

The detection if the "new" or the "legacy" path is used could be done on startup of the service; of course then you need to restart the service/instance when you migrate the workflows. But if that happens on the startup, you could use a session with elevated privileges to detect that and then the user does not have to have read access to /var/workflow/packages anymore.

gicig:
Would a

joerghoh:
@gicig Can you review the bugfix I provided in #2077 and check if it is working for you?

davidjgonzalez:
Fixed via #2078

gicig:
[...]

PR #2077, opened 2019-10-18
avoid problems when the user does not have read permissions on (/etc|var)/workflow/packags

joerghoh:
Adressing #2019 

I had to introduce a service user for this purpose, and to avoid the runtime penalty for it I extracted it to the activate method.

Thus the change of the workflow base directory (which happens exactly once) is not reflected immediately, but it requires at least a restart of the service/bundle or the complete AEM instance.


joerghoh:
anyone interested in reviewing?

PR #2085, opened 2019-10-31
updated to make dry run mode on all MCP processes that have it checke…

gutsy:
…d by default, to help prevent accidental runs of very powerful processes. Also updated the tests to turn dry run mode off so that they still succeed

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/cfi/ContentFragmentImport.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/cfi/ContentFragmentImport.java
@@ -90,9 +90,10 @@ public class ContentFragmentImport extends ProcessDefinition {
 
     @FormField(
             name = "Dry run",
-            component = CheckboxComponent.class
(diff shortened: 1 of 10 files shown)


bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/cfi/ContentFragmentImport.java
```diff
@@ -90,9 +90,10 @@
 
     @FormField(
             name = "Dry run",
-            component = CheckboxComponent.class
+            component = CheckboxComponent.class,
+            options = "checked"
     )
-    transient boolean dryRunMode;
+    transient boolean dryRunMode = true;
```
davidjgonzalez: Why is this instance transient?

gutsy: Because it was already that way, I was just setting the bool to true and didn't know if the original author had it transient for a reason.

PR #2086, opened 2019-11-01
improve ResourceResolver handling (fixes #1864)

joerghoh:
* use try-with-resource, 
* remove unnecessary references to the ResourceResolverFactory
* add some todos for further improvement


joerghoh:
I raised #2094 to cleanup the RemoteAssets feature further on; there I plan to remove some code and increase test coverage again.

bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsBinarySyncImpl.java
```diff
@@ -155,11 +155,6 @@ private void flagAssetAsFailedSync(ResourceResolver remoteAssetsResolver, Resour
             remoteAssetsResolver.commit();
         } catch (Exception e) {
             LOG.error("Error flagging remote asset '{}' as failed - asset may attempt to sync numerous times in succession", resource.getPath(), e);
-            try {
-                remoteAssetsResolver.revert();
```
davidjgonzalez: @joerghoh is this `.revert()` handled elsewhere now?

joerghoh: this method is called from syncAsset only; and immediately after this call to flagAssetAsFailedSync() the resolver is closed.
No need to revert pending changes when the resolver is closed immediately after anyway.

PR #2096, opened 2019-11-07
Update .content.xml

YentlFrickx:
Add wrapper client library definition in comments


klcodanr:
@YentlFrickx why is this added as a comment in the code instead of on the documentation site: https://adobe-consulting-services.github.io/acs-aem-commons/#features

Unless there's a good reason to do so, I would recommend closing this PR and updating the doc site instead: https://github.com/adobe-Consulting-Services/adobe-consulting-services.github.io

davidjgonzalez:
Thanks!

Issue #2087, opened 2019-11-03
Automated Package Replicator Touch UI Support


JiangLong2019:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc:  All versions
* [ ] ACS AEM Commons Version: All versions
* [ ] Reproducible on Latest? yes

### Expected Behavior

Automated Package Replicator does not support touch ui, convert it to touch ui to make it more used in the future.

### Actual Behavior

Classic ui base

### Steps to Reproduce
Create a configuration page under "/miscadmin#/etc/acs-commons/automatic-package-replication", classic ui shows when open it

klcodanr:
@JiangLong2019 -- Based on the repository restructuring here: https://helpx.adobe.com/experience-manager/6-5/sites/deploying/using/all-repository-restructuring-in-aem-6-4.html

I'd expect that the path should be /conf/global/settings/acs-commons/automated-package-replicator

JiangLong2019:
Thanks @klcodanr, I'll plan to refactor it.

PR #2097, opened 2019-11-07
Feature/evaluate automated package replicator touchui #2087

JiangLong2019:


klcodanr:
@JiangLong2019 Is there any reason to keep the angular front end if you're moving this out of /etc? The whole reason that was necessary was because of the strange limitations Adobe put on /etc.

JiangLong2019:
Hi @klcodanr, since I couldn't find any front development guidelines in ACS commons, I just kept similar function with Report Builder. Any suggestions on this?

klcodanr:
@JiangLong2019 - A few things seem off here:

- This PR does not change the location of the configurations from /etc
- There are some missed updates from the reports UI (+ Report)
- You can't edit the configurations
- The benefit of switching to TouchUI is that we could remove the custom Angular code to edit the configurations, this has not been done

APR

JiangLong2019:
@klcodanr Thanks for the feedback! Let me add some comments inline.

>This PR does not change the location of the configurations from /etc
This was my concern while developing, should I change it to /conf?

>There are some missed updates from the reports UI (+ Report)
>You can't edit the configurations
I couldn't reproduce. Anyway, I plan to refactor it as you suggested we should go for TouchUI.

>The benefit of switching to TouchUI is that we could remove the custom Angular code to edit the configurations, this has not been done
OK, got it.

PR #2101, opened 2019-11-08
#2019 cleanup public API of remoteAssets feature

joerghoh:


joerghoh:
(rebased branch on master)

Issue #2103, opened 2019-11-11
Update Mocking Libraries to work with Java11


joerghoh:
The naive approach just to switch the JVM from 8 to 11 is not working well. We have a few failing test cases which point to issues with mocking libraries (especially powermock). But upgrading powermock alone does not help, but we should upgrade mockito as well (and we are on a pretty old version right now).

badvision:
It is possible to switch to the newer mockito and, with the right annotations at the top to ignore some problematic java classes, have the static mocks work again.  Ultimately, though, we might do better in the long run to reduce our dependency on mockito for this, as the jigsaw changes from Java 9 and onward make it much harder for stuff like Mockito to work correctly.  Instead, we should refactor the code where possible to ensure unit tests don't have to do backflips.

badvision:
So one example is using static methods like JcrUtils which do not work well in unit tests.  If we refactor those into facade services then we can more easily inject a mock service implementation that we can use in the unit tests.

joerghoh:
I am sure that in many cases we could switch to AemContext/SlingContext and get rid of cascades of mocks. That would be definitely some great contribution.

joerghoh:
#2104  has been merged.

PR #2104, opened 2019-11-11
[Looking for reviewers] jdk11 compatible unittests (adresses #2103 and 1916)

joerghoh:
this is an approach to update mockito and powermock, and fix the resulting test failures. Consider it as WIP right now.


joerghoh:
Anyway, I am stuck right now with this type of failure:

```
[INFO] Running com.adobe.acs.commons.remoteassets.impl.RemoteAssetsBinarySyncImplTest
[ERROR] Tests run: 5, Failures: 0, Errors: 5, Skipped: 0, Time elapsed: 1.689 s <<< FAILURE! - in com.adobe.acs.commons.remoteassets.impl.RemoteAssetsBinarySyncImplTest
[ERROR] testSycnAssetSuccessWithRemovalOfNotFoundRendition(com.adobe.acs.commons.remoteassets.impl.RemoteAssetsBinarySyncImplTest)  Time elapsed: 0.123 s  <<< ERROR!
org.apache.sling.testing.mock.osgi.NoScrMetadataException: No OSGi SCR metadata found for class org.mockito.codegen.RemoteAssetsConfigImpl
at com.adobe.acs.commons.remoteassets.impl.RemoteAssetsBinarySyncImplTest.setup(RemoteAssetsBinarySyncImplTest.java:117)
```

Problematic is here, that for some unknown reason the code assumes, that the classname is ```org.mockito.codegen.RemoteAssetsConfigImpl```, and of course there is no SCR metadata available for that.
(Interestingly, when I run this testcase in Eclipse, the classname is correct ...)

joerghoh:
Isolating the test executions seems to help (but creates more overhead). I'll do it for the moment to focus on the real issues. Reverting to the old testing behavior later on.

joerghoh:
The failures are consistent over Java8(Oracle and OpenJDK) and Java 11 (OpenJDK), so it's something with the Mockito update.

joerghoh:
I created #2120 to eliminate the mocks for the JCRHttpCacheStoreImpl. Let's wait for its merge and continue then here.

joerghoh:
With Java11 there seem to be some issues when using Sling Mocks and the default configuration of Surefire. I had to reconfigure it to use a single thread per testclass. The overhead of it should be minimal, but seems to provide consistent positive test results.

joerghoh:
Hm, coveralls is waiting. Is anyone able to see what's going wrong here?

badvision:
I've seen this before where coveralls just dies and isn't as easy to restart compared to travis.  No matter, I can still merge it anyway.

badvision:
Oh wait, it's a draft PR.  Not going to merge.  Didn't see that.  😊

.travis.yml
```diff
@@ -1,6 +1,8 @@
 language: java
 dist: trusty
-jdk: oraclejdk8
+jdk:
```
badvision: Specifying both JDKs means that we will run all tests for all languages in both runtimes; which also means doubling the number of tests.  Is there a way similar to how Asset Share Commons is defined such that we just run one test in JDK11 or, since that's the direction, run all but one test in JDK11 and leave one test on JDK8 for backwards compatibility assurance?

bundle/src/test/java/com/adobe/acs/commons/errorpagehandler/cache/impl/PowerMockErrorPageCacheImplTest.java
```diff
@@ -39,8 +40,13 @@
  *
  * Split out as ErrorPageCacheImplTest was having problems with @Spy'ied vars
  */
+
+// See https://stackoverflow.com/questions/52966897/powermock-java-11 for details about the 
+// @PowerMockIgnore statement
+
 @RunWith(PowerMockRunner.class)
 @PrepareForTest(ResourceDataUtil.class)
+@PowerMockIgnore({"com.sun.org.apache.xerces.*", "javax.xml.*", "org.xml.*", "org.w3c.*"})
```
badvision: Yes this change works for sure, I've had to do this in other projects as well.  Ultimately however we might try to find alternatives to mocking static methods altogether where possible.

bundle/src/test/java/com/adobe/acs/commons/genericlists/impl/GenericListAdapterFactoryTest.java
```diff
@@ -130,11 +130,6 @@ protected GenericList adaptToGenericList(Page page) {
     @Test
     public void test_that_adapting_page_with_null_template_returns_null() {
         Page wrongPage = mock(Page.class);
-        when(wrongPage.getProperties()).thenAnswer(new Answer<ValueMap>() {
```
badvision: TBH I'm not sure if this test is useful at all anymore; when I rewrote the generic list feature I implemented it as sling models.  Therefore, it's feasible that adaptTo would always return non-null as long as the resource itself isn't null.

bundle/src/test/java/com/adobe/acs/commons/httpcache/store/jcr/impl/writer/BucketNodeFactoryTest.java
```diff
@@ -46,12 +46,6 @@
     @Mock Session session;
     @Mock Node cacheRootNode;
 
-    @Before
-    public void setup() throws Exception{
-        when(session.isLive()).thenReturn(true);
-        when(cacheRootNode.getPath()).thenReturn(MockSettings.VALID_ROOT_PATH);
```
badvision: Removing this doesn't break the tests?  Really?  Interesting...

bundle/src/test/java/com/adobe/acs/commons/i18n/impl/I18nProviderImplTest.java
```diff
@@ -100,8 +100,6 @@ public void setUp() {
         i18nMap.put(I18N_KEY, TRANSLATED_FROM_ENGLISH);
 
         final Answer<String> answer = (Answer<String>) invocationOnMock -> i18nMap.get(invocationOnMock.getArguments()[1]);
-        doAnswer(answer).when(i18nProvider).translate(anyString(), any(HttpServletRequest.class));
```
badvision: Without a proper mock, do these methods fail elsewhere?

bundle/src/test/java/com/adobe/acs/commons/workflow/synthetic/impl/SyntheticGraniteWorkflowRunnerImplTest.java
```diff
@@ -168,7 +168,6 @@ public void testExecute_Restart() throws Exception {
 
     @Test
     public void testExecute_Terminate() throws Exception {
-        when(session.hasPendingChanges()).thenReturn(true).thenReturn(false);
```
badvision: @davidjgonzalez -- was there a reason you mocked it this way originally?

.travis.yml
joerghoh: Good idea. Let me figure it how to achieve it. This was just the fastest way to get test coverage with Java 11.

bundle/src/test/java/com/adobe/acs/commons/genericlists/impl/GenericListAdapterFactoryTest.java
joerghoh: Mockito 2 is quite picky and complains about unused mockings. So this means that ```wrongPage.getProperties()``` is not called in the test execution.

bundle/src/test/java/com/adobe/acs/commons/httpcache/store/jcr/impl/writer/BucketNodeFactoryTest.java
joerghoh: Mockito 2 is telling you alot about your tests ... :-)

joerghoh:
Thanks for the review. Indeed I was able to remove quite a number of stubbings in various test classes, because mockito 2 is pretty picky about them. The search term is ```UnnecessaryStubbingException```. 

And instead of disabling this feature I just removed the offending lines. And it's quite interesting that assumptions were made about stubs which are never used at all.

joerghoh:
fixed some final merge conflicts and reduced to only 1 Java 11 build.

davidjgonzalez:
@joerghoh is it ready for review?

joerghoh:
@davidjgonzalez It is. The rebasing brought in quite some code which is already on master :-/

joerghoh:
Thanks!

PR #2110, opened 2019-11-13
Adding File Fetcher 

klcodanr:
Adding a feature to support fetching files from a remote source, loading them into the AEM DAM and replicating the file to publish. 

**Justification:**

Some of our customers have an issue where Launch is being by Ad-Blockers but Adobe doesn't support CNAMEs for changing the domain name. Options for handling this could include setting up a Lambda webhook handler to a CNAME'ed bucket S3 bucket or allowing FTP uploads to a production server, but each comes with significant downsides in terms of complexity to maintain or generally just being a bad idea. 

**Concept**

The idea here is to configure AEM to pull down the Launch files on a regular schedule, cache it in the DAM and automatically replicate it when updated.

An administrator would configure the URL of the file to download, the path to place it in and the schedule to download the file in OSGi:

image

Since you could provide configurations per runmode this would support multiple Launch environments.


bundle/src/main/java/com/adobe/acs/commons/fetchfile/impl/FileFetchImpl.java
```diff
@@ -0,0 +1,229 @@
...
+  @Override
+  public void updateFile() throws IOException, ReplicationException {
+    log.trace("updateFile");
+
+    try (ResourceResolver resolver = factory
+        .getServiceResourceResolver(Collections.singletonMap(ResourceResolverFactory.SUBSERVICE, SERVICE_USER_NAME))) {
+      AssetManager manager = Optional.ofNullable(resolver.adaptTo(AssetManager.class))
+          .orElseThrow(() -> new PersistenceException("Failed to get Asset Manager"));
+
+      byte[] bin = fetchFile();
```
joerghoh: Can get problematic, if the fetcher is misused for large binaries. Is it possible to use the InputStream from the HTTP response and stream the output into the repo?

bundle/src/main/java/com/adobe/acs/commons/fetchfile/impl/FileFetchImpl.java
```diff
@@ -0,0 +1,229 @@
...
+            return new ImmutablePair<String, String>(h[0], value);
+          }).collect(Collectors.toList());
+      for (Pair<String, String> p : header) {
+        log.trace("Adding request property {}={}", p.getKey(), p.getValue());
+        con.addRequestProperty(p.getKey(), p.getValue());
+      }
+      con.setUseCaches(false);
+      int rc = con.getResponseCode();
+      log.debug("Retrieved response code: {}", rc);
+      if (ArrayUtils.contains(config.validResponseCodes(), rc)) {
```
joerghoh: Wouldn't it make sense to check the statuscode or even send an "if-modified-since" header to avoid downloading the file alltogether if it hasn't changed?

bundle/src/main/java/com/adobe/acs/commons/fetchfile/impl/FileFetchImpl.java
```diff
@@ -0,0 +1,229 @@
...
+    return new String(Base64.encode(DigestUtils.getMd5Digest().digest(data)));
+  }
+
+  @Override
+  public boolean isLastJobSucceeded() {
+    return this.lastJobSucceeded;
+  }
+
+  protected HttpURLConnection openConnection() throws IOException {
+    return (HttpURLConnection) new URL(config.remoteUrl()).openConnection();
```
joerghoh: We should add timeout settings here.

bundle/src/main/java/com/adobe/acs/commons/fetchfile/impl/FileFetchImpl.java
```diff
@@ -0,0 +1,229 @@
...
+      lastJobSucceeded = true;
+      lastException = null;
+    } catch (Exception e) {
+      log.warn("Failed to run fetch file job for {}", config.remoteUrl(), e);
+      lastException = e;
+      lastJobSucceeded = false;
+    }
+  }
+
+  public void setFactory(ResourceResolverFactory factory) {
```
joerghoh: is not required, as this is the default.

bundle/src/main/java/com/adobe/acs/commons/fetchfile/impl/FileFetchImpl.java
```diff
@@ -0,0 +1,229 @@
...
+    this.factory = factory;
+  }
+
+  public void setReplicator(Replicator replicator) {
```
joerghoh: is not required, as this is the default.

bundle/src/main/java/com/adobe/acs/commons/fetchfile/impl/FileFetchImpl.java
```diff
@@ -0,0 +1,229 @@
...
+        log.info("Adding new file...");
+        asset = manager.createAsset(path, new ByteArrayInputStream(bin), config.mimeType(), true);
+        assetResource = Optional.ofNullable(asset.adaptTo(Resource.class))
+            .orElseThrow(() -> new PersistenceException("Failed to adapt Asset to Resource"));
+        updateHash(fileHash, assetResource);
+        publish = true;
+      }
+
+      if (publish) {
+        log.info("Replicating...");
```
joerghoh: can you add the path of the asset to the log message?

joerghoh:
A more general comment: I am not really sure if providing the URL via OSGI configuration is the right way to do. Is it really an AEM config which should only be changed by deployment or should it rather be something which a "super author" should be able to its own, without AEM reconfiguration or deployment?

bundle/src/main/java/com/adobe/acs/commons/fetchfile/impl/FileFetchImpl.java
klcodanr: Ah yes, this is a much better idea and allows for avoiding the large binary issue above since I no longer have to compute the hash.

klcodanr:
Hey @joerghoh, I thought about whether to make this an authorable console vs the OSGi configuration and ended up coming down on the side of OSGi based on the use case I'd created this for. The use case I created this for is downloading a single file for referencing from a template / component script. To reference the file you would need to update code anyway so having an authorable console for configuring this would not be helpful

OSGi also helps in this case as you can provide different configurations per run mode and thus support having different files per environment in the same codebase. 

The nice thing is that this isn't _limited_ to Launch but could also be used for other DM scripts or non-authenticated JSON requests where the Same-Origin Policy is a problem, but in all of these cases, to actually _use_ the file you'd need some sort of code.

bundle/src/main/java/com/adobe/acs/commons/fetchfile/impl/FileFetchImpl.java
klcodanr: Fixed by no longer having to support generating the hash

bundle/src/main/java/com/adobe/acs/commons/fetchfile/impl/FileFetchImpl.java
klcodanr: Added here and generally cleaned up the log messages

bundle/src/main/java/com/adobe/acs/commons/fetchfile/impl/FileFetchImpl.java
klcodanr: I'm just using this as a shortcut in the JUnit tests to add the private fields.

klcodanr:
awesome, thank you @badvision !

PR #2120, opened 2019-11-21
Rewrite the testsuite for the JCRHttpCache

joerghoh:
* use SlingContext instead lots of mocks.
* and fix some issues on the way


joerghoh:
@niekraaijmakers WDYT?

niekraaijmakers:
I approve :)

PR #2124, opened 2019-11-28
cleanup logging to reduce noise in unittests

joerghoh:
* review exception logging
* convert a few occasions where JUL is used to SLF4J
* suppress messages from Oak which are always written when using a JCR_OAK repository for unittests.

The job logs should now be much clearer.


joerghoh:
Yay, the job log of Travis now fits into the regular screen, no need to always load the raw logs.

joerghoh:
coverall job is already fixed on master.

joerghoh:
@davidjgonzalez @badvision it's ready, can you please merge?

PR #2127, opened 2019-11-29
AEM6.3-CFP6.3.3.6-Multifield-RTE-Tag-Fix

pollenshukla:
**Issue:** The RTE and Tag component is not pre-populating the authored text after the installing the [https://www.adobeaemcloud.com/content/marketplace/marketplaceProxy.html?packagePath=/content/companies/public/adobe/packages/cq630/cumulativefixpack/AEM-CFP-6.3.3.6](AEM 6.3 Cumulative Fix Pack)

**Cause:**
The RichText component received enhancements in the 6.3.3.6 CFP. It was breaking multifield RTE and Tagfield prepopulation.

**Impacted AEM sling:resourceType in multifield scenarios:** 
- cq/gui/components/authoring/dialog/richtext
- cq/gui/components/coral/common/form/tagfield

**Impacted ACS AEM versions:**
- 3.11 (Project is running this version)
- 4.3.4 (I have upgraded the ACS AEM to 4.3.4, still the issue is not resolved, hence raising this PR as a fix for the community)

### DIFF
--- a/ui.apps/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/composite-multifield/source/touchui-composite-multifield.js
+++ b/ui.apps/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/composite-multifield/source/touchui-composite-multifield.js
@@ -188,7 +188,7 @@
                         }
 
                         _.each(record, function (rValue, rKey) {
-                            $field = $($fieldSets[i]).find("[name='./" + rKey + "']").last();
+                            $field = $($fieldSets[i]).find("[name='./" + rKey + "']").first();
 				
 			    if ($field.hasClass("coral-RichText-editable")) {
 	                        $field = $($fieldSets[i]).find("input[name='./" + rKey + "']").first();


pollenshukla:
Dear @joerghoh @badvision,
Please review.

badvision:
If you're on 6.3 why aren't you using the multifield component supplied by the product?

davidjgonzalez:
@pollenshukla  is this fixed already in master via https://github.com/Adobe-Consulting-Services/acs-aem-commons/commit/38ddbfd513520213e6a9451013988813c5ca7f72 ?

AtikahNaza:
Hi @davidjgonzalez ,
The fix in https://github.com/Adobe-Consulting-Services/acs-aem-commons/commit/38ddbfd513520213e6a9451013988813c5ca7f72 is only working for RTE in multifields. So, yes, it does need a condition for Tagfields.

Or, my suggestion is, as the PR raised by @pollenshukla , replace ".last()" with ".first()" without the condition, and it will work with both RTE and the Tagfields / other components in general.

davidjgonzalez:
@AtikahNaza im trying to figure out if there was a reason why `.last()` was used to begin with (instead of `.first()`) and if there are *ANY* use-cases were only checking `.first()` would break it. How do we know that switching it wholesale like this won't break for other (perhaps less common) widgets?

AtikahNaza:
@davidjgonzalez yes, you are right. Even though we had tested this on all of our components where it does not break; there is no guarantee that it would not break for others.

davidjgonzalez:
@AtikahNaza could you merge master into this, and then add a condition to that checks specifically for the Tags field? I dont think we should change all other behavior without knowing the consequences. it seems `.first()` works for everything except RTE (which is fixed in master) and Tags field (which your PR was tackling).

AtikahNaza:
Hi @davidjgonzalez 

We have tested this and it is backward compatible with ACS v3.13.0.

**New observation:** We found out that this fix breaks the validator for the text-field inside multifield upon field pre-population when opening the dialog. 

Even if the values are **correctly prepopulated**, the validator was still showing the red highlight indicating the error. 

However, we found that this is a **cosmetic error**, as the values populated are correct (and updated), and we were still able to submit and save the values inside the field.

PR #2128, opened 2019-11-29
Further cleanup in unittests

joerghoh:
Cleanup unittests to make the log more readable.


joerghoh:
We have some stacktraces left and I don't know how to avoid them. I don't want to remove them from being logged.

But the logs are much cleaner than before, so I would leave it as it. PR can get merged.

PR #2131, opened 2019-12-05
Feature/bug2084 cmf checkbox

badvision:
This PR adds some neat introspection hacking features to MCP Forms.
1) The introspection util class has methods to decipher default values defined on fields in beans, e.g. ```String test = "defaultValue"``` can be detected with this utility.
2) Checkbox components are checked by default if the boolean field value is set to true by default
3) Text components (and similar) take the defined value as the "value" parameter, generally interpreted as the default field value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 ### Added
 - #2118 - Adding functionality to showhidedialogfields TouchUI widget
 - #2110 - Adding File Fetcher for downloading and caching remote files in AEM Assets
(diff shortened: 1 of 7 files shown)


badvision:
This PR adds some neat introspection hacking features to MCP Forms.
1) The introspection util class has methods to decipher default values defined on fields in beans, e.g. ```String test = "defaultValue"``` can be detected with this utility.
2) Checkbox components are checked by default if the boolean field value is set to true by default
3) Text components (and similar) take the defined value as the "value" parameter, generally interpreted as the default field value.

badvision:
@gutsy Let me know if this answers your original issue sufficiently.

PR #2138, opened 2019-12-20
Remove the dependency to the joda-time library

joerghoh:
and replace it by java.time classes.


badvision:
I'm not sure where TypeUtil is being used but it would be a good idea to ensure that there are no side effects with doing this.  What's wrong with Joda time anyway?

joerghoh:
There's nothing wrong with it, I just came across when updating the test libraries, and found it only used for formatting. The change in TypeUtil is actually tested ( and that test helped me quite a bit ...)

PR #2153, opened 2020-01-07
Initial commit for SyntheticSlingHttpRequestExecutor - to create full…

niekraaijmakers:
configurable synthetic sling requests that can be executed at will, without a reference to any sling request / response.

vs the getRequestDispatcher, this will trigger an actual request, not include, and is fully configurable (including extension). This is handy for example if you need to get the .model.json wrapped in java to send as payload for SSR.

still needs unit tests. @justinedelson WDYT?


joerghoh:
@niekraaijmakers I think that this could also be a valuable contribution to Sling itself. WDYT?

justinedelson:
I agree with @joerghoh. With the exception of the WCMMode piece, this seems like something which would be better done in Sling. In fact, such a thing more or less already exists: https://sling.apache.org/documentation/bundles/servlet-helpers.html

See also the original thread (started by @joerghoh  although he may not remember): https://markmail.org/message/4m6f32cj4eoxqzyx#query:+page:1+mid:4m6f32cj4eoxqzyx+state:results

niekraaijmakers:
@justinedelson that is true. thing is it's not included in the AEM stack so it needs to be imported in AEM on a project basis. I end up having reimport this every time. Maybe we could make it happen to have this (servlet helpers) included in the AEM stack? Then we could use that simply as dependency instead.

niekraaijmakers:
@justinedelson  do we still want to do this? if yes il improve the test coverage and fix the tests, if not il put it on ice.

PR #2018, opened 2019-08-13
Add more granular environment indicator css control

paul-bjorkstrand:
* enables the ability to always include the base CSS (regardless if cssOverride is set)
* enables the ability to always include the background color CSS (regardless if cssOverride is set)
* separates the three components of the CSS into separate style rules, making it easier to use common to all environments override CSS in conjunction with a custom CSS background color
* keeps backwards compatibility intact

PR #2174, opened 2020-01-24
Add more granular environment indicator css control - take 2

paul-bjorkstrand:
New PR (superseding #2018), updating the existing branch seemed a little ugly, so I created a new one

* enables the ability to always include the base CSS (regardless if cssOverride is set)
* enables the ability to always include the background color CSS (regardless if cssOverride is set)
* separates the three components of the CSS into separate style rules, making it easier to use common to all environments override CSS in conjunction with a custom CSS background color
* keeps backwards compatibility intact


paul-bjorkstrand:
@badvision and @davidjgonzalez, I gave up on the AssertJ problem from the old PR, and (begrudgingly) rewrote this with Hamcrest and vanilla Junit. I also cleaned up the extraneous files as well.

davidjgonzalez:
What’s with those marketo commits?

paul-bjorkstrand:
@davidjgonzalez the build was failing because of the baseline plugin (telling me to up the versions)

paul-bjorkstrand:
So, I blindly obliged, perhaps in error. I just wanted it to build locally :)

PR #2175, opened 2020-01-27
added CQ Page wrapper class

niekraaijmakers:
small PR :)


joerghoh:
Hi @niekraaijmakers 

what is the purpose of this class? Can you outline usecases where this class is useful? 

Actually I don't see any purpose to do that, because from my point of view the page interface should only be used, and never be implemented on project side; it's just missing the ``@ProviderType`` annotation right now.

Issue #2177, opened 2020-01-30
MCP Data importer not working in 6.5.3


slreynol:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: AEM 6.5.3
* [ ] ACS AEM Commons Version: 4.4.0
* [ ] Reproducible on Latest? yes

### Expected Behavior
MCP Data importer should upload spreadsheet

### Actual Behavior

Data importer fails, exact same spreadsheet worked on 6.5.2, does not appear to even start the process as the 
![image](https://user-images.githubusercontent.com/5315439/73466667-35f64c80-4350-11ea-9387-654ccedb12c9.png)

Error log: 
ava.lang.NoSuchMethodError: org.apache.poi.ss.usermodel.Cell.getCellType()I
at com.adobe.acs.commons.data.Variant.setValue(Variant.java:99)
at com.adobe.acs.commons.data.Variant.<init>(Variant.java:79)
at com.adobe.acs.commons.data.Spreadsheet.readRow(Spreadsheet.java:185)
at com.adobe.acs.commons.data.Spreadsheet.buildSpreadsheet(Spreadsheet.java:156)
at com.adobe.acs.commons.data.Spreadsheet.buildSpreadsheet(Spreadsheet.java:137)
at com.adobe.acs.commons.mcp.impl.processes.DataImporter.buildProcess(DataImporter.java:200)
at com.adobe.acs.commons.mcp.impl.ProcessInstanceImpl.run(ProcessInstanceImpl.java:199)
at com.adobe.acs.commons.mcp.impl.ControlledProcessManagerServlet.doStartProcess(ControlledProcessManagerServlet.java:152)
at com.adobe.acs.commons.mcp.impl.ControlledProcessManagerServlet.doGet(ControlledProcessManagerServlet.java:97)
at com.adobe.acs.commons.mcp.impl.ControlledProcessManagerServlet.doPost(ControlledProcessManagerServlet.java:84)
at org.apache.sling.api.servlets.SlingAllMethodsServlet.mayService(SlingAllMethodsServlet.java:146)
at org.apache.sling.api.servlets.SlingSafeMethodsServlet.service(SlingSafeMethodsServlet.java:342)
at org.apache.sling.api.servlets.SlingSafeMethodsServlet.service(SlingSafeMethodsServlet.java:374)
at org.apache.sling.engine.impl.request.RequestData.service(RequestData.java:552)
[...]

Issue #2162, opened 2020-01-15
MCP Asset Folder Creator Fails on AEM 6.5 SP3


shsteimer:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc:6.5.3.0
* [ ] ACS AEM Commons Version: 4.4.0
* [ ] Reproducible on Latest? yes

### Expected Behavior

The folders should be created successfully without an error

### Actual Behavior

No folders are created, the following stack trace is produced, pointing to a no such method error when calling POI.

```
java.lang.RuntimeException: java.lang.NoSuchMethodError: 
org.apache.poi.ss.usermodel.Cell.setCellType(I)V at com.adobe.acs.commons.fam.impl.ActionManagerImpl.runActionAndLogErrors(ActionManagerImpl.java:175) at 
com.adobe.acs.commons.fam.impl.ActionManagerImpl.lambda$deferredWithResolver$0(ActionManagerImpl.java:159) at 
com.adobe.acs.commons.fam.impl.ActionManagerImpl$$Lambda$339/1098834023.run(Unknown Source) at com.adobe.acs.commons.fam.impl.TimedRunnable.run(TimedRunnable.java:91) at 
java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) at 
java.util.concurrent.FutureTask.run(FutureTask.java:266) at 
java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) at 
java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617) at 
java.lang.Thread.run(Thread.java:745) 
Caused by: java.lang.NoSuchMethodError: org.apache.poi.ss.usermodel.Cell.setCellType(I)V at 
com.adobe.acs.commons.mcp.impl.processes.AssetFolderCreator.parseAssetFolderCell(AssetFolderCreator.java:189) at 
com.adobe.acs.commons.mcp.impl.processes.AssetFolderCreator.parseAssetFolderRow(AssetFolderCreator.java:170) at 
com.adobe.acs.commons.mcp.impl.processes.AssetFolderCreator.lambda$parseAssetFolderDefinitions$0(AssetFolderCreator.java:151) at 
com.adobe.acs.commons.mcp.impl.processes.AssetFolderCreator$$Lambda$342/335134771.accept(Unknown Source) at 
[...]

PR #2178, opened 2020-01-30
Feature/support poi 4

shsteimer:
Thanks @badvision for making these changes.  Rebased against master as we discussed so this can resolve both #2177 and #2162 .

I've tested this against 6.5.3 and 6.5.0 with Asset Folder created and it works flawlessly.  I haven't tested 6.4 or 6.3, or other MCP tools, but can do that if required.

Opening Draft PR for now so code climate and all the auto checks can do their thing while I finalize testing locally.  Will convert to full PR later today.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 
 ### Fixed
 - #2082 - ETag filter never sends 304
-- #2148 - Bugfix for displaying sizes (addresses #2132)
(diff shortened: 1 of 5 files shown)


badvision:
@shsteimer We need that _NONE constant to remain as-is but you can tell Code Climate to take a hike by disabling the check squid:S00115 for that method

badvision:
The reason the constant starts with an underscore is because the newer version of POI has its own enum that uses that exact name, so we need a 1:1 to do the conversion correctly without extra logic in the code.

shsteimer:
@badvision CC is now happy, thanks for the tip.

PR #2180, opened 2020-02-02
[DoNotMerge] Drop support of AEM 6.3 and older ( fix #2167)

### LINKED ISSUE #2167: Poll: increase minimal version to AEM 6.4
This is an informal poll about dropping support for AEM 6.3 (all servicepacks/cfps) and older version. We have a number of changes (mostly related to the repo restructuring in AEM 6.4), which are currently blocked by the necessity to support AEM 6.3 as well.

I found these related issue.
* #1909 and #1975 
* #2029 could be closed
* #1408 

(and probably much more)

WDYT?

joerghoh:
Only support AEM 6.4 and higher


davidjgonzalez:
This will require a major version bump on release. If there are other back-breaking changes we should do them as part of this as well.

joerghoh:
```EMailServiceImplTest``` is hard to migrate to AEM 6.4 because the uber.jar is missing the class ```com.day.cq.commons.mail.impl.HtmlParserAccessor```. This class is not part of the official API, but it's directly referenced from the class ```com.day.cq.commons.mail.MailTemplate``` (official API). Thus some tests inside this class fail.
This is clearly a product problem (or at last a problem of the uber.jar of AEM 6.4), and I don't think that the 6.4 uber.jar is going to be fixed for this. We have to live with it.

Which means that we probably need to change tests to adapt to this behavior, and we will likely loose coverage, because testing that functionality will get much harder (I see the need for quite some mocking).

@justinedelson, @davidjgonzalez, @badvision What is your opinion on this?

joerghoh:
(Adobe internal ticket reference: CQ-4288948)

joerghoh:
I don't think that we can fix this issue with the AEM uber.jar here, and rewriting the test cases for it also does not seem very reasonable. Thus I recommend to

* ignore these testcases
* accept the drop in code coverage
* add a huge TODO at the top of the test class

joerghoh:
I merged this PR into feature/228-prepare-release-500.

PR #2181, opened 2020-02-03
Only run dependency-check-maven-plugin in dedicated profile

### LINKED ISSUE #2166: Only execute dependency-check-maven on Travis CI
Right now the goal "dependency-check-maven:check" might take a lot of time. I think it is enough to only execute that check with Travis CI but not for local Maven builds.
To achieve that I would propose executing that goal only with a specific Maven profile (which is only active on Travis).

kwin:
By default only active in one of the Travis matrix builds (enabled via
environment variable DEPENDENCY_CHECK

This closes #2166


badvision:
image
Not sure if this was faster...

A run from another PR looked like this:
image

Is this working as expected?

badvision:
image

Weird.. that must have just been a fluke.  Well, this doesn't look different than the normal builds that run already in the main branch but I don't see the harm of accepting this PR.

kwin:
@badvision That is a misunderstanding. The profile is active on Travis (i.e. there it is exactly the same build speed). Only local builds without any explicit profiles are a lot faster due to the dependency-check-maven-plugin not being executed then.

badvision:
Awesome, works for me.  Thanks for cleaning that up!

Issue #2190, opened 2020-02-18
AEM as a Cloud Service vs QuickStart (AMS/On-Prem) Detection


davidjgonzalez:
It would be nice to have internal tooling in ACS Commons that could be used to detect if ACS Commons is deployed to AEM as a Cloud Service (either a true Cloud Env or AEM SDK Quickstart env) or an AMS or OnPrem deployment.

The tooling should support 2 uses:

1) Java API that allows for in-execution detection
    ```
    if (requireAem.isCloudService()) {  ... } else { ... }
    ```
2) OSGi Service that can be used to enable/disable components entirely.
    ```
    @Reference(target="(cloud-ready=true)" scope = ReferenceCardinality.MANDATORY)
    RequireAem requireAem;

    ... OR ...

    @Reference(target="(cloud-ready=true)" scope = ReferenceCardinality.MANDATORY)
    RequireAem requireAem;
    ```

I *think* we can do this by simply looking at the Product Info Versions.... GA Cloud versions are start at `2019.12.x` and AMS/On-Prem are (currently) in the `6.x` version.

kwin:
I like especially proposal 2. But I think the OSGi service should be more generic so that it can support arbitrary requiring either specific versions or hosting options together with the target attribute (https://osgi.org/javadoc/osgi.cmpn/7.0.0/org/osgi/service/component/annotations/Reference.html#target--). I would propose to introduce a RequireAemVersion interface with implementations having service properties `version`, `hosting` and in the future maybe others. That way one could easily depend e.g. on a specific version range or combine multiple hosting options.

davidjgonzalez:
@kwin Ok - made some tests; Moved it to be named `RequireAem` as like that name better, and deprecated the old `AemCapabilityHelper`

kwin:
@davidjgonzalez Can we close this now?

PR #2191, opened 2020-02-18
#2190 - AEM as a Cloud Service detection

davidjgonzalez:
See #2190 

@joerghoh  WDYT?


kwin:
As already mentioned in https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2190#issuecomment-587718748 I would combine version, hosting and maybe other properties in the future in a single service implementation having those information exposed as service properties. That way one could easily depend on them with the target attribute.

davidjgonzalez:
@kwin good call - updated POC .. were you thinking something like that? I kinda feel like version is a YAGNI ... ?

bundle/src/main/java/com/adobe/acs/commons/util/compatability/impl/AemCompatibilityImpl.java
```diff
@@ -0,0 +1,76 @@
...
+    @Reference(
+            policy = ReferencePolicy.DYNAMIC,
+            policyOption = ReferencePolicyOption.GREEDY
+    )
+    private transient ProductInfoProvider productInfoProvider;
+    private transient ProductInfo productInfo;
+
+    @Override
+    public Hosting getHosting() {
+        if (productInfo.getVersion().compareTo(originalCloudServiceVersion) > 0) {
```
kwin: is it ok to derive that from the version? What about running with AEM SDK 20xx locally? Shouldn't we rather derive the hosting option from the documented run modes from AEMaaCS (https://docs.adobe.com/content/help/en/experience-manager-cloud-service/implementing/deploying/overview.html#runmodes)?

davidjgonzalez: so - i guess there's some ambiguity here...  here's how i think of it:

* AEM as a Cloud Service (Cloud)
   * Running in the Adobe Cloud 
   * AEM SDK QuickStart Jar running locally
* AMS/On-Prem (Non-Cloud)
   * Any 6.x based AEM deployment

For "AEM as a Cloud Service" .. i'd think we'd want to limit the local AEM SDK QuickStart Jar to run the same feature set as in the "real" Cloud, otherwise, devs might get confused as to why something works locally but not when deployed.

Are you thinking there are 3 types we should target?

* AEM as a Cloud Service running in the Cloud
* AEM as a Cloud Service SDK QuickStart Jar running locally for developers
* AEM 6.x (AMS and OnPrem)

Also - perhaps "hosting" is a bad term ..

kwin: I think indeed the property should be named differently. Cannot come up with a good proposal right now... Maybe something like "OakCompositeNodeStore" or "StrictMutableImmutableSeparation".

davidjgonzalez: @kwin what's the usecase for running something on local dev but not in the cloud?

kwin:
It would be good to expose whether a root point is read-only or not (https://jackrabbit.apache.org/oak/docs/nodestore/compositens.html#Checking_for_read-only_access) with specific properties.

bundle/src/main/java/com/adobe/acs/commons/util/compatability/impl/AemCompatibilityImpl.java
kwin: most probably one should rather rely on a specific property whether a mount point is read-only or not.

davidjgonzalez: That gets us a split at: "AEM running IN the Cloud"  and "Everything else" ... as noted before i see the split at the Product level: AEM as a Cloud Service (including local dev quickstart) and AEM 6.x (AMS/OnPrem) ...

We shouldnt be encouraging use of features on AEM as a CS local dev that wont work when deployed to AEM in the Cloud. Thats just confusing.

kwin: then maybe "type" is the right property name here because in fact it is different product types

bundle/src/main/java/com/adobe/acs/commons/util/compatability/AemCompatibility.java
```diff
@@ -0,0 +1,58 @@
+package com.adobe.acs.commons.util.compatability;
```
joerghoh: ...compatibility :-)

davidjgonzalez: How embarrassing...

bundle/src/main/java/com/adobe/acs/commons/util/compatability/impl/AemCompatibilityImpl.java
davidjgonzalez: Type, mode...

Also, is there a compelling reason to differentiate between AEM as a cloud service in the cloud and local dev? If there is we should try to figure out what app the splits are now and how to detect. I’m in favor of the 2 types split by logical product (cloud vs 6.x) but if we have a good reason to add further granularity that justifies the additional complexity, we could do it (assuming we can detect any new  splits)

davidjgonzalez:
@kwin what would we do w the mutable/immutable knowledge, if we exposed that?

davidjgonzalez:
@kwin  @joerghoh I slept on this - and made a few changes before work today. After thinking about it, I think we should atleast start off simple:

Cloud-Ready or Not-Cloud-Ready.

* **Cloud-Ready**: encompasses AEM as a Cloud Service running IN the Adobe Cloud _as well as_ AEM as a Cloud Service SDK QuickStart Jar.

* **Not-Cloud-Ready**: AEM running on AMS or OnPrem.

I don't think we need to differentiate between local-dev and hosted (stage/prod) as local dev should be in service of developing an app FOR prod. So local dev should be as close to prod as possible to avoid confusion and surprises.

Based on this, I updated the OSGi Component/Service to have a simple filter: `cloud-ready=true|false` ..
I also added it to the existing `AemCaptabilityHelper` to avoid yet another API surface. 

We can always add other bits of exposure, but ideally, we keep it very simple as to what features work where, and not leak in concepts like specific version #s, mutable mounts, etc; We either have Features that:

1. Work everywhere
1. Work ONLY on AEM as a Cloud Service
1. Work ONLY on AMS/OnPrem

bundle/src/main/java/com/adobe/acs/commons/util/AemCapabilityHelper.java
```diff
@@ -21,19 +21,70 @@
...
+import com.adobe.acs.commons.util.compatability.AemCompatibility;
 import org.osgi.annotation.versioning.ProviderType;
 
+
+import org.osgi.annotation.versioning.ProviderType;
+
+/**
+ * This OSGi Service can be used in 2 different ways....
+ *
+ * 1. It can be used In-line in code to detect the Hosting type (Cloud or Non-Cloud)
```
kwin: This should be adjusted as well

bundle/src/main/java/com/adobe/acs/commons/util/impl/AemCapabilityHelperImpl.java
```diff
@@ -21,21 +21,89 @@
...
+    @Deactivate
+    protected void deactivate() {
+        productInfo = null;
+
+        if (serviceRegistration != null) {
+            serviceRegistration.unregister();
+        }
+    }
+
+    /** Deprecated **/
```
kwin: what is deprecated?

davidjgonzalez:
I moved some code around (I like the `RequireAem` name.. and deprecated the old `AemCapabilityHelper` .. wrote some tests, and consolidated hamcrest deps as it was causing me build issues for some reason..?

davidjgonzalez:
@kwin @joerghoh the other solution some folks in Eng floated was creating 2 release artifacts, one for CloudReady and one for NotCloudReady, and have the maven build strip out everything that per each build... TBH, im not entirely sure how to do this and/or how clean/maintainable it would be. If one of you know how to do this and feel like its a better solution, we could table this PR for that -- I just dont know of the top of my head and dont have any appetite to fiddle around with it :)

bundle/src/main/java/com/adobe/acs/commons/util/impl/RequireAemImpl.java
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * Allows ACS Commons features determine if they are running on AEM as a Cloud Service or not.
+ */
+@Component(
+        service = {},
+        immediate = true
+)
+public class RequireAemImpl implements RequireAem {
+    private static final Logger log = LoggerFactory.getLogger(RequireAemImpl.class);
+    private static final String PN_CLOUD_READY = "cloud-ready";
```
kwin: what about `distribution` with values `cloud-ready` and `classic` or `onprem`? That way we could extend that further, like `cloud-ready-minimum`, `cloud-ready-forms` and whatever may come up in the future. Maybe just as addition, but I really see that in the future we may have even more than those two distributions...

davidjgonzalez: that works... im leaning towards `classic` since AMS is typically not considered "onprem" ..

davidjgonzalez: @kwin Ok - updated the Impl per your suggestion

davidjgonzalez:
@joerghoh any last thoughts before merge?

bundle/src/main/java/com/adobe/acs/commons/util/impl/RequireAemImpl.java
```diff
@@ -0,0 +1,105 @@
...
+            distribution =  Distribution.CLASSIC.getValue();
+            log.debug("Registering [ {} ] as an OSGi Service so it be be used to enable/disable other OSGi Components", this.getClass().getSimpleName());
+        }
+
+        properties.put(PN_DISTRIBUTION, distribution);
+        properties.put(PN_VERSION, version);
+
+        serviceRegistration = bundleContext.registerService(RequireAem.class.getName(), this, properties);
+
+        log.error("Registering [ RequireAem.class ] as an OSGi Service with OSGi properties [ distribution = {}, version = {} ] so it be be used to enable/disable other OSGi Components",
```
joerghoh: why ```error()```? ```info()``` should suffice here.

joerghoh:
Where does the final documentation go? We should be very clear, what everyone can expect in case of
* Quickstart (on-prem, AMS)
* AEMCS in local usage (using the SDK)
* "real AEMCS"

(you lost me in the discussion above).

joerghoh:
> @kwin @joerghoh the other solution some folks in Eng floated was creating 2 release artifacts, one for CloudReady and one for NotCloudReady, and have the maven build strip out everything that per each build... TBH, im not entirely sure how to do this and/or how clean/maintainable it would be. If one of you know how to do this and feel like its a better solution, we could table this PR for that -- I just dont know of the top of my head and dont have any appetite to fiddle around with it :)

I was part of the discussion, and indeed it's unclear how achieve this in a maintainable way. Of course our implementation of this distinction is limited in a way, that all code of ACS AEM Commons is always deployed and resolved, thus everything's fine if the API/exported packages are the same on Quickstart and AEMCS. As soon as there are major changes, our approach might break, and then we have to look for a new solution.

I think we should accept that for now, and look for a better solution if we face such a situation.

davidjgonzalez:
@joerghoh  thanks for the catch, just to edify you before the merge, there are 3 "modes":

1. Only works on AEM as a Cloud Service
    * `(distribution=cloud-ready)`
    * This is determined by checking if the AEM version is 2019.12 or greater (which is the first Major/Minor GA version AEM as a clo
    * This will resolve on EITHER: 
       * local development AEM as a Cloud Service SDK Quickstart Jar
        * The "real" AEM as a Cloud Sevice (running in Adobe's Cloud)

2. Only works on AMS/OnPrem 
    * `(distribution=classic)`
    * This is determined by checking if the AEM version is less than 2019.12 (which is the first Major/Minor GA version AEM as a Cloud Service) -- and this really targets AEM versions: 6.x.x (which also happens to be less than version 2019.12)
    * This will resolve on EITHER: 
       * AEM running on AMS or OnPrem Services
       * local development Quickstart Jar

3. Works everywhere!
    * In this case, you don't use RequireAem at all.

This service is intended to be used internally in ACS Commons to toggle features. This could be added to the contributor's document, but I think, in reality, we need to make sure as PR's come in, they are using this where applicable.

We are already marking ACS Commons documentation pages as being AEM as a Cloud Service incompatible for features that work on 6.x but not on AEM as a Cloud Service.

Issue #2192, opened 2020-02-20
Generic Lists | Edit, Quick publish, Manage publication options are missing In Touch UI


nshindhe:
## Required Information

AEM VERSION: 6.5.3
ACS AEM COMMONS VERSION: 4.4.0
Reproducible Yes

Issue 1: Edit, Quick publish, Manage publication options are missing In Touch UI
URL: http://localhost:4502/generic-lists.html/etc/acs-commons/lists

Steps to reproduce: 
1. Open url http://localhost:4502/generic-lists.html/etc/acs-commons/lists as admin user
2. Select one of the list  e.g Font Awesome Icons
3. now observe the missing toolbar options "Edit, Quick publish, Manage publication"

Expected Behavior:
There should be options edit and publish the lists

Actual Behavior:
Options to edit and publish lists are missing 

Issue 2: When the list pages are placed under /content/project and /conf/project no data is being displayed however the same page is working fine when placed under /etc/acs-commons/lists

After observing the logs we see the below error:

`10.02.2020 16:19:36.674 *ERROR* [0:0:0:0:0:0:0:1 [1581331776670] GET /conf/project/common-configurations/settings/region-list-en.html HTTP/1.1] org.apache.sling.servlets.get.impl.DefaultGetServlet No renderer for extension html, cannot render resource JcrNodeResource, type=acs-commons/components/utilities/genericlist, superType=null, path=/conf/project/common-configurations/settings/region-list-en/jcr:content`

badvision:
Please confirm if this works for you now @nshindhe .

PR #2194, opened 2020-02-21
Add actions of Quickpublish and Manage publication to Generic List console

josmora:


badvision:
This looks like an appropriate fix for #2192 .  Thank you!

PR #2202, opened 2020-02-28
Feature/feature dynamic script resolver support

badvision:
This evolution will not be televised.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/mcp/DynamicScriptResolverService.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/mcp/DynamicScriptResolverService.java
@@ -1,6 +1,9 @@
 /*
- * Copyright 2020 Adobe.
- *
+ * #%L
+ * ACS AEM Commons Bundle
+ * %%
+ * Copyright (C) 2020 Adobe
+ * %%
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
@@ -12,8 +15,10 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ * #L%
  */
-package com.adobe.acs.commons.mcp;
+
+ package com.adobe.acs.commons.mcp;
 
 import com.adobe.acs.commons.mcp.form.FieldComponent;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


badvision:
@joerghoh, @davidjgonzalez  this introduces a necessary extension point to mcp, doesn't change any functionality otherwise.

bundle/src/main/java/com/adobe/acs/commons/mcp/DynamicScriptResolverService.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Provides hooks for injecting additional MCP process definitions.
+ * For example, definitions that are not written in Java as OSGi services.
+ * Identifier is up to the resolver to determine, examples might be a unique name or a path.
+ * It is recommended to use something as specific as JCR path if at all possible for identifier.
+ */
+@ConsumerType
+public interface DynamicScriptResolverService {
+    Map<String, ProcessDefinitionFactory> getDetectedProcesDefinitionFactories(ResourceResolver rr);
```
joerghoh: Can you provide some javadoc for this publicly exposed interface and its methods?

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/ControlledProcessManagerImpl.java
```diff
@@ -64,6 +73,9 @@
     @Reference(cardinality = ReferenceCardinality.MANDATORY_MULTIPLE, bind = "bindDefinitionFactory", unbind = "unbindDefinitionFactory", referenceInterface = ProcessDefinitionFactory.class, policy = ReferencePolicy.DYNAMIC)
     private final List<ProcessDefinitionFactory> processDefinitionFactories = new CopyOnWriteArrayList<>();
 
+    @Reference(cardinality = ReferenceCardinality.OPTIONAL_MULTIPLE, bind = "bindScriptResolverService", unbind = "unbindScriptResolverService", referenceInterface = DynamicScriptResolverService.class, policy = ReferencePolicy.DYNAMIC)
```
joerghoh: if the bind and unbind methods are as simple as here, I don't think that they are necessary. IIRC these implementations a kind of default (though I don't have the specific chapter of the OSGI spec handy).

bundle/src/main/java/com/adobe/acs/commons/mcp/DynamicScriptResolverService.java
badvision: Sure thing!

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/ControlledProcessManagerImpl.java
badvision: Yeah, I think so to but at some point during development I ran into a weird bug where optional services weren't binding and I needed these methods to debug it.  I think it's a hold-over but ultimately benign.

PR #2212, opened 2020-03-03
Sitemap: Allow exclude with multiple properties and templates

maheshkarle70:
There might be many pages in the project which should be excluded from sitemap for example if page is using certain properties to make it protected or hideInNav is enabled or hideFromSitemap

With this implementation, we can simply give those properties which we can check if they are set to true on that page then that page will be excluded
Similarly certain page templates to be excluded can be specified against exclud.templates="temp1,temp2"


justinedelson:
IMO, instead of using a comma-delimited value, it would be better to use a multi-valued property. If a single-valued property is specified (e.g. as `exclude.property` since to date that has been a single-valued property), that should be automatically handled by `PropertiesUtil.toStringArray`.

maheshkarle70:
> delimited

maheshkarle70:
> IMO, instead of using a comma-delimited value, it would be better to use a multi-valued property. If a single-valued property is specified (e.g. as `exclude.property` since to date that has been a single-valued property), that should be automatically handled by `PropertiesUtil.toStringArray`.

This comment is addressed

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -124,6 +124,9 @@
     @Property(label = "Character Encoding", description = "If not set, the container's default is used (ISO-8859-1 for Jetty)")
     private static final String PROP_CHARACTER_ENCODING_PROPERTY = "character.encoding";
 
+    @Property(label = "Exclude Page from Sitemap", description = "The String [cq:Template] property on /jcr:content of page")
```
davidjgonzalez:
```suggestion
    @Property(label = "Exclude Pages (by Template) from Sitemap", description = "The String [cq:Template] property on /jcr:content of page")
```

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -298,8 +305,25 @@ private void write(Page page, XMLStreamWriter stream, ResourceResolver resolver)
...
+            }
+        }
+        return flag;
+    }
+
+    private boolean isExcluded(Page page){
+        boolean flag = false;
+        if(this.template != null){
+            for(String pageTemplate : this.template){
+                flag = flag || page.getProperties().get("cq:template", String.class).equalsIgnoreCase(pageTemplate);
```
davidjgonzalez:
```suggestion
                flag = flag || page.getProperties().get("cq:template", "").equalsIgnoreCase(pageTemplate);
```

Changing default value in `.get(.., "")` to empty string to avoid potential NPE.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -153,6 +156,8 @@
 
     private boolean removeTrailingSlash;
 
+    private List<String> template;
```
davidjgonzalez:
```suggestion
    private List<String> excludedPageTemplates;
```
Better to be clear what this variable represents.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -175,6 +180,8 @@ protected void activate(Map<String, Object> properties) {
         this.urlRewrites = ParameterUtil.toMap(PropertiesUtil.toStringArray(properties.get(PROP_URL_REWRITES), new String[0]), ":", true, "");
         this.removeTrailingSlash = PropertiesUtil.toBoolean(properties.get(PROP_REMOVE_TRAILING_SLASH),
                 DEFAULT_REMOVE_TRAILING_SLASH);
+        this.template = Arrays.asList(PropertiesUtil.toStringArray(properties.get(TEMPLATE_EXCLUDE_FROM_SITEMAP_PROPERTY),new String[0]));
```
davidjgonzalez:
```suggestion
        this. excludedPageTemplates = Arrays.asList(PropertiesUtil.toStringArray(properties.get(TEMPLATE_EXCLUDE_FROM_SITEMAP_PROPERTY),new String[0]));
```

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -298,8 +305,25 @@ private void write(Page page, XMLStreamWriter stream, ResourceResolver resolver)
...
+            for (String prop: multipleProperties) {
+                flag = flag || page.getProperties().get(prop, Boolean.FALSE);
+            }
+        }
+        return flag;
+    }
+
+    private boolean isExcluded(Page page){
+        boolean flag = false;
+        if(this.template != null){
```
davidjgonzalez:
```suggestion
        if(this. excludedPageTemplates != null){
```

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -298,8 +305,25 @@ private void write(Page page, XMLStreamWriter stream, ResourceResolver resolver)
...
+            for(String pageTemplate : this.template){
```
davidjgonzalez:
```suggestion
            for(String pageTemplate : this. excludedPageTemplates){
```

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -298,8 +305,25 @@ private void write(Page page, XMLStreamWriter stream, ResourceResolver resolver)
         stream.writeEndElement();
     }
 
-    private boolean isHidden(final Page page) {
-        return page.getProperties().get(this.excludeFromSiteMapProperty, false);
+    private boolean isHiddenOrProtected(Page page) {
+        boolean flag= false;
+        if(this.excludeFromSiteMapProperty != null){
+            String [] multipleProperties = this.excludeFromSiteMapProperty.split(",");
```
davidjgonzalez: This still needs updating to @justinedelson 's suggestions of making the osgi property a multi-value, which means you dont need to split on commas.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -298,8 +305,25 @@ private void write(Page page, XMLStreamWriter stream, ResourceResolver resolver)
         stream.writeEndElement();
     }
 
-    private boolean isHidden(final Page page) {
-        return page.getProperties().get(this.excludeFromSiteMapProperty, false);
+    private boolean isHiddenOrProtected(Page page) {
```
davidjgonzalez: Dont follow this method naming? ..  what does `...OrProtected(..)` mean?

It seems the more semnatic names for these 2 methods would be:

`isHiddenByPageProperty(..)` and  `isHiddenByPageTemplate(...)`

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
maheshkarle70: Yes, its done now

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
maheshkarle70: Yes, its done now

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
maheshkarle70: Yes, its done now

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
maheshkarle70: Yes, its done now

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
maheshkarle70: Yes, its done now

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
maheshkarle70: there might be pages like article which will be completely available to read only when user logs in, so these kind of pages can be considered as protected pages, but agree by what you suggested thanks

PR #2215, opened 2020-03-03
Parameterized granite include

niekraaijmakers:
Include granite resource type that allows for parameters to namespace and parameterize included.
Similar to : https://adobe-consulting-services.github.io/acs-aem-commons/features/deprecated/cqinclude-namespace/index.html

Bildschirmfoto 2020-03-03 um 18 49 51

Then in the snippet:

Bildschirmfoto 2020-03-03 um 18 52 03

Will result with a hidden equals true, and the name being "./block1/defaultInformation", with a  text label: someParameter.

I tried quite a bit with the EL resolver from granite but I cannot get this to work with namespacing and that is the whole purpose. I'm open to anybody getting this to work though. 1 Thing is that this approach requires you to know the type your taking out of the parameters so the now optional typecast would become required.

The format is as follows:

${{(OptionalTypeCast)paramKey:defaultValue}}

OptionalTypeCast can be Boolean, Date, Calendar, Double and Long. String is default.
Before writing tests / polishing it out i'd be happy to hear if you guys think it's a good idea or not.

Using this approach I've been able to prevent a lot of dialog duplication my project by reusing snippets with parameters and namespacing. This also works with nesting, you can namespace a namespaced include and it will get chained and parameters are passed downwards


justinedelson:
@niekraaijmakers Minor suggestion - I think it would be cleaner if the `namespace` property was on the include node, not the `parameters` node. Reason being that just in case the included component had a property named `namespace` it would be a conflict. Granted, this is probably not likely, but I think it would be safer. WDYT?

niekraaijmakers:
Well that makes sense to be honest. If you only specify the namespace you can leave out the parameters child node. 

What I would say is that you should be able to use namespace as a replacement as well though, that its the same thing basically.
Unless you mean to specify namespace as a parameter and NOT as a real namespace.. then I'd have to change a few stuff. But i'd rather have it as a reserved property. Thoughts?

justinedelson:
@niekraaijmakers I take back what I said. I had read the connection between the property names of the `parameters` resource and the included component (`granite/ui/components/coral/foundation/form/textfield` in your example) to be more direct than it actually is. But re-reading the example, I see there is no required connection -- the names of the parameters has no relationship with the names of the properties on the included component. My bad.

But I still think that moving `namespace` up to the `acs-commons/granite/ui/components/include` might make sense. It comes down (at least IMO) to what you wrote:

> What I would say is that you should be able to use namespace as a replacement as well though, that its the same thing basically.

If `namespace` is available as a parameter inside the expressions, then what you have make sense and it is just a "special" name. But if `namespace` isn't available as a parameter, then it should be moved up a level (to the include resource).

It's up to you as the author (and someone who needs this, presumably). I will just say that from my perspective the two things involved here -- namespacing the POST-submitted properties and allowing the properties of the included component to be parameterized -- do not seem connected. So my bias is to keep the namespace and the parameters separate.

Granted, this all may be retroactive justification for my initial misunderstanding :)

niekraaijmakers:
Agreed, I moved the namespace to the include itself. Please check again :)

bundle/src/main/java/com/adobe/acs/commons/granite/ui/components/include/IncludeDecoratorFilterImpl.java
```diff
@@ -0,0 +1,121 @@
...
+import javax.servlet.ServletResponse;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+
+
+import java.io.IOException;
+
+@Component(
+        service = Filter.class,
+        configurationPolicy = ConfigurationPolicy.OPTIONAL,
```
davidjgonzalez: Should this be Optional? Wont this add this Filter to the Request processing chain even for people that arent using this feature?

bundle/src/main/java/com/adobe/acs/commons/granite/ui/components/include/IncludeDecoratorFilterImpl.java
```diff
@@ -0,0 +1,121 @@
...
+                "sling.filter.resourceTypes=" + IncludeDecoratorFilterImpl.RESOURCE_TYPE
+        }
+        
+)
+public class IncludeDecoratorFilterImpl implements Filter {
+
+    static final String RESOURCE_TYPE = "acs-commons/granite/ui/components/include";
+    static final String NAMESPACE = "namespace";
+    static final String PARAMETERS = "parameters";
+    public static final String PREFIX = "PREFIX_";
```
davidjgonzalez: Would it hurt to make this value very "ACS Commons" specific just to "guarantee" no collisions?
... like `ACS_AEM_COMMONS__NAMESPACE_INCLUDE_` or something?

bundle/src/main/java/com/adobe/acs/commons/granite/ui/components/include/IncludeDecoratorFilterImpl.java
```diff
@@ -0,0 +1,121 @@
...
+        property= {
+                EngineConstants.SLING_FILTER_SCOPE+"="+ EngineConstants.FILTER_SCOPE_INCLUDE,
+                "sling.filter.resourceTypes=" + IncludeDecoratorFilterImpl.RESOURCE_TYPE
+        }
+        
+)
+public class IncludeDecoratorFilterImpl implements Filter {
+
+    static final String RESOURCE_TYPE = "acs-commons/granite/ui/components/include";
+    static final String NAMESPACE = "namespace";
```
davidjgonzalez: Would it hurt to make this value very "ACS Commons" specific just to "guarantee" no collisions?
... like ACS_AEM_COMMONS__NAMESPACE_ or something?

bundle/src/main/java/com/adobe/acs/commons/granite/ui/components/include/IncludeDecoratorFilterImpl.java
davidjgonzalez: @justinedelson WDYT? while this Filter would become active w/out someone using it, it does scope to resource type so it shouldn't be invoked unless someone intentionally apply the resource type...  Don't want to be too dogmatic; feels like a grey area wrt to "disabled by default" - i could go either way.

bundle/src/main/java/com/adobe/acs/commons/granite/ui/components/include/IncludeDecoratorFilterImpl.java
niekraaijmakers: agreed, let me put it there

niekraaijmakers:
Hang on, I'm going to add the option to namespace more then just "name". for the image component you need fileReference to be namespaced as well. I will add a OSGI configurable option.

niekraaijmakers:
Ok done & tested locally. Can somebody retrigger the build? 1 out of 5 failed randomly.

justinedelson:
@niekraaijmakers I have retriggered the build. Seems better now.

bundle/src/main/java/com/adobe/acs/commons/granite/ui/components/NamespacedTransformedResourceProvider.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * NamespacedTransformedResourceProvider
+ * <p>
+ * Transforms a resource underlying children with a namespace.
+ * </p>
+ *
+ * @author raaijmak@adobe.com
+ * @since 2020-03-09
+ */
+public interface NamespacedTransformedResourceProvider {
```
justinedelson: I think this interface is missing `@ProviderType`. Or is there a reason why a consumer would implement this interface?

ui.apps/src/main/content/jcr_root/apps/acs-commons/granite/ui/components/include/include.jsp
```diff
@@ -0,0 +1,26 @@
+<%@ include file="/libs/granite/ui/global.jsp" %><%
```
justinedelson: This might be a dumb question, but is there a way to *not* use a JSP here? Can this just be Java code? That would obviate the need (I think) for the `NamespacedTransformedResourceProvider` interface to be exported.

niekraaijmakers: Excellent point. Il give it a try and let you know.

niekraaijmakers: Excellent point. Il give it a try and let you know.

niekraaijmakers: Ok I really tried with a sling GET servlet, but this introduces a lot of complexity and I didn't get it to work. PR stands as is.

justinedelson: thanks for trying :)

niekraaijmakers: Can you look at the 1 remainder point from @davidjgonzalez above? :D

bundle/src/main/java/com/adobe/acs/commons/granite/ui/components/NamespacedTransformedResourceProvider.java
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.sling.api.resource.Resource;
+import org.osgi.annotation.versioning.ProviderType;
+
+/**
+ * NamespacedTransformedResourceProvider
+ * <p>
+ * Transforms a resource underlying children with a namespace.
+ * </p>
+ *
+ * @author raaijmak@adobe.com
```
davidjgonzalez: ?

bundle/src/main/java/com/adobe/acs/commons/granite/ui/components/include/IncludeDecoratorFilterImpl.java
davidjgonzalez: @justinedelson any strong opinion one way or another? if not, i'm inclined to let sleeping dogs lie.

justinedelson: I don't fully grok why this is optional. It seems like you are effectively opting in to using this by virtue of referencing the component resource type.

@davidjgonzalez is your concern that a consumer of ACS AEM Commons will use the resource type `acs-commons/granite/ui/components/include` for some _other_ purpose and be surprised to see this filter invoked?

davidjgonzalez: @justinedelson my concern (IIUC) is more this Sling Filter will be "registered" into the Sling Filter-stack even if you aren't actively using it (using it, meaning you don't have a resource w/ this resource type). Admittedly it won't (shouldn't) ever be executed unless this resource type is present, but, technically, the code *is* "active", and Sling has to make a decision in request processing to invoke it or not.

Optional vs Required shouldn't change any behaviors, its a more pedantic question; do we want OSGi Components to be active, even if they aren't invoked by Sling, by default - or do we want users to always, explicitly turn things on.

justinedelson: Yeah, I don't know what the performance impact is of just having the filter there without using the component. I can see that leaving this as optional is safer. But... what is the behavior when you _use_ the `acs-commons/granite/ui/components/include` resource type and this filter *isn't* active? A wise man once told me that "documentation is a polyfill for confusing design" :)

davidjgonzalez: Wise indeed! Let's keep it as is! :P

Issue #2228, opened 2020-03-23
SlingBindings is null and results in NullPointerException in usage with @SlingServlet


AnselmHa:
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.4.7
* [x] ACS AEM Commons Version: 4.4.2 / 4.4.3-SNAPSHOT
* [x] Reproducible on Latest? yes

### Expected Behavior
In case of using a @SlingServlet to resolve a request to a Sling model with a @ChildResourceFromRequest(injectionStrategy ...) the class OverridePathSlingRequestWrapper should not throw a NullPointerException.

### Actual Behavior
The class OverridePathSlingRequestWrapper throws a NullPointerException at:
`SlingBindings slingBindings = (SlingBindings) getSlingRequest().getAttribute(ATTR_SLING_BINDINGS);`
------------------------vvvvvvvvvvvv is null
`this.adapterManager = slingBindings.getSling().getService(AdapterManager.class);`

This happens because because no SlingBindings is found in case of usage of @SlingServlet with SlingModel with ChildResourceFromRequest

### Steps to Reproduce
- Create a @SlingServlet with `extends SlingAllMethodsServlet` e.g.:
`    @Override
    protected final void doPost(final SlingHttpServletRequest request, final SlingHttpServletResponse response)
        throws IOException {
       MySlingModelWithChildResource model = request.adaptTo(MySlingModelWithChildResource.class);
`
- Have a field in the MySlingModelWithChildResource like:
`
    @ChildResourceFromRequest(injectionStrategy = InjectionStrategy.OPTIONAL)
    private List<MyChildResource> childResource;

`

### Links
Will provide it if needed.

davidjgonzalez:
FWIW I experience something similar and had to manually add SlingBindings to my SlingHttpRequest [1] ... I'm not sure if there is a more elegant way to do this, but I think this is a larger problem that you have to fix in your Servlet (via a fix similar to below) or by Sling.

[1] https://github.com/Adobe-Marketing-Cloud/asset-share-commons/pull/470/files#diff-f2facd29e8ed61188f5a11e760681b2bR39

AnselmHa:
I created a pull request with using BundleContext if SlingBindings is missing.

AnselmHa:
Full stack trace of this issue:
```
[...]

PR #2229, opened 2020-03-23
SlingBindings is null and results in NullPointerException:

AnselmHa:
- If SlingBindings is null using BundleContext to get Service works fine

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/util/OverridePathSlingRequestWrapper.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/util/OverridePathSlingRequestWrapper.java
@@ -22,12 +22,13 @@ package com.adobe.acs.commons.util;
 import com.adobe.cq.sightly.WCMBindings;
 import com.day.cq.wcm.api.Page;
 import com.day.cq.wcm.api.PageManager;
-import com.day.text.Text;
 import org.apache.sling.api.SlingHttpServletRequest;
 import org.apache.sling.api.adapter.AdapterManager;
 import org.apache.sling.api.resource.Resource;
 import org.apache.sling.api.scripting.SlingBindings;
 import org.apache.sling.api.wrappers.SlingHttpServletRequestWrapper;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.FrameworkUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


AnselmHa:
#2228 Targets this issue

bundle/src/main/java/com/adobe/acs/commons/util/OverridePathSlingRequestWrapper.java
```diff
@@ -80,6 +82,22 @@ public OverridePathSlingRequestWrapper(final SlingHttpServletRequest request, fi
...
+     *
+     * @param slingBindings the slingBindings to get Sling Services
+     * @return        the AdapterManager from slingBindings.getSling() or the current bundle context
+     */
+    public static AdapterManager getAdapterManager(SlingBindings slingBindings) {
+        if (slingBindings != null && slingBindings.getSling() != null) {
+            return slingBindings.getSling().getService(AdapterManager.class);
+        } else {
+            BundleContext bundleContext = FrameworkUtil.getBundle(AdapterManager.class).getBundleContext();
+            return (AdapterManager) bundleContext.getService(bundleContext.getServiceReference(AdapterManager.class.getName()));
```
justinedelson: This technically creates a resource leak since the service reference is never "ungot". This also doesn't handle the case when the `AdapterManager` service is unregistered and reregistered. There's two alternatives I can suggest:

1. Move this logic to `com.adobe.acs.commons.util.impl.Activator` and use a `ServiceTracker` to track the service.
2. Deprecate the two-argument constructor and add the BundleContext as a 3rd argument. In this case, the AdapterManager service reference should be got and ungot inside the `adaptTo()` method. In theory, it would be possible to use a `ServletRequestListener` but given that this is a wrapper, it seems like it would be quite complex to discover the wrapper based on the destruction of the "true" request.

AnselmHa: Thanks for your suggestions.

- to 1.
I guess using Activator could be to much for a service
- to 2.
Do you mean something like this:
```
            if (result == null) {
                AdapterManager mgr = getAdapterManager(myBindings);
                if (mgr == null) {
 ```
instead of:
```
            if (result == null) {
                AdapterManager mgr = this.adapterManager;
                if (mgr == null) {
 ```
Do you know how to test this adaptTo in OverridePathSlingRequestWrapper, it isn't called during htl resolution or adaptTo in Servlet.

justinedelson: 1. I'm not sure what you mean by "too much". I don't think it is substantially more lines of code.
2. Not exactly (and this is why I'd suggest the first option). What I think you'd have to do is try to get the adapterManager from the bindings and store that in the instance variable. Then, in `adaptTo()` if that instance variable is null, use the service registry to get the AdapterManager, do the adaptation and then unget the service reference.

I believe the easiest way to test this would be to use `registerAdapter` inside the unit test.

I created #2230 as an alternate solution

davidjgonzalez:
Replaced by #2230

PR #2230, opened 2020-03-24
handle case where SlingBindings is null. fixes #2228

### LINKED ISSUE #2228: SlingBindings is null and results in NullPointerException in usage with @SlingServlet

justinedelson:
This is alternative to #2229


bundle/src/test/java/com/adobe/acs/commons/models/injectors/impl/ChildResourceFromRequestInjectorTest.java
```diff
@@ -34,13 +33,14 @@
...
 import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertSame;
 
 public class ChildResourceFromRequestInjectorTest {
+
+    private ActivatorHelper activatorHelper = new ActivatorHelper();
+
     @Rule
-    public final AemContext context = new AemContext(ResourceResolverType.JCR_MOCK);
+    public final AemContext context = new AemContext(activatorHelper.afterSetup(), activatorHelper.beforeTeardown(), ResourceResolverType.JCR_MOCK);
```
justinedelson: @stefanseifert can you suggest a better way to do this? Basically I need to invoke these methods whenever the context is created. This approach seems to work fine, but it seems a bit inelegant. I thought about creating a custom subclass of `AemContextImpl` but that would mean a fair bit of duplication from `AemContext`.

davidjgonzalez:
@justinedelson  we good to pull this into the next release? If @stefanseifert has ideas how to clean up the Unit Test, i expect its not a problem to clean it up later?

justinedelson:
yep. I'll be happy to get @stefanseifert 's advice at a later point. Purely about testing so easy enough to change later.

Issue #2232, opened 2020-03-25
ACS AEM Commons includes extra META-INF/vault files that are generated by plugins


davidjgonzalez:
## Required Information

* [X] AEM Version, including Service Packs, Cumulative Fix Packs, etc: N/A
* [X] ACS AEM Commons Version: 4.4.x
* [X] Reproducible on Latest? yes

### Expected Behavior

It would be nice to have clean META-INF/vault files that are always generated by the maven plugins rather than having the old copy/paste files in SCM (and i think certain values used)

PR #2234, opened 2020-03-25
Revert "Feature/2232 - Removes META-INF/vault files in SCM; Minor sonarqube fixes"

davidjgonzalez:
Reverts Adobe-Consulting-Services/acs-aem-commons#2233

PR #2235, opened 2020-03-25
fix oakpal failure due to #2232 [task]

justinedelson:
* Ensure that expected nodetypes and privileges are available to oakpal.


justinedelson:
alternative to #2234

PR #2236, opened 2020-03-26
Added ability to specify the dialog node name

kfnorbi:
Currently the cq:dialog node name is burnt in the provider logic without any possible way to override it. This way the dialogs, for example, for Editable templates (cq:design_dialog) cannot be built using this.


davidjgonzalez:
@kfnorbi code looks fine - curious what your use case is. I could see something that takes an absolute path to some other "common" cq:dialog but this is relative to the resource type.

kfnorbi:
If you would like to use Generated Dialog actively on components, you might want to define a cq:designDialog node as well but since the current implementation only supports the generation of cq:dialog nodes, it is not possible.

badvision:
Sounds like a reasonable change but there is a code climate failure and a couple of test failures that need resolving before I can merge this:

`
[ERROR]   DialogResourceProviderFactoryTest.testEnableSwitch:72 Should resolve model one dialog
[ERROR]   DialogResourceProviderFactoryTest.testResourceResolution:80 Should resolve model one dialog
`

PR #2237, opened 2020-03-26
Remove Automatic Package Replication UI

klcodanr:
I was looking at moving the UI for Automatic Package Replication out of /etc and after thinking about it, I came to the conclusion that APR doesn't really _need_ a UI. 

This pull request is to remove the UI for APR, quite a few code changes/improvements, improving test coverage, add health checks / JMX introspection and implement a new configuration method with OSGi configurations.


joerghoh:
@klcodanr I raised #2240 to change a few things in this PR

joerghoh:
Actually that's a huge PR, and affecting quite a lot of features. For example I see that *a lot* of functions in the area of Automatic package replication has been added, which is not mentioned.

@klcodanr  would it be possible to split this PR up into a multiple smaller Pull Requests?

klcodanr:
Agree @joerghoh, it looks like as well some unrelated changes crept in. I'm going to close this PR and re-open 2 more targeted PR's

PR #2238, opened 2020-03-26
Publish the Site Root Page along with the publishing page if Shared Component content modified 

lokeshvajrala:
**Problem**: Currently, the Site Root Page needs to be published separately with the content page where the shared/global component properties modified.

**Solution**: This new feature is going to activate the site root page as a reference of the page that is going to published, which triggers the cache flush of the root page and serves the latest content in the site. 
image

_Note_: The dispatcher cache flush techniques should be used along with this feature to serve the latest content across the site, like, TTL, `statfileslevel`.


lokeshvajrala:
I used the junit5 framework for unit testing the code in this PR, which caused coverage reduction of other classes that using the junit4 framework. I'm working on updating the unit test case that I added with the junit4 framework.

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesReferenceProvider.java
```diff
@@ -0,0 +1,189 @@
...
+     * @return Resource
+     */
+    private Resource getComponentResource(String resourceType, ResourceResolver resourceResolver) {
+        String[] resolverSearchPaths = resourceResolver.getSearchPath();
+
+        for (String searchPath : resolverSearchPaths) {
+            String tempResourceType = searchPath + resourceType;
+            if (resourceResolver.getResource(tempResourceType) != null) {
+                resourceType = searchPath + resourceType;
+                return resourceResolver.getResource(resourceType);
```
davidjgonzalez, line 168:
```suggestion
            Resource resource = resourceResolver.getResource(searchPath + resourceType);
            if (resource != null) {
                return resource;
```

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesReferenceProvider.java
```diff
@@ -0,0 +1,189 @@
...
+     *
+     * @param page {@link Page}
+     * @return page title | name
+     */
+    private String getReferenceName(Page page) {
+        String title = null;
+        if (page != null) {
+            title = page.getTitle() != null ? page.getTitle() : page.getName();
+        }
+        return title;
```
davidjgonzalez, line 185:
```suggestion
        if (page != null) {
            return StringUtils.defaultIfEmpty(page.getTitle(), page.getName());
        } else {
           return null;
        }
```

davidjgonzalez: little more easy to read.

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesReferenceProvider.java
```diff
@@ -0,0 +1,189 @@
...
+ * this adds the root page to the list of references that needs to be published.
+ */
+@Component(service = ReferenceProvider.class,
+        configurationPolicy = ConfigurationPolicy.REQUIRE)
+public class SharedComponentPropertiesReferenceProvider implements ReferenceProvider {
+
+    @org.osgi.service.component.annotations.Reference
+    private PageRootProvider pageRootProvider;
+
+    private Set<String> componentResourceTypes = new HashSet<>();
```
davidjgonzalez, line 52: Since this is an OSGi service, I dont think you want this here (and if you did, youd need a concurrent hash map).

IIUC this feature correctly, the componentResourceTypes should be collected for the page being activated, each time. Since OSGi components are singletons, this componentResourceTypes set would be shared across all uses of this service.

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesReferenceProvider.java
```diff
@@ -0,0 +1,189 @@
...
+
+        ResourceResolver resourceResolver = resource.getResourceResolver();
+        for (String resourceType : componentResourceTypes) {
+
+            Resource componentResource = getComponentResource(resourceType, resourceResolver);
+            if (componentResource == null) {
+                continue;
+            }
+
+            if (!componentResource.getResourceType().equalsIgnoreCase("cq:Component")) {
```
davidjgonzalez, line 132: Would this be better to check if the `jcr:primaryType` .. while the primary type IS the fall back for a non-existent sling:resourceType, you technically could have a sling:resourceType on a cq:Component for some "weird" use case, and looking for the jcr:primaryType is simply more clear since that's what you actually want to check (i think).

lokeshvajrala: Updated the check, and it looks cleaner now. Thanks!

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesReferenceProvider.java
lokeshvajrala: Thank you for pointing out this, removed the class level field.

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesReferenceProvider.java
lokeshvajrala: oh yeah, updated it. 👍

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesReferenceProvider.java
lokeshvajrala: Updated the code with the suggestions. Thank you for reviewing it.

bundle/src/test/resources/com/adobe/acs/commons/wcm/shared/impl/shared-properties-reference-provider/apps-content.json
```diff
@@ -0,0 +1,659 @@
...
+  "content": {
+    "text": {
+      "jcr:primaryType": "cq:Component",
+      "jcr:createdBy": "admin",
+      "jcr:title": "Text (v2)",
+      "jcr:created": "Tue Oct 15 2019 14:00:38 GMT-0600",
+      "jcr:description": "Rich Text Section",
+      "cq:icon": "textLeft",
+      "componentGroup": ".core-wcm",
+      "cq:dialog": {
```
HitmanInWis, line 221: I know the details of the `cq:dialog` are not hurting anything, but are they providing any necessary function for the test?  If not, my vote would be to omit those details since its a ton of irrelevant stuff to wade through and represents a cut in time of what this component looks like which will over time become obsolete.

HitmanInWis: This is a general comment - would like to see the content of this file largely slimmed down to remove irrelevant nodes.

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesReferenceProvider.java
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    /**
+     * Get the page title prop if it's null get the name
+     *
+     * @param page {@link Page}
+     * @return page title | name
+     */
+    private String getReferenceName(@Nonnull Page page) {
+        return StringUtils.defaultIfEmpty(page.getTitle(), page.getName());
```
HitmanInWis, line 188: Worth maybe using `defaultIfBlank()` as extra safety?

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesReferenceProvider.java
```diff
@@ -0,0 +1,192 @@
...
+    /**
+     * Returns the Component Resource by searching the configured resolver search paths ["/libs", "/apps"]
+     * @param resourceType {@link String}
+     * @param resourceResolver {@link ResourceResolver}
+     * @return Resource
+     */
+    private Resource getComponentResource(String resourceType, ResourceResolver resourceResolver) {
+        String[] resolverSearchPaths = resourceResolver.getSearchPath();
+
+        for (String searchPath : resolverSearchPaths) {
```
HitmanInWis, line 172: Should we add logic here to see if `resourceType` starts with `/` and if so then fetch directly rather than by appending search path?  Though it is standard practice to use a relative path `sling:resourceType` such as `mysite/components/mything` the absolute path of `/apps/mysite/components/mything` generally works.

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesReferenceProvider.java
```diff
@@ -0,0 +1,192 @@
...
+
+    @Override
+    public List<Reference> findReferences(Resource resource) {
+        List<Reference> references = new ArrayList<>();
+
+        if (!resource.getPath().startsWith("/content")) {
+            return references;
+        }
+
+        ResourceResolver resourceResolver = resource.getResourceResolver();
```
HitmanInWis, line 62: Im curious if we should be using a service user for this, to ensure proper permissions to read everything that needs to be read.

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesReferenceProvider.java
```diff
@@ -0,0 +1,192 @@
...
+                .getResource(componentResource.getPath() + "/dialogglobal");
+            if (sharedDialog == null && globalDialog == null) {
+                continue;
+            }
+
+            Page rootPage = pageRootProvider.getRootPage(resource);
+            if (rootPage != null) {
+                references.add(new Reference("sharedProperties", getReferenceName(rootPage),
+                    rootPage.getContentResource(),
+                    rootPage.getLastModified() != null ? rootPage.getLastModified()
```
HitmanInWis, line 154: The wrapping of this ternary is quite confusing, b/c `.getTimeInMillis()` splits to the next line.

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesReferenceProvider.java
```diff
@@ -0,0 +1,192 @@
...
+        return null;
+    }
+
+    /**
+     * Get the page title prop if it's null get the name
+     *
+     * @param page {@link Page}
+     * @return page title | name
+     */
+    private String getReferenceName(@Nonnull Page page) {
```
HitmanInWis, line 187: Technically speaking, could probably eliminate this function since the entire logic is a one-liner and is never repeated.  However, I'm cool if you think this is best.

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesReferenceProvider.java
HitmanInWis: Could probably make the argument that if the user doesnt have permission to read something in here, they couldn't have edited a related shared property, so _maybe_ this is fine?  Generally speaking i would expect a service user for this type of operation rather than depending on the current user's permissions.

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesReferenceProvider.java
```diff
@@ -0,0 +1,192 @@
...
+                .getResource(componentResource.getPath() + "/dialogshared");
+            Resource globalDialog = componentResource.getResourceResolver()
+                .getResource(componentResource.getPath() + "/dialogglobal");
+            if (sharedDialog == null && globalDialog == null) {
+                continue;
+            }
+
+            Page rootPage = pageRootProvider.getRootPage(resource);
+            if (rootPage != null) {
+                references.add(new Reference("sharedProperties", getReferenceName(rootPage),
```
HitmanInWis, line 152: Before adding the reference, do we have to validate that the current user has the ability to publish the root page?  Or is that handled outside of this function by the general framework that identifies references?

PR #2240, opened 2020-03-28
Avoid setters on injected services

joerghoh:
There is no need to have setters on injected fields, as AemContext can handle this for us.
I had to rewrite the unittests to make that work, but IMHO they look much nicer now :-)


klcodanr:
Thanks @joerghoh !

PR #2243, opened 2020-03-30
Adding a servlet to handle Robots.txt

shsteimer:
@davidjgonzalez inspired by our slack conversation the other day.

I considered adding support to crawl the page tree and generate allow/deny rules by page props, and might still add that if anyone thinks it's needed, but this seemed sufficient for now, and to judge if there is even interest in having this here.  Assuming there is, I will work on and submit a PR to doc the feature.

As it exists, the servlet works in one of two ways.
1) read the contents from either a relative or absolute property.  See the unit tests for examples of this.  Technically any String or Binary property should work, but in most cases this would either be a page property on the site root page, or the jcr:data property of an nt:file e.g. the original rendition of a dam asset.
2) by specifying the contents entirely via osgi configuration.  Those configurations are grouped by name, such that the allow/deny rules can be associated to the proper user-agent rules.

Note: the servlet must be addressed on the page's jcr:content node, e.g. /content/we-retail/us/en/_jcr_content.robots.txt. I'm unsure of why this is the case since it is not necessary for the sitemap servlet, but my best guess is something related to how AEM/Sling handles xml extensions on pages vs txt.   If anyone knows of a way to resolve that, LMK.

I don't think that's a deal breaker since the file has to be served from the site root anyway (will be sure to document the RewriteRule to do that), so the internal AEM path is somewhat irrelevant.

Update: per what @joerghoh noted in his comments, I now added the ability to specify page property names for generating the allow, disallow, and sitemap directives.


joerghoh:
@shsteimer you need to bind your servlet to the resourcetype "cq/page", then page.robots.txt will be resolved properly.

joerghoh:
Regarding the configuration: From what I understand you always need to create an OSGI configuration when you rollout a new domain, which typically requires a deployment. I don't like that approach, because setting up a new site is typically done completely on author side and should not require admin intervention (deployment).

I would love if some general settings are defined via OSGI, but all domain-rollout related aspects on AEM side can be purely handled by business and does not need any change outside of /content.

shsteimer:
@joerghoh 

Assuming all your pages use the same resource type, that should work if you specify a relative property for `robots_content_property_path`

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/RobotsServlet.java
```diff
@@ -0,0 +1,418 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+
+@Component(service = Servlet.class, immediate = true, property = {
```
davidjgonzalez, line 68:
```suggestion
@Component(service = Servlet.class, property = {
```
Shouldn't need to be immediate=true

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/RobotsServlet.java
```diff
@@ -0,0 +1,418 @@
...
+        "sling.servlet.selectors=robots",
```
davidjgonzalez, line 69: Should this be bound ot a resource type? For example `cq/Page` or something?

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/RobotsServlet.java
```diff
@@ -0,0 +1,418 @@
...
+
+    private void writeFromOsgiConfig(SlingHttpServletRequest request, SlingHttpServletResponse response) throws IOException {
+        PrintWriter writer = response.getWriter();
+
+        rules.getGroups().forEach(group -> writeGroup(group, request, writer));
+
+        rules.getSitemaps().stream().map(sitemap -> buildSitemapString(sitemap, request.getResourceResolver())).forEach(writer::println);
+        if(!rules.getSitemapProperties().isEmpty()) {
+            PageManager pageManager = request.getResourceResolver().adaptTo(PageManager.class);
+            Page page = pageManager.getContainingPage(request.getResource());
```
davidjgonzalez, line 127: How do we know this will always be a page? Isnt this simply bound to some selectors/extension?

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/RobotsServlet.java
```diff
@@ -0,0 +1,418 @@
...
+        }
+    }
+
+    private void writeGroup(RobotsRuleGroup group, SlingHttpServletRequest request, PrintWriter writer) {
+        if (printGroupingComments) {
+            writer.println("# Start Group: " + group.getGroupName());
+        }
+
+        PageManager pageManager = request.getResourceResolver().adaptTo(PageManager.class);
+        Page page = pageManager.getContainingPage(request.getResource());
```
davidjgonzalez, line 138: Are we sure this is not null?

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/RobotsServlet.java
```diff
@@ -0,0 +1,418 @@
...
+        if (!robotsContentsPropertyPath.startsWith("/")) {
+            robotsContentsPropertyPath = request.getResource().getPath() + "/" + robotsContentsPropertyPath;
+        }
+
+        Session session = request.getResourceResolver().adaptTo(Session.class);
+        boolean written = false;
+        try {
+            if (session.itemExists(robotsContentsPropertyPath)) {
+                Item item = session.getItem(robotsContentsPropertyPath);
+                if (item instanceof Property) {
```
davidjgonzalez, line 192: Should we log a WARN if this points to something that isn't a property/string/binary?

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/RobotsServlet.java
shsteimer: it gets bound based on the resource type specified in the configuration

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/RobotsServlet.java
shsteimer: The (or at least my) expectation, is that the config would bind it to a resource type of the page component, and it would be addressed as /path/to/page/_jcr_content.robots.txt. It is I support possible it would be bound some other resource type, but that isn't really a case I have in mind to support.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/RobotsServlet.java
shsteimer: See above for expected usage pattern. I suppose I can add null checks and return a 400 bad request if that somehow happens

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/RobotsServlet.java
shsteimer: added some additional logging to make this less ambiguous.

PR #2242, opened 2020-03-30
Ensure APR Service User

klcodanr:
Fixes #2241: Leveraging ensure service user to ensure that the service user for the automatic package replicator exists and has sufficient permissions

PR #2246, opened 2020-04-02
Remove Automatic Package Replication UI

klcodanr:
I was looking at moving the UI for Automatic Package Replication out of /etc and after thinking about it, I came to the conclusion that APR doesn't really need a UI.

This pull request is to:

 - Remove the custom UI for the Automatic Package Replicator
 - Change the APR configurations under /etc to 3 OSGi Configs:
     - Scheduled APR
     - Event-Based APR
     - Filtered Event-Based APR
     Note that the filtered event-based is required as OSGi seems to interpret an unset event.filter as not matching any events.
 - Updating the API / implementation (including in JMX) to reflect this)
 - Bonus - including a JMX method to migrate from the old configuration to new
 - Removing the no longer required ConfigurationUpdateListener and ResourceChangeManager services
 - Adding a health check to monitor APR replication statuses
 - Adding Unit Tests (yay!)
 - Including #2242 since otherwise, it's hard to test

I know it's a lot, but at least IMO this is a much easier to use and maintain option as compared to the node-based option in /etc.


davidjgonzalez:
@klcodanr how backwards compat is this? ... Would this warrant a major version bump per semantic versioning rules?

klcodanr:
@davidjgonzalez I included a JMX MBean to migrate the old configurations into the new, but it has to be manually triggered. IMO a new major version would seem appropriate. 

I'm not sure why the Travis CI status hasn't updated but it does seem like the build was successful: https://travis-ci.org/github/Adobe-Consulting-Services/acs-aem-commons/builds/700889666

PR #2247, opened 2020-04-02
Fix #2245 - Marketo Endpoint Protocol

### LINKED ISSUE #2245: Error description message in Marketo plugin
I notice that the description message in Marketo configuration shows an example with "https"
However the backend code also append a https to the front.

image

image

klcodanr:
As #2245 describes the tooltip for the Marketo configuration indicates that the endpoint should start with https:// but the client prepends https://. This PR adjusts the logic to strip https:// if it is provided so the code should work whether or not the protocol is provided in the endpoint URL.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 <!-- Keep this up to date! After a release, change the tag name to the latest release -->
 [unreleased changes details]: https://github.com/Adobe-Consulting-Services/acs-aem-commons/compare/acs-aem-commons-4.3.2...HEAD
 
+### Fixed
+ - #2245 Marketo Endpoint Protocol Documentation Issue
+
 ## [4.5.0] - 2020-03-25
 
 ### Added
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 [unreleased changes details]: https://github.com/Adobe-Consulting-Services/acs-aem-commons/compare/acs-aem-commons-4.3.2...HEAD
 
 ### Fixed
+- #2245 Marketo Endpoint Protocol Documentation Issue
```
davidjgonzalez, line 13:
```suggestion
- #2245 - Marketo Endpoint Protocol Documentation Issue
```

bundle/src/main/java/com/adobe/acs/commons/marketo/impl/MarketoClientConfigurationImpl.java
```diff
@@ -80,6 +80,9 @@ public String getClientSecret() {
 
   @Override
   public String getEndpointHost() {
+    if(endpointHost.startsWith("https://")){
+      return endpointHost.substring(8);
```
davidjgonzalez, line 84:
```suggestion
      return endpointHost.substring("https://".length());
```
Just a but more clear than the magic numbe `8`

PR #2252, opened 2020-04-07
Make granite workitem comment available to email template 

drewrobinson:
Make granite workitem comment available to email template as: <code>{wfStepComment}</code>

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/email/process/impl/SendTemplatedEmailProcess.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/email/process/impl/SendTemplatedEmailProcess.java
@@ -274,6 +274,10 @@ public class SendTemplatedEmailProcess implements WorkflowProcess {
                     .getTitle());
             // Set workflow initiator
             wfParams.put(SendTemplatedEmailConstants.WF_INITIATOR, workItem.getWorkflow().getInitiator());
+            
+            if(workItem.getMetaDataMap().containsKey("comment")) {
+                wfParams.put(SendTemplatedEmailConstants.WF_STEP_COMMENT, workItem.getMetaDataMap().get("comment").toString());
+            }
         } catch (Exception e) {
             log.warn("Error getting workflow title and workflow step title {}", e);
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 
 ### Added
 - #2215 - Added Parameterized granite include to support generic dialog snippets
+- #2252 - Make comment available as email template variable
 
 ### Fixed
 - #2225 /etc/designs/acs-aem-commons no longer readable by everyone in AEM 6.4+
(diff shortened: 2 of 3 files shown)


bundle/src/main/java/com/adobe/acs/commons/email/process/impl/SendTemplatedEmailProcess.java
```diff
@@ -274,6 +274,10 @@ public final void execute(WorkItem workItem, WorkflowSession workflowSession, Me
                     .getTitle());
             // Set workflow initiator
             wfParams.put(SendTemplatedEmailConstants.WF_INITIATOR, workItem.getWorkflow().getInitiator());
+            
+            if(workItem.getMetaDataMap().containsKey("comment")) {
+                wfParams.put(SendTemplatedEmailConstants.WF_STEP_COMMENT, workItem.getMetaDataMap().get("comment").toString());
```
godanny86, line 279: @drewrobinson I'm curious if maybe it makes sense to just shove the entire workItem.getMetadataMap() into the `wfParams` map. Similar to how the payload properties are converted to a `<String, String>` i.e 

```java
Map<String, String> workItemMetadata = SendTemplatedEmailUtils.getWorkItemMetadata(workItem, sdf);
            if (workItemMetadata != null) {
                emailParams.putAll(workItemMetadata);
            }
```
I think you may need to prefix the key with something like `workItem-` to ensure that you don't accidentally override a payload property key... but then again maybe this is overkill...

https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/email/process/impl/SendTemplatedEmailUtils.java#L69

drewrobinson: Great thought, @godanny86. I like that idea a lot and appreciate the feedback. I'll get started on the improvements.

davidjgonzalez: @drewrobinson Are you planning on coming up w/ a way to move all metadata properties over?

Also, IIRC,`putAll(..)` wont work due to the values of the metadata map not being actual Strings, but rather those weird Value proxy objects (i think we ran into this when you initially tried the templated email stuff?)

godanny86: its true, you would need to introduce an additional method to convert the metadataMap to a simple HashMap <String,String>. However you should be able to re-use some of the methods in  https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/email/process/impl/SendTemplatedEmailUtils.java to make it easier

davidjgonzalez:
@drewrobinson ill merge this since you've defined a namespaced (`wfStep`) for WF Step variables.. so this should position you nicely to add other WF Step variables to the map as you need later. 

If you do take this on - Carefully handle the various data-types that can be passed through though, vs doing some generic blanket serialization to Stings.

drewrobinson:
@davidjgonzalez @godanny86 Great recommendations - sounds great. I absolutely agree. This addresses my immediate use case and sets up the next steps beautifully. I appreciate your help with this.

PR #2258, opened 2020-04-10
Make the logging more uniform

joerghoh:
* fixes a lot of instances of the sonar issue "java:S3457"
* fixed a few instances of the "log-and-throw" exception handling type


bundle/src/main/java/com/adobe/acs/commons/httpcache/invalidator/HttpCacheInvalidationJobConsumer.java
```diff
@@ -95,7 +95,7 @@ public JobResult process(final Job job) {
             invalidateReferences(path);
         }
 
-        log.trace("Invalidation job for the path processed.", path);
+        log.trace("Invalidation job for the path '[}' processed.", path);
```
davidjgonzalez, line 98:
```suggestion
        log.trace("Invalidation job for the path [ {} ] processed.", path);
```

PR #2259, opened 2020-04-11
improve logging of FileFetcher

joerghoh:


bundle/src/main/java/com/adobe/acs/commons/filefetch/impl/FileFetcherImpl.java
```diff
@@ -122,7 +122,7 @@ public void run() {
       lastJobSucceeded = true;
       lastException = null;
     } catch (Exception e) {
-      log.warn("Failed to run fetch file job for {}", config.remoteUrl(), e);
+      log.warn("Failed to run fetch file job for {}: {}", config.remoteUrl(), e.getMessage());
```
davidjgonzalez, line 125: @joerghoh why not the stack trace, since this exception is swallowed here otherwise.

joerghoh: I don't really think that the stacktrace would help here, as the locations where the exception actually happened is quite obvious.
But I agree, in case there is a discussion if a stacktrace would be usefull, print it :-) Will update the PR.

PR #2263, opened 2020-04-14
#2261 update test dependencies 

### LINKED ISSUE #2261: Update to latest mocking libraries
Some new versions of the sling-mocks and aem-mocks are out, we should update. The simple approach did not really work, so it isn't contained in #2260.

PR: #2263

joerghoh:
updated some major test dependencies to newer versions.

Notable changes: 
* I needed to downgrade guava to 18, because it was required by Oak. SlingMocks new comes with Oak 1.6 and something was apparently changed. Should follow up and find if there's a way to deal with that dependency within SlingMocks.
* Requires Maven 3.5 as minimum version.


kwin:
@joerghoh What happened to this PR? I need newer test dependencies as I require `aem-mocks 3.0+` for https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/2413. This requires updating `org.apache.sling.testing.sling-mock` as well.

kwin:
Superseded by #2700.

Issue #2265, opened 2020-04-15
Review ResourceChangeListeners 


joerghoh:
I came across the behavior that any ResourceChangeListener is notified when a servlet is unregistered/registered (more precise: when their corresponding ResourceProvider is unregisted/registered) if the property ```resource.change.types``` is not defined. This happens quite often during deployments.
Depending on the implementation of the ResourceChangeListeners this can have a huge performance impact on deployments.

This issue is to validate all implementations of this service interface and check if they need to handle such (un-) registrations or if they are searching for simple resource changes in the repository.

joerghoh:
If enabled JcrNodeChangeEventHandler (part of the httpcache) creates a new (Sling) job on stop/start of a Servlet. This is not necessary.

joerghoh:
The ConfigurationUpdateListener of the Automatic Package Replicator feature refreshes its cache also on every start/stop of a servlet, which isn't necessary.

joerghoh:
SystemNotificationsImpl has this property defined and does not need any change.

joerghoh:
fix in #2268

joerghoh:
commit merged, thanks!

PR #2268, opened 2020-04-18
#2265 do not listen to changes of ResourceProviders

joerghoh:


bundle/src/main/java/com/adobe/acs/commons/httpcache/invalidator/event/JCRNodeChangeEventHandler.java
```diff
@@ -72,6 +72,9 @@
                   description = "List of paths to watch. Entries with the 'glob:' prefix are interpreted as globs, "
                           + "i.e. the * and ** wildcards are supported.",
                   name = ResourceChangeListener.PATHS),
+        @Property(label = "Type of change to listent to",
```
davidjgonzalez, line 75:
```suggestion
        @Property(label = "Type of change to listen to",
```

PR #2274, opened 2020-04-23
Update/add plugins for release with OSSRH

### LINKED ISSUE #2273: Publish to Maven Central instead of Adobe Repo
Similarly to Core WCM Components the binary artifacts should be pushed to Maven Central (https://github.com/adobe/aem-core-wcm-components/issues/356) instead of only the Adobe Repo. The time delay is much less, and it works flawlessly for thousand other open source tools. For details refer to https://central.sonatype.org/pages/ossrh-guide.html.

kwin:
This fixes #2273


RELEASING.md
```diff
@@ -17,14 +17,28 @@ The password is your API key (not your password), which you can find via https:/
...
 
+These credentials are used to deploy to [OSSRH][OSSRH].
+
+In addition you need to setup [GPG](https://central.sonatype.org/pages/working-with-pgp-signatures.html) to create OpenPGP signatures. After installing https://www.gnupg.org/download/ you need to create key pair (if you don't have one yet) and make sure that the public key is distributed via hkp://pool.sks-keyservers.net.
+
+It is recommended that your private key is protected with a passphrase. You can persist the passphrase in the settings.xml as well
+
+     <server>
+         <!-- has the passphrase for the gpg signing in encrypted format: http://maven.apache.org/plugins/maven-gpg-plugin/sign-mojo.html#passphraseServerId -->
+         <id>gpg.passphrase</id>
```
davidjgonzalez, line 36: @kwin  is this right? Wouldn't this just be a `<properties>` in a ossrh `<profile>`?

kwin: GPG and OSSRH deployment are not directly related. This is for storing your GPG passphrase, otherwise you have to explicitly parameterize the mvn release goals. You should not use a simple property as that cannot leverage the encryption (http://maven.apache.org/guides/mini/guide-encryption.html). As servers cannot be given below a profile I don't really get your remark...

PR #2276, opened 2020-04-27
issue #2187 upgrade to oakpal 2.0.0

### LINKED ISSUE #2187: Transitive dependencies on Jackrabbit Oak
The project directly uses certain Oak APIs, but does not itself declare these dependencies; these are all transitive dependencies via oakpal.

This has several problems:

- updating oakpal might break the build if transitive dependencies change
- no control about what version of Oak is used

In particular: oak-core-spi will have an incompatible API change as of Oak 1.26.0 (https://issues.apache.org/jira/browse/OAK-7358). Acs-aem-commons will not work with that Oak version, unless the import range for org.apache.jackrabbit.oak.spi.security.principal is relaxed to allow version 2.0.0 (or the whole project is updated to build against Oak 1.26.0).

adamcin:
fixes #2187 .

Important changes:
* Oakpal 2.0.0 extracts ProgressCheck-related API classes into a new oakpal-api dependency that is not encumbered with oak-core and oak-core-spi dependencies. The `oakpal-checks` module now depends on `oakpal-api` with `scope=compile` and `oakpal-core` with `scope=test`, which is necessary because we actually need to run oakpal, and therefore oak, to perform our unit tests.
* I've taken the initiative to remove the CompositeStoreAlignment check from acs-aem-commons, and adopt it as part of the oakpal basic checklist. It is fundamentally tightly coupled to oak-core-spi classes related to managing `Mount`s for composite stores, and it seems to be configurable enough that it could be useful for non-AEM oak application targets compatible with filevault. I've retained the ACS Commons Bundle copyright header on the file.


adamcin:
@justinedelson This change might also eliminate the need for oakpal exceptions to the `CQRules:CQBP-84--dependencies` Cloud Manager Sonar rule.

adamcin:
@badvision @davidjgonzalez This is ready for review.

oakpal-checks/pom.xml
```diff
@@ -266,27 +266,38 @@
     </profiles>
 
     <dependencies>
+        <dependency>
+            <groupId>org.jetbrains</groupId>
```
davidjgonzalez, line 270: @adamcin what's this dep for? Is this IDEA specific?

adamcin: It's just for `@NotNull` and `@Nullable`. Not IDEA specific, necessarily, but I haven't tested it with Eclipse, for example. Anyways, I chose this for nullability annotations purely because it's what is used in the jackrabbit-oak and filevault projects and it works for me (/shrug). I think jackrabbit switched away from the javax annotations to avoid the runtime package dependency and the whole JEE trademark mess.

Issue #2282, opened 2020-05-01
Do not start services on AEMaaCS instances if they are not working there


joerghoh:
In the documentation a number of features are listed already as "do not use them when you are deploying on AEM as a CloudService".

Right now this is not technically enforced, but with the RequireAem feature (see https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/b391b2455bb1aac018183ae4b76794379ebb8a59/bundle/src/main/java/com/adobe/acs/commons/util/RequireAem.java) it can be enforced quite easily.

Right now these features are tagged as "do not run within AEMaaCS":
```
bulk-workflow-manager
ensure-oak-index
healthchecks/smtp
healthchecks/status-emailer
healthchecks
json-event-logger
versioned-clientlibs
workflow-processes/assets-rendition-matter
workflow-processes/assets-watermark-process
workflow-processes/brand-portal
file-fetcher
mcp-tools/asset-ingestion/asset-ingestor
mcp-tools/asset-ingestion
mcp-tools/asset-ingestion/s3-asset-ingestor
mcp-tools/asset-ingestion/url-asset-ingestor
mcp-tools/refresh-folder-thumbnails
redirect-map-manager
remote-assets
```
(I have extracted this list from the official documentation)

joerghoh:
see #2285

joerghoh:
#2285 has been merged, closing this one.

PR #2285, opened 2020-05-02
#2282 disable features on AEM as a CloudService

joerghoh:


royteeuwen:
Sad that the versioned clientlibs does not work, it's a really handy one for javascript invalidation. Can this one be fixed or is it inherently bad design for AEM aaC?

davidjgonzalez:
@royteeuwen this should be OOTB now in AEMaaCS.

joerghoh:
Well, some testing is still outstanding :-) Plan to work on it later today on a quickstart instance and a AEMaaCS SDK instance.

badvision:
@joerghoh did you want to hold off on the merge until after you do more testing?

joerghoh:
Yes, I do. Please don't merge until further notice.

joerghoh:
I tested locally with AEM 6.5.2 and the listed components are active. After overcoming some issues unrelated to this change I was able to confirm, that these components are not active on a recent SDK version.

So all fine, we can go on and merge.

davidjgonzalez:
@joerghoh any idea if these are marked as incompatible on the ACS Commons docs site? I wonder if we should change the wording on these to state that they won't start on AEM CS (vs. saying they are incompatible, so don't use them.... now you *cant* use them)

joerghoh:
https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io/pull/195

PR #2294, opened 2020-05-09
Feature/2293 - Copy Process workflow process

davidjgonzalez:
WF Step that copies properties from one location to another.

1. Paths are relative to the payloads resolved dam:Asset or cq:Page resource
1. Paths can be absolute as well (weird, but possible)
1. Params are in the format: `<source property> -> <destination property>` (`->` acts as a semantic delimiter; to be clear, `<-` is not supported. To do that, you'd simply flip the properties)
1. This emits a event-user-data of `acs-aem-commons.workflow.copy-properties` to help configure launches to prevent cyclic behavior.

image


bundle/src/test/resources/com/adobe/acs/commons/workflow/process/impl/CopyPropertiesProcessTest.json
```diff
@@ -0,0 +1,35 @@
+{
+  "one.png" : {
```
bstopp, line 2: You should name the test resources something that correlates to the test you are validating - so that it is clear what is impacted by changing the test data.
```suggestion
  "copyProperties.png" : {
```

ui.apps/src/main/content/jcr_root/apps/acs-commons/components/workflow/copy-properties/_cq_dialog/.content.xml
```diff
@@ -0,0 +1,114 @@
...
+                                    jcr:primaryType="nt:unstructured"
+                                    jcr:title="Properties to Copy"
+                                    sling:resourceType="granite/ui/components/coral/foundation/form/fieldset">
+                                <items jcr:primaryType="nt:unstructured">
+                                    <skip-empty-property
+                                            jcr:primaryType="nt:unstructured"
+                                            sling:resourceType="granite/ui/components/coral/foundation/form/checkbox"
+                                            text="Skip empty source properties"
+                                            fieldDescription="If the source property is empty or missing, leave the destination property alone."
+                                            named="./metadata/SKIP_EMPTY_SOURCE_PROPERTY"
```
bstopp, line 83: Is this the new convention, constant-esq metadata properties?

bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/CopyPropertiesProcess.java
```diff
@@ -0,0 +1,162 @@
...
+        } catch (RepositoryException e) {
+            log.warn("Unable to set user-data to [ " + EVENT_DATA + " ]", e);
+        }
+    }
+
+    protected void copyProperties(MetaDataMap metaDataMap, ResourceResolver resourceResolver, String payload) {
+        final Resource resource = workflowHelper.getPageOrAssetResource(resourceResolver, payload);
+
+        if (resource == null) {
+            log.debug("Could not process payload [ {} ] as it could be resolved to a Page or Asset", payload);
```
bstopp, line 97: Debug, not warn?

ui.apps/src/main/content/jcr_root/apps/acs-commons/components/workflow/copy-properties/_cq_dialog/.content.xml
davidjgonzalez: This allows you to the WF Dialogs to control the input of your metaData properties; vs. PROCESS_ARGS which are the "primitive" blobs... so yeh. I've mostly not don't this before cuz I'm lazy, but i'd argue this is the "right" way to do it, and planning on using this pattern going forward.

One thing worth noting here, is the `.execute(..., MetadadaMap map)` map variable, is not your typical valuemap that can look down the tree; it can ONLY see values directly on the metadata node, which is why this property mapping isn't modeled as sub-nodes, with individual fields for source and dest.

bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/CopyPropertiesProcess.java
```diff
@@ -0,0 +1,162 @@
...
+        } catch (RepositoryException e) {
+            throw new WorkflowException(String.format("Could not find the payload for '%s'", wfPayload), e);
+        }
+    }
+
+    protected void setJcrSessionUserData(ResourceResolver resourceResolver) {
+        try {
+            resourceResolver.adaptTo(Session.class).getWorkspace().getObservationManager().setUserData(EVENT_DATA);
+        } catch (RepositoryException e) {
+            log.warn("Unable to set user-data to [ " + EVENT_DATA + " ]", e);
```
joerghoh, line 89: please use 
```
log.warn("Unable to set user-data to [ {} ]", EVENT_DATA, e);
```

instead string concatenation (much more readable).

bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/CopyPropertiesProcess.java
```diff
@@ -0,0 +1,162 @@
...
+            propertyMap = StringUtils.replace(propertyMap, ALTERNATE_SEPARATOR, SEPARATOR);
+            Map.Entry<String, String> entry = ParameterUtil.toMapEntry(propertyMap, SEPARATOR);
+
+            try {
+                PropertyResource source = new PropertyResource(StringUtils.trim(entry.getKey()), resource.getPath(), resourceResolver);
+                PropertyResource destination = new PropertyResource(StringUtils.trim(entry.getValue()), resource.getPath(), resourceResolver);
+
+                destination.setValue(source.getValue(), skipEmptyValue);
+            } catch (WorkflowException e) {
+                log.error("Could not copy properties", e);
```
joerghoh, line 114: Can you add source and desination resource to that message? Otherwise it's hard to debug.

bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/CopyPropertiesProcess.java
```diff
@@ -0,0 +1,162 @@
...
+
+        for (String propertyMap : propertyMaps) {
+            propertyMap = StringUtils.replace(propertyMap, ALTERNATE_SEPARATOR, SEPARATOR);
+            Map.Entry<String, String> entry = ParameterUtil.toMapEntry(propertyMap, SEPARATOR);
+
+            try {
+                PropertyResource source = new PropertyResource(StringUtils.trim(entry.getKey()), resource.getPath(), resourceResolver);
+                PropertyResource destination = new PropertyResource(StringUtils.trim(entry.getValue()), resource.getPath(), resourceResolver);
+
+                destination.setValue(source.getValue(), skipEmptyValue);
```
joerghoh, line 112: I miss the ```resourceResolver.commit()``` statement.

bundle/src/main/java/com/adobe/acs/commons/util/impl/WorkflowHelperImpl.java
```diff
@@ -48,7 +47,6 @@
 import java.util.Map;
 
 @Component
-@Service
```
joerghoh, line 51: Unfortunately it's not obvious if this change comes with the PR or is caused by the merge/rebasing :-(

bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/CopyPropertiesProcess.java
davidjgonzalez: I believe Workflow Engine will persist changes as needed; is this not the case?

bundle/src/main/java/com/adobe/acs/commons/util/impl/WorkflowHelperImpl.java
davidjgonzalez: I changed this file to use OSGi annotations (instead of Felix) therefor the service annotations are not required anymore (implicit via the interface). I can add it to make this more clear though.

I changed this file because locally I was having problems w Felix scr annotated services no generating OSGi-INF metadata XML files. Im not sure if other's are experiencing the same?

bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/CopyPropertiesProcess.java
joerghoh: Actually I don't think so. Maybe you should test it inside a real AEM instance and not only using the mocking libraries :-)

davidjgonzalez: Just for posterity; AEM WF Engine does infect commit the WF's session for you.

Issue #2271, opened 2020-04-20
Environment Indicator setting causes Contexthub calls to fail


pegr69:
## Required Information

* {] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.4.6
* {] ACS AEM Commons Version: 4.5.0
* [] Reproducible on Latest? yes, always.

### Expected Behavior
http://localhost:4502/etc/cloudsettings.kernel.js/libs/settings/cloudsettings/legacy/contexthub
Should return the contexthub configuration or a 304 no content if already cached.

_Please describe briefly the expected behavior, i.e. when I do X, Y should happen._

### Actual Behavior
Every OTHER call return an error 500.
If browser is in dev mode(F12), and no cache enabled,  all calls are fine since the browser always request a full response.

### Steps to Reproduce

1, Set Enviroment Indicator setting color to ex: "blue", save.
2, Call contexthub URL: http://localhost:4502/etc/cloudsettings.kernel.js/libs/settings/cloudsettings/legacy/contexthub

- First call response is OK, but the following error message appears in the log:
`20.04.2020 14:34:54.617 *WARN* [192.168.237.1 [1587386094576] GET /etc/cloudsettings.kernel.js/libs/settings/cloudsettings/legacy/contexthub HTTP/1.1] com.adobe.acs.commons.util.BufferedServletResponse Prevent committing the response, it will be committed deferred, i.e. once this buffered response`

- Second call causes error 500 in response and the following ends up in the error.log:

`20.04.2020 14:35:39.919 *ERROR* [qtp190519350-7502] org.apache.felix.http.jetty Exception while processing request to /etc/cloudsettings.kernel.js/libs/settings/cloudsettings/legacy/contexthub (java.io.IOException: org.eclipse.jetty.http.BadMessageException: 500: Content for no content response)
java.io.IOException: org.eclipse.jetty.http.BadMessageException: 500: Content for no content response
        at org.eclipse.jetty.server.HttpOutput.write(HttpOutput.java:225) [org.apache.felix.http.jetty:4.0.6]
        at org.eclipse.jetty.server.HttpOutput.flush(HttpOutput.java:397) [org.apache.felix.http.jetty:4.0.6]
        at org.eclipse.jetty.server.Response.flushBuffer(Response.java:1233) [org.apache.felix.http.jetty:4.0.6]
[...]

PR #2295, opened 2020-05-10
#2271 bugfix for contexthub request

joerghoh:
It seems that Jetty does some internal validation on the PrintWriter, and you are allowed to invoke flushBuffer() only if there is some body at all. This situation was affecting the AemEnvironmentIndicatorFilter only when then response did not contain a "html" in the content-type and had an empty body (which is indeed a rare situation).


pegr69:
Thx, Joerg!, 
I'll have look tomorrow on our staging servers.

/Peter

pegr69:
Sorry but this causes the same issue as before.

pegr69:
Tested again, this time with the correct branch.. Works fine!

Issue #2296, opened 2020-05-11
Workflow Delegation step does not delegate to transient workflow


danxchapman:
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.5.4.0
* [x] ACS AEM Commons Version: 4.1.0
* [x] Reproducible on Latest? yes

### Expected Behavior

Workflow Delegation step delegates to transient workflow

### Actual Behavior

Workflow Delegation step does not delegate to transient workflow

### Steps to Reproduce

* Overlay OOTB Request for Activation workflow, adding Delegate Workflow step as the first step
* Add workflow delegation configuration for Request for Activation (delegate workflow step) to delegate to another workflow that is marked as transient
* Run Request for Activation for activation workflow (delegation step is ignored)

davidjgonzalez:
@danxchapman can you setup a DEBUG logger for `com.adobe.acs.commons.workflow.process.impl` and do this and post the logs?

Also, can you confirm what you're passing in as the PROCESS_ARGS?

```
* workflowModelProperty=<propName>
* defaultWorkflowModel=<pathToDefaultWorkflowModel>
* terminateWorkflowOnDelegation=true|false
``

Also, are you sure the workflowModelProperty is correctly set on your content tree? and/or are you using the defaultWorkflowModel -- if so, what is the path you are using?

WF's location moved around in 6.4, and i think you need to point to the `/var/workflow/models/...` workflow node for this to resolve correctly (since that's the runtime model).

**Comment by joerghoh** [comment]
I worked on this issue and came out with a way to cause an issue (not sure if this is the one @danxchapman was meaning).


* AEM 6.5.2, ACS AEM Commons Master branch
* Added a process step as first step to the "request for deletion" workflow
* configured that step with the "WorkflowDelegationStep" and provided these arguments
```
workflowModelProperty=foo
defaultWorkflowModel=/var/workflow/models/dam/update_asset
``` 
* invoked that on a random asset

(I know that this new workflow doesn't make any sense, but I just used it as vehicle to call a random transient workflow.)

Then I got these messages:
```
15.05.2020 19:54:57.057 *DEBUG* [JobHandler: /var/workflow/instances/server0/2020-05-15/request_for_deletion_8:/content/dam/global/products/phb7_4_heizung.pdf] com.adobe.acs.commons.workflow.process.impl.WorkflowDe
[...]

PR #2305, opened 2020-05-16
do not reuse the current wfData to start a new one

joerghoh:


badvision:
What is the overall effect of this change?

joerghoh:
it will fix #2296 :-)

badvision:
Works for me.  I'll merge if I get Travis-CI confirmation then.  Thanks @joerghoh!!

PR #2308, opened 2020-05-21
improve tests for DialogResourceProvider

joerghoh:
* enumerate the model classes for this test, so no changes are required when changes in other areas happen.
* cover the bind/unbind methods with a unittest.


joerghoh:
Ok, the reduction of test coverage is just a side-effect of limiting the test to the 3 slingmodels; now some code of the mentioned models is not executed anymore, which limits their test-coverage.

Shouldn't be a problem, instead we should improve the test coverage of the mentioned models.

badvision:
Thanks @joerghoh!

Issue #2313, opened 2020-05-24
DialogResourceProviderImpl throws StringIndexOutOfBoundsException


joerghoh:
## Required Information

* [X] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.5.2
* [X] ACS AEM Commons Version: latest
* [X] Reproducible on Latest? yes

I have ticked the "enabled" checkbox for the PID com.adobe.acs.commons.mcp.impl.DialogResourceProviderFactoryImpl and now I get periodically this exception in the logs:

```
24.05.2020 18:11:05.405 *DEBUG* [sling-default-4-health-com.adobe.granite.repository.hc.impl.content.sling.SlingContentHealthCheck] com.adobe.acs.commons.mcp.impl.DialogResourceProviderImpl Get resource at path: /apps/unknown/type/cq:dialog
24.05.2020 18:11:05.405 *ERROR* [sling-default-4-health-com.adobe.granite.repository.hc.impl.content.sling.SlingContentHealthCheck] com.adobe.granite.queries.impl.hc.QueryHealthCheckMetrics Error in refreshing the health check gauge
java.lang.StringIndexOutOfBoundsException: String index out of range: -2
        at java.lang.String.substring(String.java:1931)
        at com.adobe.acs.commons.mcp.impl.DialogResourceProviderImpl.getResource(DialogResourceProviderImpl.java:113) [com.adobe.acs.acs-aem-commons-bundle:4.7.1.SNAPSHOT]
        at org.apache.sling.resourceresolver.impl.providers.stateful.AuthenticatedResourceProvider.getResource(AuthenticatedResourceProvider.java:135) [org.apache.sling.resourceresolver:1.6.8]
        at org.apache.sling.resourceresolver.impl.helper.ResourceResolverControl.listChildren(ResourceResolverControl.java:298) [org.apache.sling.resourceresolver:1.6.8]
        at org.apache.sling.resourceresolver.impl.ResourceResolverImpl.listChildren(ResourceResolverImpl.java:518) [org.apache.sling.resourceresolver:1.6.8]
        at org.apache.sling.resourceresolver.impl.ResourceResolverImpl.hasChildren(ResourceResolverImpl.java:1005) [org.apache.sling.resourceresolver:1.6.8]
        at org.apache.sling.api.resource.AbstractResource.hasChildren(AbstractResource.java:111) [org.apache.sling.api:2.20.0]
[...]

PR #2317, opened 2020-05-27
Adding new annotation processor for generated dialogs

badvision:
This is a fix for #2313 as well as general deployment headaches caused by the factory service for the dialog generator.  Now an annotation processor generates small lightweight OSGi services for each dialog only as the models use @DialogProvider directly.  Therefore, dialog generation is purely opt-in and no feature flag is needed anymore either!

This is still undergoing testing but I've put it up here for comments in case anyone has any strong opinions either positive or negative.  Let me know what you folks think!


badvision:
@adamcin -- any ideas on why OakPal checks fail on the this?  I saw a message about OSGi annotations so I switched to SCR but still get the same error.  There is a new annotation processor but it generates code under impl directories to avoid exporting generated code -- otherwise I haven't a clue what the OakPal error is referring to.

badvision:
@adamcin the error reported is: `[INFO] --- oakpal-maven-plugin:2.0.0:verify (oakpal-verify) @ acs-aem-commons-ui.apps ---
[INFO] OakPAL Check Reports
[INFO]   com.adobe.acs.acs-aem-commons-oakpal-checks/acs-internal/base-version-compatibility
[ERROR]    +- <SEVERE> Package import org.osgi.service.component.annotations;version="[1.3.0,2.0.0)" cannot be satisified by AEM Version 6.3. Package is not exported. [acs-aem-commons-ui.apps-4.7.1-SNAPSHOT.zip]
[ERROR]    +- <SEVERE> Package import org.osgi.service.component.annotations;version="[1.3.0,2.0.0)" cannot be satisified by AEM Version 6.4. Package is not exported. [acs-aem-commons-ui.apps-4.7.1-SNAPSHOT.zip]
[INFO]   com.adobe.acs.acs-aem-commons-oakpal-checks/acs-internal/base-version-compatibility
[ERROR]    +- <SEVERE> Package import org.osgi.service.component.annotations;version="[1.3.0,2.0.0)" cannot be satisified by AEM Version 6.3. Package is not exported. [acs-aem-commons-ui.apps-4.7.1-SNAPSHOT.zip]
[ERROR]    +- <SEVERE> Package import org.osgi.service.component.annotations;version="[1.3.0,2.0.0)" cannot be satisified by AEM Version 6.4. Package is not exported. [acs-aem-commons-ui.apps-4.7.1-SNAPSHOT.zip]
[ERROR] ** Violations were reported at or above severity: MAJOR **`

badvision:
@adamcin never mind, I figured it out.  My annotation processor was importing the class and since my processor is an exported class it was creating that dependency.  I didn't actually need to import that class since I was just using the string representation (getClass().getCanonicalName()) so I removed that and used the raw string value instead.  OakPal is happy again.

badvision:
OK this is actually working so I think it's safe to leave review-only status.  I did a confirmation test with an additional project that was using the dialog generator and I didn't have to do any code changes but I did have to change my project's ACS Commons version to match and recompile.  I don't think that's a breaking change per se for folks that regularly use CI/CD.   Thoughts?

badvision:
LOL yeah that test coverage is NOT 100%.  Not sure why it decided to skip covering anything.  Might be a weird exclude pattern.  I'll try pulling that back and see if I can get something that makes sense again. :D  Good news is that the jacoco stuff executes now.

davidjgonzalez:
Coverage at 100%?! APPROVE AND MERGE! QUICK!!!

badvision:
> Coverage at 100%?! APPROVE AND MERGE! QUICK!!!

hahaha.. if only!!

adamcin:
> @adamcin never mind, I figured it out. My annotation processor was importing the class and since my processor is an exported class it was creating that dependency. I didn't actually need to import that class since I was just using the string representation (getClass().getCanonicalName()) so I removed that and used the raw string value instead. OakPal is happy again.

@badvision  Sorry I didn't see your comment earlier :) I'm glad you figured it out.

badvision:
Man what a slew of annoying build issues!  Glad it's all resolved and building cleanly now.  I also manually tested against existing implementations outside of ACS Commons that use the DialogProvider interface and they work with this new version with no changes needed outside of recompiling and redeploying so that the annotation processor can do its job.  This is ready to go.

badvision:
Not sure why the PR build shows pending, it reports successful completion if you click Details.

bundle/src/main/java/com/adobe/acs/commons/mcp/form/DialogProviderAnnotationProcessor.java
```diff
@@ -0,0 +1,136 @@
...
+import org.apache.sling.models.annotations.Model;
+
+/**
+ * Processes the DialogProvider annotation, producing a corresponding OSGi
+ * service to provide the generated sling dialog resources. This annotation
+ * processor will skip any classes which do not identify their corresponding
+ * sling model either as part of the model annotation or by a property or getter
+ * method.
+ */
+@Generated("Don't run code coverage, JaCoco!")
```
bstopp, line 46: Is _this_ class generated? Why not do code coverage?

bundle/src/main/java/com/adobe/acs/commons/mcp/form/DialogResourceProvider.java
```diff
@@ -0,0 +1,88 @@
...
+
+    default void doDeactivate() {
+        ServiceRegistration providerRegistration = registeredProviders.get(getTargetClass());
+        if (providerRegistration != null) {
+            providerRegistration.unregister();
+        }
+        registeredProviders.remove(getTargetClass());
+    }
+
+    public static String getServiceClassName(String modelClass) {
```
bstopp, line 68: This is probably why you're getting `squid:S1214` - can't just make this `default`?

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/DialogResourceProviderFactoryImpl.java
```diff
@@ -21,205 +21,46 @@
 
 import com.adobe.acs.commons.mcp.DialogResourceProviderConfiguration;
 import com.adobe.acs.commons.mcp.DialogResourceProviderFactory;
-import com.adobe.acs.commons.mcp.form.DialogProvider;
```
bstopp, line 24: While this is being deprecated - why are you deleting the implementation details now, rather than just waiting for it to be removed in 5.x?

bundle/pom.xml
```diff
@@ -1,785 +1,815 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd ">
-	<modelVersion>4.0.0</modelVersion>
```
bstopp, line 3: Really? A spacing format commit - with every line changed?

bundle/src/test/java/com/adobe/acs/commons/mcp/DialogResourceProviderFactoryTest.java
```diff
@@ -100,23 +103,63 @@ public void testResourceResolution() {
     }
 
     @Test
-    public void testDynamicRegistration() {
-        assertEquals("Should have 3 services registered", 3, dialogProvider.getActiveProviders().size());
+    public void testDeactivation() {
+        assertTrue("Should resolve model one dialog", resourceExists("/apps/test/model1/cq:dialog"));
+        providers.get("com.adobe.acs.commons.mcp.model.SimpleModelOne").doDeactivate();
+        assertFalse("Should resolve model one dialog", resourceExists("/apps/test/model1/cq:dialog"));
```
bstopp, line 109:
```suggestion
        assertFalse("Should not resolve model one dialog", resourceExists("/apps/test/model1/cq:dialog"));
```

bundle/src/test/java/com/adobe/acs/commons/mcp/DialogResourceProviderFactoryTest.java
```diff
@@ -100,23 +103,63 @@ public void testResourceResolution() {
...
+        factory.registerClass((String) null);
+        factory.unregisterClass((Class) null);
+        factory.unregisterClass((String) null);
+    }
 
-        dialogProvider.unbind(adapterFactory, properties);
-        assertEquals("Should have 3 services registered", 3, dialogProvider.getActiveProviders().size());
-        assertFalse("Should NOT resolve model four dialog", resourceExists("/apps/test/model4/cq:dialog"));
+    @Test
+    public void testAnnotationProvider() {
```
bstopp, line 127: Negative test cases? 
test cases when the annotation is added to an invalid class definition? Or one that has `getResourceType` and/or a field `resourceType` and/or the Resource adaptable annotation?

badvision: Negative discrimination, as in the annotation processor sees a class with the annotation but it lacks a resource type so it decides not to render an OSGi service.  This is because rendering an OSGi service with no resource type will result in an error later when that service tries to start and cannot figure out where in the JCR to latch itself.

badvision: So there's some positive and negative discrimination cases in the unit test to ensure that logic works -- in addition I also wrote something that calls the annotation processor directly so I could improve the code coverage.  Honestly I could just keep that google compiler test instead of the discrimination tests as they do the same thing, but the first set confirm that the annotation processor fired off in Maven prior to testing, the other runs the processor more directly within the test.

badvision: However the google compiler library only lets me check if a class was written. Funny enough it doesn't have a negative assert rule to assert that a class was NOT written.  Oh well.

badvision: There are cases where the annotation is used in a base class that is itself not a model and does not have a resource type.  Because the annotation supports inheritance I allow for it, such that the base class does not get a resource provider but its fully-implemented subclasses will.

badvision: This is also what allows all these code changes to be backwards compatible with no breaking changes from the previous code.  I would encourage people to use the annotation directly on a model class as it is clearer what it is doing, but realistically there's some flexibility if people want to assert that on a base class and keep their model classes super-lean.

bstopp: Ok, that helps. I was trying to figure out why there weren't negative validations, but this clarifies. you log and move on, and the onus is on the consumer to figure out why they aren't getting their services (e.g. they didn't meet the api spec.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/DialogResourceProviderFactoryImpl.java
badvision: yeah, the whole thing is emptied out and the unit tests even confirm that the methods do nothing.

bundle/pom.xml
badvision: That was my IDE -- sorry.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/DialogResourceProviderFactoryImpl.java
bstopp: Right, but _why_? Sorry i had typo in original question

bundle/src/main/java/com/adobe/acs/commons/mcp/form/DialogResourceProvider.java
badvision: I could just move it out of this class but I didn't want someone to mistake it as a new API class.  Feels like a lot of noise for moving a single-use utility function out; I have to use interface for OSGi reasons otherwise I'd prefer it to be an Abstract class.

bundle/src/main/java/com/adobe/acs/commons/mcp/form/DialogProviderAnnotationProcessor.java
badvision: Because it was bombing!   I eventually got Surefire to work after adding the Jacoco runtime dependency but I couldn't get Corbertura to work unless I told it to skip this class.  Funny enough even with this annotation the stupid thing was running code coverage anyway so I should just remove this vestige.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/DialogResourceProviderFactoryImpl.java
badvision: I took out the implementation because I don't want it firing off -- that OSGi service was causing a lot of churn when deploying other model classes (even when disabled!) so the whole point of this exercise was to put that to an end once and for good.  If anyone was calling that service directly (I don't know why they would) then at least it gives them a chance to realize it's going away and they shouldn't try to call it prior to its complete removal.

badvision: Another possibility (me overlooking the obvious, which happens more than I care to admit) is I could just kill the impl class entirely because it's not part of the public API surface and not a breaking change either.  Would that be more reasonable?

bstopp: Ah ok, no this is good. I was looking for clarity on the reasoning.

Issue #2327, opened 2020-06-06
Create a build profile which uses aem-sdk-api instead of the uber.jar


joerghoh:
To avoid problems like #2298 I would like to create create a maven profile which compiles acs-aem-commons against the latest aem-sdk-api for AEM as a CloudService instead of the uber.jar.

This should be implemented by a proper maven profile "cloudservice", which also requires Java 11 as a build platform. The build against the existing uber.jar will be moved into a default profile "classic".

PR #2328, opened 2020-06-06
#2327 build profile for AEM as a Cloud Service

joerghoh:
setting a dedicated build profile for cloudservice (including a dedicated travis job), fixing #2327

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: java
-dist: bionic
+dist: xenial
 jdk: 
 - oraclejdk8
 - oraclejdk11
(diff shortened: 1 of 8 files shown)


joerghoh:
These failing Travis builds are strange ... my local builds are all ok

joerghoh:
It seems to be a problem when running java8 only, can it reproduce locally.

joerghoh:
There are quite some changes which are required to make master compile against a recent aem-sdk-api version.
I will revert the changes to travis, so we can merge the build profile changes and fix the compilation issues afterwards.

Issue #2341, opened 2020-06-24
ACS Commons fails to deploy to AEM as a Cloud Service due to inclusion of /var nodes


davidjgonzalez:
## Required Information

**Originally reported with clear description of symptoms by @stefanseifert  at: https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2334**

* [X] AEM Version, including Service Packs, Cumulative Fix Packs, etc: AEM as a Cloud Service
* [X] ACS AEM Commons Version: 4.x.x
* [X] Reproducible on Latest? yes

ACS AEM Commons fails to deploy to AEM as a Cloud Service because AEM's replication service user does not support replication (due to permissions of the service user) or nodes under /var, which the ACS Commons `ui.content` project has.

![image](https://user-images.githubusercontent.com/1451868/85571161-eddb0d80-b601-11ea-8002-7b9b29e7c914.png)

These nodes should should be establishing using repoinit scripts (targetting the minimal runmodes) as needed. This looks trivial for all nodes EXCEPT `/var/acs-commons/reports` which contains cq:Page definitions (which are large, and complex in structure).

@klcodanr is there a reason these nodes are in `/var`? It actually seems strange these are in `/var` which historically has more transient, runtime content (these appear to be report definitions). Thoughts on if they could be easily moved outside of `/var`?

Also, can you help me confirm:

- [x] `/var/acs-commons/httpcache` -> Should exist on Author and Publish (@niekraaijmakers)
- [ ] `/var/acs-commons/mcp` -> Should exist only on Author (@badvision)
- [x] `/var/acs-commons/on-deploy-scripts` -> Should exist on Author and Publish (@HitmanInWis)
- [ ] `/var/acs-commons/reports` -> Should exist only on Author (@klcodanr)

The working PR to resolve this is at: https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/2342

klcodanr:
@davidjgonzalez 

We should be able to relocate. As a bridge, the listing component could be changed to look in two locations whereas new reports would be created in the new location and either add a feature or a note for AEM Cloud Service users telling them to relocate reports.  There are a couple options I could think of:

### 1. Put under /content

We could move to /content/acs-commons/reports or similar path.  There are a couple of downsides:

 - It makes me uncomfortable to put non-publishable content under the /content tree, especially since it's a reporting feature, which would seem _bad_ to accidentally publish
[...]

PR #2342, opened 2020-06-24
#2341 - Removed /var resources from ui.content package

davidjgonzalez:
Working PR to remove nodes from /var in ui.content (along w/ moving supporting service users to repo init)

Pending guidance from @klcodanr  on how to best move `/var/acs-commons/reports`

Moving `/var/acs-commons/reports` may take some effect. PRing as this at least fixes the `/var/workflow` issue (im unsure if `/var/acs-commons` will be problematic though; not sure what the AEM CS service users actual permissions are on AEM Publish service in the cloud.


badvision:
What is the minimum AEM version required that supports repoinit?

davidjgonzalez:
@badvision good call, repoinit was supported since AEM 6.4.0

PR #2345, opened 2020-06-25
Incorrect Injectors ordering with service.ranking #2344 - fix service…

### LINKED ISSUE #2344: Incorrect Injectors ordering with service.ranking
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.5.0
* [ ] ACS AEM Commons Version: 4.7.0

### Expected Behavior

Injectors must be executed in order defined in service.ranking osgi property

### Actual Behavior
Following Injectors does not execute in order defined in service.ranking property as they are using OSGI R6 annotations without explicit type specification. So service.ranking property treated as String but must be Integer and during service sorting their ranking treated as 0.

com.adobe.acs.commons.models.injectors.impl.ChildResourceFromRequestInjector
com.adobe.acs.commons.models.injectors.impl.HierarchicalPagePropertyInjector
com.adobe.acs.commons.models.injectors.impl.I18nInjector
com.adobe.acs.commons.models.injectors.impl.JsonValueMapValueInjector

### Steps to Reproduce
Open http://localhost:4502/system/console/status-slingmodels
and see that I18nInjector with ranking 5500 listed above OSGiServiceInjector with ranking 5000

bobi:
Fixes #2344, fix ranking according to OSGI specification

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/ChildResourceFromRequestInjector.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/ChildResourceFromRequestInjector.java
@@ -52,7 +52,7 @@ import java.util.function.Function;
 @Component(
         service = {Injector.class},
         property = {
-                Constants.SERVICE_RANKING + "=3000"
+                Constants.SERVICE_RANKING + ":Integer=3000"
         }
 )
 public class ChildResourceFromRequestInjector implements Injector {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 - #2311 - ResourceTypeHttpCacheConfigExtension does not work with multiple allowed paths
 - #2314 - Fixed java.lang.IllegalStateException: Not a JSON Object for CQIncludePropertyNamespaceServlet
 - #2330 - Deactivated VersionedClientlibsTransformerFactory.VersionableClientlibsTransformer for static page exports
-
+- #2344 - Fixed Injectors ordering according to service.ranking property
+ 
 ### Changed
 - #2324 - On-Deploy-Scripts are not supported on AEMaaCS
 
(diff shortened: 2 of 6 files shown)


bobi:
please include this fix in nearest future release

bobi:
ScrMetadataIT is failed because i have changed properties type, and this was expected.
How should i fix this?

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/ChildResourceFromRequestInjector.java
```diff
@@ -52,7 +52,7 @@
 @Component(
         service = {Injector.class},
         property = {
-                Constants.SERVICE_RANKING + "=3000"
+                Constants.SERVICE_RANKING + ":Integer=3000"
```
kwin, line 55: I would recommend switching to https://docs.osgi.org/javadoc/osgi.cmpn/7.0.0/org/osgi/service/component/propertytypes/ServiceRanking.html but that requires an update of the `maven-bundle-plugin` to a version >= 4.0

bobi: comments in pom.xml says that there some issues with 4.x.x version

davidjgonzalez:
@bobi You just need to add entries for each of those services (and the changed property) in the ScrMetadataIT.java file 

Similar to..

https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/2351/files#diff-5a3c21d38893ea910f9f7ab092cb17cdR120

bobi:
> @bobi You just need to add entries for each of those services (and the changed property) in the ScrMetadataIT.java file
> 
> Similar to..
> 
> https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/2351/files#diff-5a3c21d38893ea910f9f7ab092cb17cdR120

thank you, i have updated test

PR #2355, opened 2020-07-02
Feature/2354 - ACS AEM Commons web console

davidjgonzalez:
Per @royteeuwen simply brilliant (as well as brilliantly simple) suggestion of re-using the `/sites.html` tech to create an ACS AEM Commons Touch UI web console that replaces our dependency on Classic UI `/miscadmin`

This console is available at `/acs-commons.html` and direct linked via ACS Commons cards.

Note that the Generic Lists console is excluded from this, as they do not appear to be editable outside the new Generic List framework (i haven't looked into this too deeply though, @badvision might have ideas how to make this easily compatible)

Note that this keeps all the ACS Commons "content" under `/etc/acs-commons` (for now) but is easy enough to re-configure it to pull from another location in the future (ie. if we figure out a clean way to put it under `/content/acs-commons` and hide it from the OOTB Siteadmin; which i think will require some light resource decoration.)

Currently to access this console you either have to:

+ Enter `/acs-commons.html` into the address bar
+ Use one of the ACS Commons cards which have been updated to direct link into this area (ie. one of the features that maintains its content here)

@klcodanr assuming this is approved, we should be able to move Reports here.

image

image

image


klcodanr:
@royteeuwen @davidjgonzalez 

image

I'll work on getting the reports code updated to move over!

Issue #2354, opened 2020-07-02
TouchUI Console for ACS Commons (replace /miscadmin)


davidjgonzalez:
It would be nice to have a TouchUI console that has the same core features of the ClassicUI `/miscadmin` console.

* Create new page
* Copy
* Move 
* Delete
* Lock/Unlock
* Publish/Unpublish

![image](https://user-images.githubusercontent.com/1451868/86371401-d679d100-bc4e-11ea-94ee-78137e092608.png)

kwin:
Why not using Context-Aware Configurations with https://wcm.io/caconfig/editor/?
As "/etc" is going away anyways, these pages need to be moved to another location anyways...

davidjgonzalez:
@kwin most of the ACS here aren't configurations (rather content or utility pages)

Relying on wcm.io (which I very much wcm.io projects, don't get me wrong), violates the ACS Commons tenant of not requiring 3rd party dependencies in a big way; so their CAConfig editor wouldn't work.

Instead of moving to `/conf` (i have a fundamental irritation with how `/conf` is generally being (IMO) abused (even by the product) and turning into another `/etc`) -- I'd much rather see these moved to `/content/acs-commons` with a clean way to add `/content/acs-commons` to the OOTB Siteadmin exclude list (as noted in original issue). 

One nice thing about this approach is it is effectively, non-breaking parity w/ the current miscadmin-based ACS Commons functionality; so we aren't creating more work for ourselves in the future if/when we move away from `/etc/acs-commons` .. all this does is help us work in a fully TouchUI world (understanding we need to move the pages themselves off classic UI, but we have to do that regardless)

davidjgonzalez:
@justinedelson, do you have any insight you can share into the short/medium/long term realities of "/etc going away"? and/or what it's replacement will be (a mix of /conf and /content, or a new root, or a continuation of using /etc for specific use cases not satisfied by /conf or /content)

davidjgonzalez:
Though, this appears to be a problem w/ this impl:

```
[INFO] --- oakpal-maven-plugin:2.0.0:verify (oakpal-verify) @ acs-aem-commons-ui.apps ---
[INFO] OakPAL Check Reports
[INFO]   com.adobe.acs.acs-aem-commons-oakpal-checks/content-class-aem64/content-classifications
[...]

Issue #1909, opened 2019-05-24
Move /etc/acs-commons to /content/acs-commons


kwin:
### Expected Behavior

Since AEM 6.4 the migration of `/etc` to other areas of the repository has started. That means that in the end no code/content should end up below `/etc`. Further information in https://helpx.adobe.com/experience-manager/6-5/sites/deploying/using/repository-restructuring.html#

### Actual Behavior

ACS Commons still places content below `/etc/acs-commons`. Also clientlibraries are still placed below `/etc/clientlibs/acs-commons/`. 

In my opinion the support for Classic UI should be deprecated and the actual pages below `/etc/acs-commons` should be moved to `/content/acs-commons`. That would mean in the classic UI the pages are exposed below `/siteadmin` instead of `/miscadmin`. For the Touch UI nothing will change, as already now the navigation is contributed by https://github.com/Adobe-Consulting-Services/acs-aem-commons/tree/master/content/src/main/content/jcr_root/apps/cq/core/content/nav/tools/acs-commons (which hides the location of the actual underlying pages). At the same time the client libs should be moved from `/etc/clientlibs/acs-commons` to `/apps/acs-commons/clientlibs` (and made accessible via the proxy option).

justinedelson:
I think for the non-clientlib cases, /conf would make more sense.

This should really be decomposed into separate issues. #986 should cover the Generic Lists case, but there are obviously others which need to be dealt with (packagers, notifications, reports, etc.).

kwin:
`/conf` is IMHO not exposed at all in the Classic UI, so those pages would be invisible in Classic UI. If you don't want to place them below `/content` I would rather go for `/apps/acs-commons/content` because all those pages are in fact not editable (so they are comparable to e.g. /libs/granite/operations/content/systemoverview.html`). WDYT?

kwin:
I think as first step we could migrate all those pages which are marked in https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/content/src/main/content/META-INF/vault/filter.xml#L33 with an include pattern ending with `(.*)?` to `/apps/acs-commons/content`. For the other ones we need to think a bit more where to place them.

justinedelson:
Right, I was thinking of the authorable case that I mentioned. For standalone tools, /apps makes more sense. It'll even make sense for the admin tools which allow editing of the authorable things, but the things themselves belong under /conf

royteeuwen:
[...]

PR #2365, opened 2020-07-10
Moving reports from /var/acs-commons/reports to /conf/acs-commons-reports

### LINKED ISSUE #2341: ACS Commons fails to deploy to AEM as a Cloud Service due to inclusion of /var nodes

klcodanr:
To resolve #2341 

Old reports are still accessible and can simply be copied from /var to /conf


klcodanr:
@davidjgonzalez -- could you take a look at this? It's to resolve #2341

davidjgonzalez:
@klcodanr LGTM - just want to make sure `/conf` is the best place for this stuff... 

Relates to: https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/1909 and 
https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2354

See like we can:

+ Keep stuff in `/etc/acs-commons`
+ Move stuff to `/conf/acs-commons` (similar to generic lists v2), the oddity here is this stuff is generally not context aware configurations.
+ Move stuff to `/content/acs-commons` if we can figure out how to hide that from the normal Site admin (which im not sure how to do)

joerghoh:
I don't understand that report builder feature that good (haven't used it at all), but why can't we ship the the feature in /apps (the report page itself and also the default reports) and create custom reports in /conf?

davidjgonzalez:
@joerghoh - The Reports listing UI is already under /apps. Shipping the samples under /apps would make them read-only of course, so we 'd need to handle that, which is more code/change to this feature.
This Issue is to make minimal changes to this feature to make it AEM CS compatible (ie. removing the samples from /var).

My updates to this branch moved the content from /conf to /etc --  Simply because that's where everything else "like this" in ACS Commons writes to today (except Generic Lists v2), and we dont know yet where they will ultimately live (maybe /conf/acs-commons, maybe not). Atleast in /etc/acs-commons we should be able to perform a bulk "lift-and-shift" (w/ hopefully a simple migration mechanism) to whatever the new location is. If we fragment where this stuff lives, that "lift-and-shift" becomes harder (unless we've serendipitously picked the right place)

davidjgonzalez:
@klcodanr Any major concerns storing reports in `/etc/acs-commons` for now until we can figure out where all that stuff needs to ultimately get moved to?

Also, neither here nor there, but I removed reports from the ACS Commons TouchUI console since when I opened up a Report page for editing from it, I got an error saying AEM couldn't resolve `/bin/editor` .. and I didn't jump out to me where that was being defined (is that an details specific to Reports?) -- More curious that anything :)

klcodanr:
@davidjgonzalez I do :-( the big problem is that due to some hardcoding shenanigans in the WCMMode filter you cannot use touchui under /etc: https://blogs.perficient.com/2017/11/16/no-matter-what-a-anyone-tells-you-theres-no-touchui-under-etc/

davidjgonzalez:
Ah - ok. I'll revert!

davidjgonzalez:
@klcodanr reverted the change (left a fix in for the notification positioning)

klcodanr:
@davidjgonzalez @joerghoh -- where did we land on this? This one has been floating out here for awhile...

Issue #2366, opened 2020-07-10
CQIncludePropertyNamespaceServlet - UnsupportedOperationException


bobi:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.5
* [ ] ACS AEM Commons Version: 4.7.2
* [ ] Reproducible on Latest? yes

### Expected Behavior

Should work as described in documentation.

### Actual Behavior

i'm getting following exception for cq:Widget type with child elements, like select with options.

java.lang.UnsupportedOperationException: JsonObject
	at com.google.gson.JsonElement.getAsString(JsonElement.java:192)
	at com.adobe.acs.commons.wcm.impl.CQIncludePropertyNamespaceServlet$PropertyNamespaceUpdater.visit(CQIncludePropertyNamespaceServlet.java:289)

### Steps to Reproduce

exception occurs when you try to include widget that have child nodes, e.g. select with options

PR #2367, opened 2020-07-10
#2366 CQIncludePropertyNamespaceServlet - UnsupportedOperationException

bobi:
Fixing #2366


badvision:
I see that you've disambiguated primitive vs. element in your fix.  That makes a lot of sense.  Any chance of adding tests to assert both cases work?

bobi:
> I see that you've disambiguated primitive vs. element in your fix. That makes a lot of sense. Any chance of adding tests to assert both cases work?

i have added test to cover this case

badvision:
Thanks so much!   If all the checks pass then I'll merge this one in.  Will be included in the next release (most likely will be v4.8.0).

PR #2378, opened 2020-07-29
Sitemap Disable Vanity for URLs

### LINKED ISSUE #2377: Sitemap Disable Vanity for URLs
## Required Information

* [All] AEM Version, including Service Packs, Cumulative Fix Packs, etc: _______
* [All] ACS AEM Commons Version: _____
* [Y] Reproducible on Latest? yes/no

### Expected Behavior

When the SiteMap Generator creates a Sitemap, it creates a URL for the page based on the page's actual URL.

### Actual Behavior

When the SiteMap Generator creates a Sitemap, it creates a URL for the page based on the page's Vanity URL which is not desirable from an SEO perspective as this is usually redirected to the page's actual URL.

### Steps to Reproduce

Create a sitemap configuration that would reference a page with a vanity URL set.

klcodanr:
Resolves #2377


bundle/src/main/java/com/adobe/acs/commons/wcm/impl/SiteMapServlet.java
```diff
@@ -88,6 +87,8 @@
 
     private static final boolean DEFAULT_REMOVE_TRAILING_SLASH = false;
 
+    private static final boolean DEFAULT_DONT_USE_VANITY_URL = false;
```
joerghoh, line 90: double negation ... I would prefer a positive statement like
```
DEFAULT_DO_USE_VANITY_URL = true;
```
and use the same positive logic below as well. Much easier to understand.

joerghoh:
Hi @klcodanr

I took the liberty and inverted the logic at a few places, makes it easier to reason about it.

klcodanr:
Thanks @joerghoh  and @davidjgonzalez !

PR #2381, opened 2020-08-01
Fix compilation warnings

joerghoh:
* suppress deprecation warnings because we cannot fix them right now or with the currrent implementation
* use the static method valueOf instead of the constructors of the wrapped primitive types (Java9)
* replace type.newInstance() with the recommended version (Java9)


badvision:
@joerghoh The two code climate failures are not a result of a change in the code are they?

joerghoh:
Yeah, though about shorter signatures ... but if that is complaining, I will fix it.

davidjgonzalez:
@joerghoh What's the importance of the Java 9 call out? These obviously work in Java8 (travis passes). Did these only become recommendation in Java9?

joerghoh:
I mentioned "java 9" because these deprecations were made with that version, so you don't see it when you only build on Java 8.

Issue #2386, opened 2020-08-05
URL Asset Import Overrides Folder Titles


rbotha78:
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.5.5.0
* [x] ACS AEM Commons Version: 4.8.4
* [x] Reproducible on Latest? yes

### Expected Behavior

When a pre-created a folder structure exist, the asset importer should not replace the folder title with the node name.

### Actual Behavior

When a folder exists and assets will be place in it or its sub folders, the folder's title is replaced with the node name from the target url.

### Steps to Reproduce

Setup a clean AEM 6.5 SP5 instance.
Install latest version of ACS commons (4.8.4)
Create a folder hierarchy under Assets and give it user friendly titles.
Create a URL asset import file with target set to a location beneath the folder hierarchy.
Start MCP process with the spreadsheet.
Observe the folder hierarchy's titles all changed to node names after import completes.

### Links

Error seems to be in logic at https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/AssetIngestor.java#L389

PR #2387, opened 2020-08-05
#2386 - Bulk Asset Import Overwrites Existing Folder Titles

rbotha78:
When importing assets into an existing structure, folder titles should not be changed.


badvision:
Weird changelog.  😄 Can we make this behavior optional so that the user has the option of deciding if they want to update folder names?  I wrote it that way kind of on purpose but I can understand why you'd want it this way too.

rbotha78:
haha, was hoping to avoid that 😄
Updates pushed.

badvision:
Wow!  You're really on the ball today!  Thanks so much!

PR #2391, opened 2020-08-09
cloudservice buildprofile (preparation for 5.0 release)

joerghoh:
This PR adds support to build against

* the uber.jar of AEM 6.4 (profile "classic", default)
* the API of the AEM as a Cloud Service (profile "cloudservice")


joerghoh:
I extended the travis builds to include a run with the cloudservice profile. I had to disable the oakpal reporting actions for it, as it only makes sense on the classic profile, which needs to support both AEM 6.4 and AEM 6.5.

I don't like the way how I had to mess with the two profiles, but I haven't found any better way which worked for me. Especially you need to specify "-Dprofile=classic" if you want a "simple build" (activeByDefault isn't working because on Travis there is another jdk8 profile active). I appreciate any feedback to improve this and reduce the profile definitions.

@adamcin I found that I had to hardcode the buildinfo files with the package versions. Would love if the tests pick up all available files themselves.
Also for the 2 different profiles it would be good to specify the individual checks (AEM 6.x and/or CS) directly in the POM instead to hardcode them in Java and move the executions into the profiles.

Issue #2402, opened 2020-09-03
AemEnvironmentIndicatorFilter fails to trigger on the author environment, due to WCMMode being DISABLED


henrykuijpers:
## Bug Report

**Current Behavior**
When configuring the AemEnvironmentIndicatorFilter with the configuration: `css-color="#ff00ff"`, we do not see an environment indicator being shown in the OSGi console, in the AEM interface, etc. 

**Expected behavior/code**
An environment indicator being shown.

**Environment**
- AEM 6.4 SP8
- ACS AEM Commons 4.8.0
- JRE version 8 build 200+

**Possible Solution**
It looks like the WCMRequestFilter is being ran after the AemEnvironmentIndicatorFilter. Therefore the AemEnvironmentIndicatorFilter receives WCMMode.DISABLED as the current WCMMode, which is the default excluded WCMMode.

henrykuijpers:
It looks like the filter is being registered as a Jetty filter, while the WCMRequestFilter is a Sling Filter and therefore runs much later in the chain.

It also looks like a recent update changed the way of obtaining the WCMMode from request.getAttribute(...) to WCMMode.fromRequest(request).

https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/AemEnvironmentIndicatorFilter.java#L371

The problem here is that in the previous case, null was returned, but now WCMMode.DISABLED is returned (since WCMMode.fromRequest()) actually returns WCMMode.DISABLED when the attribute could not be found. Therefore, it never registers the filter.

WDYT @davidjgonzalez @justinedelson ?

rubnig:
If you change it to a Sling Filter instead of a Jetty Filter, you will also get the following problem: the Indicator won't show up in CRX/DE Publish (which it did before) because the WCMMode == WCMMode.DISABLED in that case too.

It looks like the Unit Test is also testing the class not as it should be, because it is using a spy: https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/test/java/com/adobe/acs/commons/wcm/impl/AemEnvironmentIndicatorFilterTest.java#L76

Maybe the test can also be updated with actual testing of the accepts method.

joerghoh:
Hi @rubnig, thanks for reporting. I will try to investigate on that.
[...]

PR #2407, opened 2020-09-12
#2402 fix wcmmode detection

joerghoh:
Do not display the environment indicator if a configured wcm mode is used for the request.
(This was never working and also not documented, but somehow there were configuration settings which made people think that it should work ...)


bundle/src/main/java/com/adobe/acs/commons/wcm/impl/AemEnvironmentIndicatorFilter.java
```diff
@@ -385,4 +425,43 @@ String getTitlePrefix() {
...
+		public void init(FilterConfig filterConfig) throws ServletException {
+			// ignore
+			
+		}
+
+		@Override
+		public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+				throws IOException, ServletException {
+
+		    SlingHttpServletRequest req = (SlingHttpServletRequest) request;
```
henrykuijpers, line 448: Unused variable can be removed

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/AemEnvironmentIndicatorFilter.java
```diff
@@ -385,4 +425,43 @@ String getTitlePrefix() {
...
+			
+			WCMMode mode = WCMMode.fromRequest(request);
+			if (isDisallowedWcmMode(mode, excludedWcmModes)) {
```
henrykuijpers, line 451: `isDisallowedWcmMode` could be moved into this inner filter class, but probably the unit tests cannot compile anymore then, since it's using that strange spy-construction.

henrykuijpers (changes requested):
Please at least  change service.ranking to an integer

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/AemEnvironmentIndicatorFilter.java
```diff
@@ -193,10 +219,9 @@ public final void doFilter(final ServletRequest servletRequest, final ServletRes
                             ? capturedResponse.getBufferedServletOutput().getBufferedString()
                             : null;
 
-            
-                            
             if (contents != null
-                    && StringUtils.contains(response.getContentType(), "html")) {
+                    && StringUtils.contains(response.getContentType(), "html")
+                    && innerFilterAcceptsInjection(request)) {
```
henrykuijpers, line 224: The `accepts`-method (which is checked before this try/finally block, for an early-exit) still obtains the `WCMMode`, which will never yield a value other than `DISABLED`. This was probably again masked by the unit tests due to the invalid use of spying/mocking internals of this class.

henrykuijpers (changes requested):

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/AemEnvironmentIndicatorFilter.java
henrykuijpers, line 224: An alternative approach could be to remove the whole inner-filter registration and change this `&& innerFilterAcceptsInjection(request)` to `&& !isDisallowedWcmMode(WCMMode.fromRequest(request), excludedWcmModes)`, since the request at this point actually will have the `WCMMode` set, since it has gone through the `WCMRequestFilter`.

The only case where things could go wrong that I could think of, would be (custom) code that is changing the `WCMMode` to `DISABLED` and fails to change it back to the original value. This kind of construction is often in place to render certain components with the `DISABLED`-mode.

joerghoh: I rewrote the unittests for the accept method.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/AemEnvironmentIndicatorFilter.java
joerghoh: It's still used within the next block to write log statements to the RequestProgressTracker

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/AemEnvironmentIndicatorFilter.java
```diff
@@ -328,10 +360,14 @@ protected final void activate(ComponentContext ctx) {
                     "(" + HttpWhiteboardConstants.HTTP_WHITEBOARD_CONTEXT_NAME + "=*)");
             filterRegistration = ctx.getBundleContext().registerService(Filter.class.getName(),
                     this, filterProps);
+            
+            // Register the innerFilter so it is invoked after the WcmRequestFilter (Ranking = 2000)
+            Dictionary<String, String> innerFilterProps = new Hashtable<String, String>();
+            innerFilterProps.put(EngineConstants.SLING_FILTER_SCOPE,EngineConstants.FILTER_SCOPE_REQUEST);
+            innerFilterProps.put(Constants.SERVICE_RANKING, "1000");
```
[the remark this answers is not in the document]
joerghoh: Updated

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/AemEnvironmentIndicatorFilter.java
```diff
@@ -328,10 +352,14 @@ protected final void activate(ComponentContext ctx) {
                     "(" + HttpWhiteboardConstants.HTTP_WHITEBOARD_CONTEXT_NAME + "=*)");
             filterRegistration = ctx.getBundleContext().registerService(Filter.class.getName(),
                     this, filterProps);
+            
+            // Register the innerFilter so it is invoked after the WcmRequestFilter (Ranking = 2000)
```
kwin, line 356: Can't we use declarative services for the inner filter as well and just depend on the outer one (by adding a reference to it)?

henrykuijpers: It's also possible to remove the whole inner filter and just move reading the WCMMode after `filterChain.doFilter()`

joerghoh: that sounds much better than the current approach, I'll give it a try.

joerghoh:
@henrykuijpers I am just trying the approach you suggested, but it doesn't work as expected. I think that the WcmMode set deeper further down the chain is not visible to this filter. I suspect that at some later point there is a wrapper, which doesn't pass this setting along. The approach with the inner filter is working fine.

henrykuijpers:
> @henrykuijpers I am just trying the approach you suggested, but it doesn't work as expected. I think that the WcmMode set deeper further down the chain is not visible to this filter. I suspect that at some later point there is a wrapper, which doesn't pass this setting along. The approach with the inner filter is working fine.

I wasn't expecting that. But indeed, then the inner filter (unfortunately(?)) needs to stay indeed. :)

badvision (approved):

Issue #2320, opened 2020-05-29
Use bnd-resolver-maven-plugin to check runtime compatibility with multiple AEM versions


kwin:
As there were recently some issues with compatibility in different versions of AEM (e.g. #2298) I suggest to use https://github.com/bndtools/bnd/tree/master/maven/bnd-resolver-maven-plugin to check the bundle constraints against multiple OSGi containers (lowest supported AEM version and newest supported AEM version).

A similar check has been implemented recently in https://github.com/apache/jackrabbit-filevault/commit/b8e952da80cf3a3cabd8fda6f89d0ed1b797cfb5.

That way the build will fail if the acs-aem-commons bundles cannot be resolved in one of those containers. It still needs to be checked if the uber-jars can be used for that check....

kwin:
At least the relevant `provide-capability` headers seem to be missing from https://repo1.maven.org/maven2/com/adobe/aem/aem-sdk-api/

joerghoh:
To address the same problem I would rather try to build against multiple versions of the uber.jar (6.4,6.5, AEMaaCS) with Travis.
My impression is that right now we cannot rely solely on OSGI metadata.

kwin:
We have to distinguish between compile time resolution failure and run time resolution failure. In some cases  although at compile time everything is correct a bundle still cannot be resolved in certain AEM versions at run time. Therefore I would really like to introduce this additional OSGi resolver check.

kwin:
I opened https://issues.apache.org/jira/browse/SLING-9883 to include the necessary manifest headers in future AEM API jars.

PR #2422, opened 2020-09-27
update to AEM 6.4 dependencies

kwin:


kwin:
@joerghoh and @davidjgonzalez Is there any way we can move forward here? This is quite fundamental for 5.0 IMHO and this merge blocks further improvements.

kwin:
@adamcin or @justinedelson 
Any idea how to easily update https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/oakpal-checks/src/main/resources/bundleinfo/6.4.json added in #1797. The information from the JSON is incorrect for AEM 6.4!

kwin:
Due to the lack of response I would propose to get rid of this specific OakPal check. Checks like these should be done on bundle level with the bnd-resolver-maven-plugin (https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2320) in the future (once Adobe starts providing full SDKs containing all bundle headers)

kwin:
This is now ready for merge. The build just fails because of #2472. (but it fails with the same error on the  target branch `feature/2288-prepare-release-500`)

Issue #1060, opened 2017-07-28
Tree activation utility


badvision:
As a regular user I would like to replicate a structure of folders (for DAM) or site/pages (for Sites) such that:
- Indicate the starting folder
- specify how many levels to go (aka 1 page, or all if desired)
- Indicate if this uses regular replication (all default publishers) or specify a list of publishers
- Pick either replication queue or self-managed queue for replication
- At the end, get a report of how many pages/folders/assets were replicated.

Also, this tool should be smart enough that once some entities are replicated, there is no need for subsequent replication of child nodes.  For example, replicating an asset is a terminal tree traversal operation.  That is, however, not the case for a page.

badvision:
@davidjgonzalez, @justinedelson any thoughts on activation of pages?  Does replicating a page automatically replicate any children/grandchildren etc as long as they themselves are not cq:page?  Are there any other gotchas?

davidjgonzalez:
Replication operates on the nt:hierarchy structure; so the cq:page and cq:page/jcr:content sub structure. 

main gotcha w page replication is replicating references which requires calls to reference provider and explicit replication of the referneced resources; with the advent of template editor based pages this web has expanded (templates, policies, etc).

I wonder if there is a tipping point where making a package ( ala SCD) of pages and supporting resources becomes better/worse than inidiviual repoication. (You'd want to do something like jcr package replication status updater to ensure packages resources get the right rep status)

davidjgonzalez:
Wdyt about making the selection criteria support a query as well? Maybe you could add the root plus level to the QueryHelperImpl and support all the resource collection options that supports?

andrewmkhoury:
@badvision and @davidjgonzalez I see a lot of AEM developers and admins asking for these abilities in an activation tool:
* Re-activate the last activated version of the content (this is used to build out new publish instances or fix content that failed to activate to one or more publish instances)
* Activate content based on a timeframe (e.g. date/time X to current date/time)
* Skip deactivated content
* Activate paths from query result

YegorKozlov:
my wish list:

- dry-run 
- replicate direct references. 
- optionally pass the activated resource through a user code. I had projects where all replications went through the Request For Activation workflow which included  workflow steps with business logic. It would be great to mimic it and execute user code before a resource is added to the replication queue. 
- throttle replication. The tool should be smart to  stop putting items in the replication queue if it is stuck or longer than N elements.

badvision:
[...]

PR #2424, opened 2020-09-29
Feature/tree activation

badvision:
This adds a tree activation (bulk publish) utility similar to the one discussed in #1060 

User can specify one or more replication agents using a comma-delimited list and/or regex patterns.

User can also specify if they want folders only or all items (useful for pushing out folder structures for assets).  Also the user can choose if they want MCP to manage the publishing which avoids the total mess of an over-stuffed publish queue.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/TreeReplication.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/TreeReplication.java
@@ -1,0 +1,316 @@
+/*
+ * #%L
+ * ACS AEM Commons Bundle
+ * %%
+ * Copyright (C) 2018 Adobe
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+package com.adobe.acs.commons.mcp.impl.processes;
+
+import com.adobe.acs.commons.fam.ActionManager;
+import com.adobe.acs.commons.fam.actions.ActionBatch;
+import com.adobe.acs.commons.functions.CheckedFunction;
+import com.adobe.acs.commons.mcp.ProcessDefinition;
+import com.adobe.acs.commons.mcp.ProcessInstance;
+import com.adobe.acs.commons.mcp.form.CheckboxComponent;
+import com.adobe.acs.commons.mcp.form.FormField;
+import com.adobe.acs.commons.mcp.form.PathfieldComponent;
+import com.adobe.acs.commons.mcp.form.RadioComponent;
+import com.adobe.acs.commons.mcp.model.GenericReport;
+import com.adobe.acs.commons.util.visitors.TreeFilteringResourceVisitor;
+import com.day.cq.commons.jcr.JcrConstants;
+import com.day.cq.contentsync.handler.util.RequestResponseFactory;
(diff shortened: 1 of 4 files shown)


bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/TreeReplication.java
```diff
@@ -0,0 +1,272 @@
...
+
+    public AtomicInteger replicationCount = new AtomicInteger();
+
+    private void performReplication(ActionManager t, String path) {
+        int counter = replicationCount.incrementAndGet();
+        if (queueMethod == QueueMethod.USE_MCP_QUEUE
+                || (queueMethod == QueueMethod.MCP_AFTER_10K && counter >= ASYNC_LIMIT)) {
+            performSynchronousReplication(t, path);
+        } else {
+            performAsynchronousReplication(t, path);
```
davidjgonzalez, line 228: Just to make sure I understand, if someone chooses NOT to use MCP they will use standard async replication which will toss the jobs into the sling job queue as fast this process can walk the tree right?

badvision: Correct.  And god help them if they do that with 20,000+ nodes.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/TreeReplication.java
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    public static int ASYNC_LIMIT = 10000;
+
+    Replicator replicatorService;
+
+    @FormField(name = "Starting Path",
+            component = PathfieldComponent.FolderSelectComponent.class,
+            description = "This item and its descendants will be published/unpublished")
+    private String startingPath = "/content/dam";
```
davidjgonzalez, line 92: This is an odd default startingPath for the default of FOLDERS_AND_PAGES_ONLY? ..

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/TreeReplication.java
badvision: The advantage of using MCP to manage the publish queue is twofold: 1) You can't kill AEM as easily even if you're publishing tens of thousands of things with it.  Standard product queue will seize up between 50k-100k items.  2) You get a progress bar that lets you know how much has been published and how much is left.  Oh, and that nice report at the end.

davidjgonzalez (approved):
LGTM - only thing is the default for path and filter seem strange since theyre at odds with one another.

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/TreeReplication.java
davidjgonzalez: yup - cool - just wanted to make sure I understood. Good default value :) 👍

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/TreeReplication.java
badvision: Oh that's true but the reason why its "Folders and Pages" is because pages have a duality of container + content -- there's really no way to say "Just these pages and not others" using the tree visitor -- it's node-type based.

I also didn't want it to default to /content because someone could wreak havoc not paying attention.  I could also make it blank by default if that makes more sense.  Actually, it might be blank in the UI because I didn't specify the default in the options of the FormField annotation.

davidjgonzalez: That's fair; I think ~everyone will be adjusting the path in some regard anyhow so not a big deal - just struck me as strange :)

badvision: 100% agreed.  Let's leave it blank so it's clear that the user has to specify it in the first place.

badvision:
For the record, I did NOT call your review stale.  That's Github trying to drive a wedge in our friendship.  Not cool, Github!

davidjgonzalez (approved):

PR #2435, opened 2020-10-14
Feature/update pom to install sling

royteeuwen:
Forgot to update one in the parent pom

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 
 
 ### Added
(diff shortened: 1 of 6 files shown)


badvision (commented):
Looks good to me.  Was the old plugin deprecated in favor of the new one?  Just want to make sure I'm understanding the change correctly.

royteeuwen:
@badvision there was a standardization at apache level to rename everything to {name}-maven-plugin. I think that officially all maven-{name}-plugins are the OOTB maven plugins if I remember correctly

badvision (approved):

Issue #2438, opened 2020-10-14
Enhancement :  Update username for Assets Moved with Renovator JCR_LAST_MODIFIED_BY


axp235:
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: AEM 6.5.2
* [x] ACS AEM Commons Version: 4.8.6
* [x] Reproducible on Latest? yes

### Expected Behavior
Update the JCR_LAST_MODIFIED_BY field of the assets moved by Renovator with username "Renovator".
Having a username updated for assets moved by "Renovator" helps organizations that have custom Asset workflows to exclude the assets moved from standard workflows (Activation Workflow) to control the publish of these moved assets to scene7 etc and also get a report of assets moved.

### Actual Behavior

Currently, the JCR_LAST_MODIFIED_BY is untouched.

### Steps to Reproduce

Upload a spreadsheet with a list of assets that need to be moved.

badvision:
It sounds like a compelling feature but we have to be careful not to impose additional metadata rules on the user.  For example, jcr last modified attributes are generally updated by the product when you save changes to a node and I'm not 100% sure we can change that to something other than the user name of the current session.  On the other hand, if we apply a new property called something slightly different we can store whatever we want and achieve the desired effect.   Unfortunately over time having two last modified by fields could get very confusing because we don't know which one is more correct by looking at it.

Instead, we might consider an alternative: An option exposed to the user that, when checked, will record the timestamp in a new property like "renovator_process_time" or something like that.   This tells us two useful things: 1) That renovator did something to this content and 2) When it happened.  So we could then use a date range to narrow down the time as well, making our query even more useful.

Thoughts?

badvision:
(it should be an optional feature that the user must opt-in, that much I'm pretty sticky about because some people might not want metadata updates of any kind because they could cause workflow triggers or other undesired side-effects)

axp235:
@badvision 
We drew inspiration for using the "LastModifiedBy" field after looking at some code in ACS Commons.
https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/4b3da3f5fdf0b59049b966bea9306d24c87d0293/bundle/src/main/java/com/adobe/acs/commons/dam/impl/ReviewTaskAssetMoverHandler.java#L145

We could go the route of adding a new property to have some constant value as we need that to exclude in workflow config. Date ranges may not be a great way to set up exclude for us. 

[...]

PR #2443, opened 2020-10-15
#2438 Adding last modified user name as Renovator.

mxp737:
#2438 Adding last modified user name as Renovator.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/renovator/MovingAsset.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/renovator/MovingAsset.java
@@ -19,10 +19,15 @@
  */
 package com.adobe.acs.commons.mcp.impl.processes.renovator;
 
+import com.day.cq.commons.jcr.JcrConstants;
+import com.day.cq.commons.jcr.JcrUtil;
 import com.day.cq.replication.ReplicationActionType;
 import com.day.cq.replication.ReplicationException;
 import com.day.cq.wcm.api.NameConstants;
+
+import java.util.Date;
 import java.util.concurrent.atomic.AtomicBoolean;
+import javax.jcr.Node;
 import javax.jcr.RepositoryException;
 import javax.jcr.Session;
 import org.apache.sling.api.resource.ModifiableValueMap;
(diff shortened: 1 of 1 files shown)


bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/renovator/MovingAsset.java
```diff
@@ -61,6 +67,14 @@ public void move(ReplicatorQueue replicatorQueue, ResourceResolver rr) throws Il
             session.getWorkspace().getObservationManager().setUserData("changedByWorkflowProcess");
             session.move(getSourcePath(), getDestinationPath());
             session.save();
+            if (session.nodeExists(getDestinationPath())) {
```
davidjgonzalez, line 70:
```suggestion
            if (session.nodeExists(getDestinationPath() + "/" + JcrConstants.JCR_CONTENT)) {
```

WDYT about that just to be ultra-safe? If were doing the check why not check exactly for the node were writing to?

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/renovator/MovingAsset.java
```diff
@@ -61,6 +67,14 @@ public void move(ReplicatorQueue replicatorQueue, ResourceResolver rr) throws Il
...
+                Node originalAssetJcrContentNode = session
+                        .getNode(getDestinationPath() + "/" + JcrConstants.JCR_CONTENT);
+
+                JcrUtil.setProperty(originalAssetJcrContentNode, JcrConstants.JCR_LASTMODIFIED, new Date());
+                JcrUtil.setProperty(originalAssetJcrContentNode, JcrConstants.JCR_LAST_MODIFIED_BY,
+                        DEFAULT_LAST_MODIFIED_BY);
```
davidjgonzalez, line 76: I assume there isnt a clean way to get the user who kicked off the Renovation?

badvision: That's the user that runs renovator so pull it from the resource resolver in hand.

axp235: @davidjgonzalez / @badvision : The driver behind adding "user name" is to have some constant value updated on an asset which acts as an indicator for our Asset Management workflow setup not to run workflows and push to scene7. This would be a similar concern for any organization who have their DAM integrated with Scene7 or any external asset storage system.

we have no concern retrieving username from resource resolver but the goal we are trying to achieve is put something on an asset which gives an indicator that it was touched by Renovator.

CC : @mxp737

davidjgonzalez (approved):

PR #2448, opened 2020-10-16
Fix #2441 - Memory Exhaustion with Large Report Download

### LINKED ISSUE #2441: Report download servlet can exhaust the heap if the report is too large
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.5.5
* [x] ACS AEM Commons Version: 4.4.0+

### Expected Behavior

Custom reports from ACS Commons should have realistic limits to prevent generating reports that are too large to be of any value to the user; unbounded queries can inadvertently return a very large report by accident.

### Actual Behavior

Report size is unbounded, begin/end conditions should be more enforced; also the report download servlet should stream data out of JCR and into CSV not hold anything in memory.  This behavior combined with an overly-broad query killed a production server with a 40gb heap!

image

### Steps to Reproduce

Find a DAM instance with millions of assets.  Create a report with an end date but no start date.  Then try to download the report.  Watch server melt to molten slag.

klcodanr:
Primary Changes:

 - uses stream instead of getting results as a list
 - fixes a problem where the alert wasn't appearing when the result writing failed (made it a big red box instead of trying to show the toast)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 <!-- Keep this up to date! After a release, change the tag name to the latest release -->
 [unreleased changes details]: https://github.com/Adobe-Consulting-Services/acs-aem-commons/compare/acs-aem-commons-4.7.2...HEAD
 
+### Fixed
+- #2441 - Memory Exhaustion with Large Report Download
+
 ## 4.8.6 - 2020-10-13
 
 ### Fixed
--- a/bundle/src/main/java/com/adobe/acs/commons/reports/api/ReportExecutor.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/reports/api/ReportExecutor.java
@@ -19,15 +19,14 @@
  */
 package com.adobe.acs.commons.reports.api;
 
-import org.apache.commons.lang.StringEscapeUtils;
-import org.apache.sling.api.SlingHttpServletRequest;
-import org.apache.sling.api.resource.Resource;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Map;
+
+import org.apache.commons.lang.StringEscapeUtils;
+import org.apache.sling.api.SlingHttpServletRequest;
+import org.apache.sling.api.resource.Resource;
+import org.slf4j.LoggerFactory;
 
 /**
  * Interface for report execution classes to implement. These should be Sling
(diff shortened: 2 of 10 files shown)


badvision:
I'm not too worried about 0.05% code coverage decrease, but any chance you can do a quick refactor to make code climate happy @klcodanr ?

badvision:
If you're swamped, let me know.  Looks like this is in a branch I can push changes to as well.

badvision (approved):
Looks good to me!  Great usage of Streams!

klcodanr:
Working on it now, I'll add a test for the getResultsList method and I _think_ that'll push the test coverage back up.

badvision:
Thanks @klcodanr for your continued commitment to quality.

badvision:
If you want to get the number up, add a test for ResultPage.equals to cover the various conditional cases.  I personally don't see how that's really all that valuable, but it would make the coverage go back up.  Thoughts, @klcodanr?

badvision:
OK that also makes sense. ;)  Curious to see the final automated test result but I anticipate we're good and can move forward with this one once everything clears.  Thanks again @klcodanr!

badvision (approved):

klcodanr:
Oops mean to leave this comment first. I looked into the codebase and found the QueryResults were not being checked for equality or put in a Map / Set so it seems safe to delete these methods.

badvision:
That's fine.  Looks like code coverage is back where it should be.  In the meantime, builds are passing but there was one transient failure: https://travis-ci.org/github/Adobe-Consulting-Services/acs-aem-commons/jobs/737126569
I'm not convinced it was related though.

klcodanr:
+0.01%

image

badvision:
Yeah, I can ignore that random failed test in this case, it was with cloud config tests unrelated to this code change.  
`[ERROR] valid(com.adobe.acs.commons.cloudconfig.impl.CloudConfigurationListImplTest)  Time elapsed: 0.042 s  <<< ERROR!
java.lang.RuntimeException: Unable to initialize JCR_MOCK resource resolver factory: Unable to inject mandatory reference 'serviceUserMapper' for class org.apache.sling.resourceresolver.impl.ResourceResolverFactoryActivator : no matching services were found.
Caused by: org.apache.sling.testing.mock.osgi.ReferenceViolationException: Unable to inject mandatory reference 'serviceUserMapper' for class org.apache.sling.resourceresolver.impl.ResourceResolverFactoryActivator : no matching services were found.`

badvision:
-0.03%  coverage?  I am confused.   Also ready to move on anyway.  Suggestions?

klcodanr:
Huh, weird. Looks like the merge caused a slight drop in code coverage:
https://coveralls.io/builds/34287731 v https://coveralls.io/builds/34287298

Either way, it's close to even.

badvision:
I think it's actually the test was returning a mock instead of calling the actual object method in this case.  I'll try this one-line change and see if it helps.

badvision (approved):

Issue #2451, opened 2020-10-20
Adding a new dispatcher cache control header filter based on the resource type of the page


mariahada:
I would like to propose a new feature to add a new dispatcher cache control header filter so that it is based on the resource type of a page combined with the filter pattern. It will only cover the max age header.

mariahada:
Does anyone know how this can be fixed? 

`Could not evaluate: org.apache.jackrabbit.vault.packaging.PackageException: javax.jcr.nodetype.ConstraintViolationException: OakConstraint0021: /home/users/system/acs-commons/resource-type-dispatcher-header-filter[[rep:SystemUser]]: Mandatory property rep:principalName not found in a new node`

PR #2452, opened 2020-10-20
#2451 - Adding a new dispatcher cache control header filter based on …

mariahada:
…the resource type of the page


badvision:
How is this filter configured?  Just trying to make sure I understand how it works.

royteeuwen:
> How is this filter configured? Just trying to make sure I understand how it works.

This is configured as an osgi config:
filter.pattern="/content/.*"
resource.types="my-app/components/page-resource-type"

This way you can override for example a default TTL pattern that you have on all content with specific ones for page types that should be invalidated faster (for example pages with dynamic lists on it)

bundle/src/main/java/com/adobe/acs/commons/http/headers/impl/ResourceTypeBasedDispatcherMaxAgeHeaderFilter.java
```diff
@@ -0,0 +1,124 @@
...
+import org.osgi.service.component.ComponentContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.servlet.http.HttpServletRequest;
+import java.util.Collections;
+import java.util.Dictionary;
+import java.util.Map;
+
+@Component(
```
davidjgonzalez, line 44: @mariahada can you update these to use the new OSGi annotations, rather than the old Felix annotations?

bundle/src/main/java/com/adobe/acs/commons/http/headers/impl/ResourceTypeBasedDispatcherMaxAgeHeaderFilter.java
```diff
@@ -0,0 +1,124 @@
...
+
+    @Override
+    @SuppressWarnings("unchecked")
+    protected boolean accepts(final HttpServletRequest request) {
+        if (!super.accepts(request)) {
+            log.debug("Not accepting request because it is not coming from the dispatcher.");
+            return false;
+        }
+
+        try (ResourceResolver resourceResolver = resourceResolverFactory.getServiceResourceResolver(AUTH_INFO)) {
```
davidjgonzalez, line 86: Is a service user required here? Doesn't the request require enough security context to be able to derive the Resource type anyhow?

Could this simply be:

```
return resourceTypes.stream().filter(rt -> { 
    if(slingRequest.getResource().isResourceType(rt)) { 
       return true; 
   } else if (slingRequest.getResource().getChild("jcr:content") != null) {
       return slingRequest.getResource().getChild("jcr:content").isResourceType(rt);
   } ).anyMatch(); 
``

File: bundle/src/main/java/com/adobe/acs/commons/users/impl/ProtectedAuthorizables.java [hunk 3]
```diff
@@ -159,6 +159,7 @@ private ProtectedAuthorizables() {}
                 "acs-commons-review-task-asset-mover-service",
                 "acs-commons-httpcache-jcr-storage-service",
                 "acs-commons-manage-controlled-processes-service",
+                "acs-commons-resource-type-dispatcher-header-filter",
```
davidjgonzalez, line 162: Same; why do we need as service user? if the request cant access the resource to read the resource type, then this will never get hit?

davidjgonzalez (commented):
Can you check the comments out? 

Any chance you could drop a unit test or 2 in?

bundle/src/main/java/com/adobe/acs/commons/http/headers/impl/ResourceTypeBasedDispatcherMaxAgeHeaderFilter.java
royteeuwen: It would have been that simple, if the dispatcher filters were sling filters, but these are registered through the http whiteboard so they don't have a SlingHttpServletRequest but a normal HttpServletRequest

davidjgonzalez: oh they are? the `@Property(name = "sling.filter.scope", value = {"REQUEST", "FORWARD"})` threw me ... i assumed they were Sling Filters .. but it does appear theyre registered with the OSGi HTTP Whiteboard...  

@bstopp  you remember why these aren't Sling? is it worth/could we safely switch these over to register as Sling Filters as it removes the need for a service user ..

royteeuwen:
any update on this @davidjgonzalez  ? Do you need anything from me? If you can already give me more input on what is necessary to get it merged that would be great

davidjgonzalez:
@royteeuwen i asked @bstopp  and he doesnt remember why he use the OSGi filter vs. Sling filter... agh!

It would be nice to move these to Sling if theyll work, i really dont like service users if they can be avoided.

davidjgonzalez:
Once #2465 passed travis, and we merge - we should be able to adapt this code to use the sling filter approach.

royteeuwen:
Super @davidjgonzalez . Will make the adjustments at that moment. Any idea yet on when release 5.0.0 is planned, so that I can convert them to OSGi DS?

bundle/src/main/java/com/adobe/acs/commons/http/headers/impl/AbstractDispatcherCacheHeaderFilter.java
```diff
@@ -64,14 +65,14 @@
 
     /**
      * Get the value to place in the Cache-Control header.
-     * 
+     *
```
joerghoh, line 68: Can you please cleanup the diff a bit? It's not only about the diff here in github, but also the standard git command line client does not have the "ignore whitespaces" setting as default.

bundle/src/main/java/com/adobe/acs/commons/http/headers/impl/ResourceTypeBasedDispatcherMaxAgeHeaderFilter.java
joerghoh: Converting to Sling Filters would make sense here. The notion of resource types only exists in Sling, and not on fetures outside of Sling (CRX/DE, Package Manager, OSGI Webconsole).

royteeuwen:
any comment on this one? What should still happen to get this one merged?

royteeuwen:
@davidjgonzalez @joerghoh  Sorry to ping again, but is there anything you would like to see changed to get this PR approved / merged? We don't mind doing the work!

kwin:
@royteeuwen Is this now obsolete with #2511 merged? Can you close this PR then?

PR #2453, opened 2020-10-20
Feature/parent resource valuemap injector

nithindasyam:


HitmanInWis:
I'll take on this review.

badvision:
Transient build failure isn't because of this PR, not sure why we're getting those unless it's a race condition @HitmanInWis

HitmanInWis:
Not the first time I've seen a completely random test failure in the test suite.  We'll see if a future commit cleans it up automagically :)

bundle/src/main/java/com/adobe/acs/commons/models/injectors/annotation/ParentResourceValueMapValue.java
```diff
@@ -0,0 +1,49 @@
+/*
```
HitmanInWis, line 1: Will need an update to CHANGELOG.md for this new feature.

bundle/src/main/java/com/adobe/acs/commons/models/injectors/annotation/ParentResourceValueMapValue.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Annotation to be used on either methods, fields to let Sling Models inject an inherited property from the
+ * parent resource
+ */
+@Target({METHOD, FIELD, PARAMETER})
+@Retention(RUNTIME)
+@InjectAnnotation
+@Source(ParentResourceValueMapValue.SOURCE)
+public @interface ParentResourceValueMapValue {
```
HitmanInWis, line 42: Will eventually want a documentation page for this new injector under https://adobe-consulting-services.github.io/acs-aem-commons/features/sling-model-injectors/index.html.  We dont have to hold up this PR for that documentation, but also dont want to forget it.

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/ParentResourceValueMapValueInjector.java
```diff
@@ -0,0 +1,85 @@
...
+import java.lang.reflect.AnnotatedElement;
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Injects the parent component property
+ */
+@Component(property = Constants.SERVICE_RANKING + ":Integer=" + Integer.MAX_VALUE, service = Injector.class)
+public class ParentResourceValueMapValueInjector implements Injector {
```
HitmanInWis, line 41: Please add unit tests for the new injector code.  Similar tests can be found for com.adobe.acs.commons.models.injectors.impl.ChildResourceFromRequestInjector

bundle/src/main/java/com/adobe/acs/commons/models/injectors/annotation/ParentResourceValueMapValue.java
```diff
@@ -0,0 +1,49 @@
...
+
+    String SOURCE = "inherited-property";
+
+    InjectionStrategy injectionStrategy();
+
+    int maxLevel() default -1;
```
HitmanInWis, line 48: do you want "any ancestor" as the default, or do you want the default functionality to only check the parent (in which case the default should be 1)?

bundle/src/main/java/com/adobe/acs/commons/models/injectors/annotation/ParentResourceValueMapValue.java
```diff
@@ -0,0 +1,49 @@
...
+ * Annotation to be used on either methods, fields to let Sling Models inject an inherited property from the
+ * parent resource
+ */
+@Target({METHOD, FIELD, PARAMETER})
+@Retention(RUNTIME)
+@InjectAnnotation
+@Source(ParentResourceValueMapValue.SOURCE)
+public @interface ParentResourceValueMapValue {
+
+    String SOURCE = "inherited-property";
```
HitmanInWis, line 44: Might make sense to name this more close to the name of the class injector to prevent future conflicts - perhaps "parent-valuemap-value"

bundle/src/main/java/com/adobe/acs/commons/models/injectors/annotation/impl/ParentResourceValueMapValueAnnotationProcessorFactory.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * #%L
+ * ACS AEM Commons Bundle
+ * %%
+ * Copyright (C) 2018 Adobe
```
HitmanInWis, line 5: Nitpick, but probably put 2020 on new files.

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/ParentResourceValueMapValueInjector.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @Override
+    public String getName() {
+        return ParentResourceValueMapValue.SOURCE;
+    }
+
+    @Override
+    public Object getValue(Object adaptable, String name, Type type, AnnotatedElement annotatedElement, DisposalCallbackRegistry disposalCallbackRegistry) {
+        Integer maxLevel = annotatedElement.getAnnotation(ParentResourceValueMapValue.class).maxLevel();
```
HitmanInWis, line 50: Should probably check the annotation, similar to what's being done in com.adobe.acs.commons.models.injectors.impl.HierarchicalPagePropertyInjector#getValue

``` 
       if (!element.isAnnotationPresent(HierarchicalPageProperty.class)) {
            //skipping javax.Inject for performance reasons. Only supports direct injection.
            return null;
        }
```

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/ParentResourceValueMapValueInjector.java
```diff
@@ -0,0 +1,85 @@
...
+            if (parentResource != null && !parentResource.getResourceType().equals(NameConstants.NT_PAGE)) {
+                parentResources.add(resource.getParent());
+                count++;
+            } else {
+                break;
+            }
+            parentResource = parentResource.getParent();
+        }
+        while (count < maxLevel || maxLevel <= 0);
+        return new MergedValueMap(parentResources).get(propertyName, (Class<?>) declaredType);
```
HitmanInWis, line 83: Getting all of the parent valuemaps and merging them all just to get a single property feels a bit heavy.  Within the loop couldn't we simply call `resource.getValueMap().get(propertyName, declaredType)` and then `break;` if a non-null value is found?  That way we dont even look at higher resources in the tree than necessary when the property is at the lower levels, and we dont have to bother merging a bunch of maps.

Thoughts?

HitmanInWis (changes requested):

bundle/src/main/java/com/adobe/acs/commons/models/injectors/annotation/ParentResourceValueMapValue.java
nithindasyam: I wanted the default as any parent resource unless specified.

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/ParentResourceValueMapValueInjector.java
nithindasyam: Yeah. Working on it.

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/ParentResourceValueMapValueInjector.java
nithindasyam: Sure. Will add the check.

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/ParentResourceValueMapValueInjector.java
nithindasyam: Yeah, this makes sense if the max-level value is high and has to iterate through a bunch of nodes or if each node has many properties.

bundle/src/test/java/com/adobe/acs/commons/models/injectors/impl/model/impl/TestParentResourceValueMapValueModelImpl.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.sling.models.annotations.DefaultInjectionStrategy;
+import org.apache.sling.models.annotations.Model;
+import org.apache.sling.models.annotations.injectorspecific.InjectionStrategy;
+
+@Model(adapters = TestParentResourceValueMapValueModel.class,
+        adaptables = {Resource.class, SlingHttpServletRequest.class},
+        defaultInjectionStrategy = DefaultInjectionStrategy.OPTIONAL)
+public class TestParentResourceValueMapValueModelImpl implements TestParentResourceValueMapValueModel {
+
+    @ParentResourceValueMapValue(injectionStrategy = InjectionStrategy.OPTIONAL, maxLevel = 1)
```
HitmanInWis, line 35: We should add a test case that ensures that a property at level 2 is not picked up when maxlevel is 1.

nithindasyam: Updated the unit test with this scenario.

bundle/src/main/java/com/adobe/acs/commons/models/injectors/annotation/ParentResourceValueMapValue.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+@Target({METHOD, FIELD, PARAMETER})
+@Retention(RUNTIME)
+@InjectAnnotation
+@Source(ParentResourceValueMapValue.SOURCE)
+public @interface ParentResourceValueMapValue {
+
+    String SOURCE = "parent-valuemap-value";
+
+    InjectionStrategy injectionStrategy();
```
HitmanInWis, line 46: This property should have a default value like the other injectors do, probably ` default InjectionStrategy.DEFAULT`

bundle/src/main/java/com/adobe/acs/commons/models/injectors/annotation/ParentResourceValueMapValue.java
```diff
@@ -0,0 +1,49 @@
...
+ * parent resource
+ */
+@Target({METHOD, FIELD, PARAMETER})
+@Retention(RUNTIME)
+@InjectAnnotation
+@Source(ParentResourceValueMapValue.SOURCE)
+public @interface ParentResourceValueMapValue {
+
+    String SOURCE = "parent-valuemap-value";
+
```
HitmanInWis, line 45: We should add the ability to name the field, in case the variable name doesnt match the property name.  This is standard on other similar annotations.

`
    String name() default StringUtils.EMPTY;
`

HitmanInWis (approved):
Everything looks good and tests out as expected.

PR #2454, opened 2020-10-20
Fixed #2450 - Write BOM to the beginning of the CSV

### LINKED ISSUE #2450: Non-Latin letters shown as "?" in the downloaded report
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.5.6
* [ ] ACS AEM Commons Version: 4.5.0

### Expected Behavior
Non-Latin letters are shown properly in the CSV report file

### Actual Behavior

Non-Latin letters are shown as "?" 

### Steps to Reproduce

Configure a report  (Tools -> Acs Aem Commons -> Reports) as described below:

1. Setup a JCR query report configuration.
 - Query language: JCR-SQL2
 - Query: SELECT * FROM [cq:Page] AS p WHERE ISDESCENDANTNODE('{{path}}')
2. Setup ACS Commons Report Builder Basic Parameter
 - Label: Path
 - Name: path
 - Field type: Text Field
3. Setup ACS Commons Report Builder Text Column
 - Heading: Title
 - Property: jcr:content/jcr:title
4. Setup ACS Commons Report Builder Text Column
 - Heading: Description
 - Property: jcr:content/jcr:description
5. Add a non-Latin letters in the description/title property on some pages
6. Open the configured report
7. Add a path that contains the pages with the edited description/title
8. Execute and download the report
=> The non-Latin letters are shown as "?" in the downloaded CSV report file
Note: The same letters are shown properly on the page when the report is executed.

badvision:
@joerghoh, @shsteimer -- do you think we should assume UTF-8 compatibility in external tools?  If so this patch is gold.  Otherwise, we might consider changing output encoding for CSV.  In honesty, there's nothing in the CSV format that notes which encoding format is mandated; and there's even ways to force Excel to assume UTF-8 without the BOM.  But this makes life easier for folks using standard tools, so I think it's worth considering this patch as-is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 
 ### Fixed
 - #2441 - Memory Exhaustion with Large Report Download
+- #2450 - Non-Latin letters shown as "?" in the downloaded report
 
 ## 4.8.6 - 2020-10-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 ### Fixed
 - #2441 - Memory Exhaustion with Large Report Download
 - #2450 - Non-Latin letters shown as "?" in the downloaded report
+- #2446 - One page is displayed in multiple lines in Report Builder export CSV file 
 
 ## 4.8.6 - 2020-10-13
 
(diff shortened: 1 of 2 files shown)


shsteimer:
@badvision I think assuming tools reading the CSV are UTF-8 compatible is a reasonable to make. As far as I know, excel, numbers, etc. should all work with that OK. We should also document that assumption on https://adobe-consulting-services.github.io/acs-aem-commons/features/report-builder/index.html just to make it clear to users, in case they are using other tools.

PR #2460, opened 2020-10-31
Don't defer flushing for unbuffered output

### LINKED ISSUE #2459: BufferedServletResponse should only defer flushing if the output is really buffered
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.5.6

### Expected Behavior

Even in case some filter leveraging the BufferedServletResponse (e.g. the AEMEnvironmentIndicatorFilter) is used, it should not defer flushing unbuffered responses. In most of the cases only the writer is buffered (i.e. text output) while the outputstream (binary output) is not buffered.

### Actual Behavior

The following warn can be observed in the log
```
30.10.2020 15:13:50.982 *WARN* [0:0:0:0:0:0:0:1 [1604067230756] GET /content/...../_jcr_content/image.img.840.560.70.20201018174339.jpg HTTP/1.1] com.adobe.acs.commons.util.BufferedServletResponse Prevent committing the response, it will be committed deferred, i.e. once this buffered response is closed
```

This is due to the following lines

https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/d4640dd5ce47233a68c4d81ee093b90c76f68f83/bundle/src/main/java/com/adobe/acs/commons/util/BufferedServletOutput.java#L195

This does not only pollute the log unnecessarily, it also leads to latency degradations and increases memory consumption.

### Steps to Reproduce

Just enable the AEM Environment filter and request some binaries (i.e. images)

kwin:
This closes #2459

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 - #2441 - Memory Exhaustion with Large Report Download
 - #2450 - Non-Latin letters shown as "?" in the downloaded report
 - #2446 - One page is displayed in multiple lines in Report Builder export CSV file 
+- #2459 - BufferedServletResponse should only defer flushing if the output is really buffered
 
 ## 4.8.6 - 2020-10-13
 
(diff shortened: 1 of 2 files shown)


davidjgonzalez (approved):

kwin:
@davidjgonzalez Resolving the conflict lead to losing your approval. Can you approve again?

davidjgonzalez (approved):

PR #2481, opened 2020-11-21
Prevent emitting buffered response twice in case AEMEnvironmentFilter is

### LINKED ISSUE #2475: Content rendered twice on publisher when environment indicator is enabled
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.5.6.0
* [ ] ACS AEM Commons Version: 4.9.2

### Expected Behavior

Content should not be rendered twice on a publisher when the AemEnvironmentIndicatorFilter is enabled.

### Actual Behavior

Pages are rendered twice on publish instance. On author it's working fine, also with wcmmode=disabled.

### Steps to Reproduce

Enable the EnvironmentIndicatorFilter on a publish instance

```
<jcr:root xmlns:sling="http://sling.apache.org/jcr/sling/1.0" xmlns:jcr="http://www.jcp.org/jcr/1.0"
          browser-title-prefix="Local"/>
```

kwin:
enabled and output is of type writer but no html

This closes #2475

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 <!-- Keep this up to date! After a release, change the tag name to the latest release -->
 [unreleased changes details]: https://github.com/Adobe-Consulting-Services/acs-aem-commons/compare/acs-aem-commons-4.7.2...HEAD
 
+### Fixed
+- #2475 - Content rendered twice on publisher when environment indicator is enabled
 
 ## 4.10.0 - 2020-11-19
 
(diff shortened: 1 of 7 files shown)


kwin:
@joerghoh Please have a look as well. I tested locally against the use case described in https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2271, i.e. requested `http://localhost:4502/etc/cloudsettings.kernel.js/libs/settings/cloudsettings/legacy/contexthub` multiple times and I haven't seen any 500 error.

joerghoh:
Is it possible to add a new unit test for this? We had quite some issues in the last releases with this, so additional tests to validate this behavior would be great.

kwin:
bq. Is it possible to add a new unit test for this? 

This would require a real integration test. Most of the problem only appear under certain circumstances (e.g. flush called in specific conditions or filter called at a specific location in the pipeline so this is hard to cover with just a unit test).

kwin:
I added https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2482 to track adding ITs. Adobe would need to come up with instances though, I don't know how feasible and realistic that is from a licensing perspective.

davidjgonzalez (approved):

PR #2485, opened 2020-11-25
Feature/generic lists asset metadata json

### LINKED ISSUE #2479: [Generic Lists][Touch UI Asset Metadata Editor]- Prefixing /mnt/acs-commons/lists to the name of a generic list in the JSON path of dropdown fields doesn't work
* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: AEMaaCS, 6.5, 6.4 (Not really dependent on AEM version I believe)
* [ ] ACS AEM Commons Version: 4.8.0
* [ ] Reproducible on Latest? yes/no : Yes

### Steps to reproduce

1. Navigate to Tools -> Assets->Metadata Schemas
2. Configure a dropdown field and add the options by configuring a JSON path - /mnt/acs-commons/lists/{name-of-the-generic-list}.json
3. Go to asset's properties, the dropdown field should list the options configured in the generic list

### Actual Behavior
The assets' properties ui is broken because it expects "text" in /mnt/acs-commons/lists/{name-of-the-generic-list}.json instead of "title"

Expected JSON

Screenshot 2020-11-18 at 4 56 58 PM

Actual JSON

image

shsteimer:
Fixes #2479

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/genericlists/impl/GenericListImpl.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/genericlists/impl/GenericListImpl.java
@@ -29,7 +29,6 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 
-import com.google.gson.annotations.SerializedName;
 import org.apache.sling.api.resource.Resource;
 import org.apache.sling.api.resource.ValueMap;
 
--- a/bundle/src/main/java/com/adobe/acs/commons/genericlists/impl/GenericListImpl.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/genericlists/impl/GenericListImpl.java
@@ -28,6 +28,8 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+
+import com.google.gson.annotations.SerializedName;
 import org.apache.sling.api.resource.Resource;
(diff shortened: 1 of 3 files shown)


bundle/src/main/java/com/adobe/acs/commons/genericlists/impl/GenericListImpl.java
```diff
@@ -39,6 +41,7 @@
 
     public static final class ItemImpl implements Item {
 
+        @SerializedName("text")
```
davidjgonzalez, line 44: Is this a breaking change? Would it be reasonable to assume someone might be consuming this JSON looking for the title property?
Perhaps we should emit 3 properties in the JSON? `title`, `text`, `value` ... where `title` is around for BC?

shsteimer: yea, that is likely safer.  

If you trace the history, the sole purpose of this JSON, as best I can tell, is to support asset metadata schema, which uses text/value. At some point, it got changed to title/value/props, which won't work for asset metadata schema. But if someone is using this json for some other purpose or maybe has customized asset metadata schema to use it as is), then removing the title property would break that.

Anyway, I'll fix it so it's BC, but just giving the context for why I did it the way I did.

davidjgonzalez (approved):

Issue #2478, opened 2020-11-18
 Dispatcher Flush Rules - Not compatible with Dispatcher Flush Refetch configuration


shashimuluguaem:
## Required Information

* [ ] AEM 6.4
* [ ] 3.19.0
* [ ] yes

### Expected Behavior

I have configured ACS commons Dispatcher Flush Rules on AEM 6.4 instance and everything is working fine till i changed my Flush agent to Re-fetch Flush Agent.

### Actual Behavior

Re-fetch Flush Agent is not being invoked for the rules configured in ACS Commons Flush Rules

shashimuluguaem:
@davidjgonzalez @justinedelson Any Help here?

davidjgonzalez:
@shashimuluguaem  What's the Serialization Type for refetch? Is it "flush"? If not, that won't work without some code adjustments. Do you happen to know what the value of the Refetch serialization type is?

joseqwd:
Hi @davidjgonzalez 

To respond to your question, on our ContentBuilder implementation for re-fetch we are using a different name instead of `flush`. 
When we tried with `flush` it doesn't get registered and is not available as a serialization type option in the Agent configuration.
Any suggestions?

atopper:
I have just been working on this topic for a certain customer (activating a lot of DAM Assets, using Rules to invalidate (evict) dispatcher cache pages during a high traffic times which swamps the Publisher).
At this stage I'd like to eventually propose a PR to this repo.  Is there anyone who I should be communicating with for this?

Looks like this was considered before: https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/143

PR #2487, opened 2020-11-26
Dispatcher Flush Re-Fetch: issues/2478

atopper:
Adding a synchronization type to allow re-fetches on a flush.
- [X] Tested on 6.5.
- [X] Verified installation and smoke tested on 6.3 and 6.4
- [X] Tests added with 91% code coverage
- [X] Updated CHANGELOG

image
image

cc @davidjgonzalez

Issue: https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2478


bundle/src/main/java/com/adobe/acs/commons/replication/dispatcher/DispatcherFlushFilter.java
```diff
@@ -130,13 +130,13 @@ private boolean isEnabled(final Agent agent) {
...
-     * Checks if the agent has a "flush" serialization type.
+     * Checks if the agent's serialization type starts with "flush" .
      *
      * @param agent Agent to check
-     * @return true if the Agent's serialization type is "flush"
+     * @return true if the Agent's serialization type starts with "flush"
      */
     private boolean isFlushingAgent(final Agent agent) {
-        return StringUtils.equals(SERIALIZATION_TYPE, agent.getConfiguration().getSerializationType());
+        return StringUtils.startsWith(agent.getConfiguration().getSerializationType(), SERIALIZATION_TYPE);
```
atopper, line 139: @davidjgonzalez 
I'd like your opinion on this change to existing code.  Instead of having the agent be exactly "flush", the check changed to be starting with "flush".  That'll be backwards compatible.  Will there be collisions?  More than 1 agent trying to do the activation if configured incorrectly?
Thoughts?

bundle/src/main/java/com/adobe/acs/commons/replication/dispatcher/refetchflush/impl/RefetchFlushContentBuilderImpl.java
```diff
@@ -0,0 +1,421 @@
...
+        if (shouldIgnore(action, path)) {
+            return ReplicationContent.VOID;
+        }
+
+        logInfoMessage("Content builder invoked for path " + path + ", with replication action "
+                + action.getType() + " and serialization type " +  action.getConfig().getSerializationType());
+
+        String[] uris;
+        int extSep = path.indexOf('.', path.lastIndexOf('/'));
+        if (extSep == -1) {
```
atopper, line 192: I thought about other additions here, to help make the activation more powerful and flexible for the user, but thought I'd smart small.

davidjgonzalez:
@atopper can you take a quick look at: https://github.com/davidjgonzalez/acs-aem-commons/commit/9957874e15519aed4a621627484cb5ed5702d870#
I updated to use new OSGi annotations.

Have you explored using the ReplicationContentFacade instead of a tmp file? I believe that lets you store the data in the JCR instead of a tmp file - though, i guess that would need service user access to the node? We could put them under /tmp so theyre auto-cleaned up?

https://helpx.adobe.com/experience-manager/6-3/sites/developing/using/reference-materials/javadoc/com/day/cq/replication/ReplicationContentFacade.html

atopper:
@davidjgonzalez Thanks for the review!  I didn't mean to make you do so much work!
I pushed the new OSGi annotations, and added @Designate and a few more properties to make it look good in the /system UI, and so the ContentBuilder could be found by the Agent.
I removed the AUTH_INFO.

Regarding **ReplicationContentFacade**, I like the idea - not a lot of documentation around that one but I played around with it.  I worry about the service user/ACLs that you mentioned.  I also can't find how often the /tmp folder is cleaned.  Customers could activate a lot of content, so I feel like I should implement folder buckets too.  I see createTempFile used in CQ and Granite.
How strongly do you feel that **ReplicationContentFacade** be used.  It feels like the complexity increases a lot.

davidjgonzalez:
Let me take a look at this tomorrow and see if we can't cut a release

davidjgonzalez (approved):

PR #2494, opened 2020-12-16
VersionedClientLibs broken in AEM 6.5.7

jfmitchell:
HTML 5 / HTML 5.2 made the type attribute optional. AEM 6.5.7 has taken this on board, the HTMLLibraryManager no longer adds the type attribute to script tags.

This updates the VersionClientLibs so that it will still work in both cases.


badvision:
Apart from needing to fix the pedantic code climate issue looks good to me. :)

badvision (approved):

davidjgonzalez:
Updating with master should fix the broken Travis build

davidjgonzalez (approved):

Issue #2471, opened 2020-11-09
Replacement for Redirect Map Manager for AEMaaCS


kwin:
As https://adobe-consulting-services.github.io/acs-aem-commons/features/redirect-map-manager/index.html is no longer supported in AEMaaCS due to the lack of cron job support, I would propose a different feature (which might leverage the same UI):

Implement a new servlet filter (one of the first ones in the chain) which would evaluate the redirect configuration and issue a 302 or 301 respectively in case of a matching incoming request url. The response should also carry a Cache-Control header to make it possible to cache the result in the CDN (https://experienceleague.adobe.com/docs/experience-manager-cloud-service/implementing/content-delivery/cdn.html?lang=en#content-delivery, https://docs.fastly.com/en/guides/http-status-codes-cached-by-default)

Update: I opened https://github.com/AdobeDocs/experience-manager-cloud-service.en/issues/56 to get clarification on CDN caching support of redirects, as Adobe CDN being Fastly right now is an implementation detail and might change in the future....

davidjgonzalez:
Resolved by #2513 (available in 4.12.0 and 5.0.0)

PR #2513, opened 2021-01-20
ACS Redirect Manager

YegorKozlov:
This is a replacement of the legacy  https://adobe-consulting-services.github.io/acs-aem-commons/features/redirect-map-manager/index.html which is no longer supported in AEMaaCS due to the lack of cron job support, see https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2471

The patch implements a servlet filter to issue HTTP 301/302 and infrastructure to manage the redirect configurations in AEM. This is 100% AEM solution which does depend on Dispatcher.
The verbose documentation is in another PR: https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io/pull/204


bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
```diff
@@ -0,0 +1,508 @@
...
+        "jmx.objectname=" + "com.adobe.acs.commons:type=Redirect Manager",
+        EventConstants.EVENT_TOPIC + "=" + ReplicationAction.EVENT_TOPIC,
+        ResourceChangeListener.CHANGES + "={ADDED, REMOVED, CHANGED}"
+
+})
+@Designate(ocd = RedirectFilter.Configuration.class)
+public class RedirectFilter extends AnnotatedStandardMBean
+        implements Filter, EventHandler, ResourceChangeListener, RedirectFilterMBean {
+
+    public static final String DEFAULT_STORAGE_PATH = "/var/acs-commons/redirects";
```
kwin, line 84: Due to the limitations with /var in Content-Packages for AEM Cloud I would rather move that to /conf (https://experienceleague.adobe.com/docs/experience-manager-learn/cloud-service/debugging/debugging-aem-as-a-cloud-service/build-and-deployment.html?lang=en#deploy-to)

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
```diff
@@ -0,0 +1,508 @@
...
+    static final String SERVICE_NAME = "redirect-manager";
+
+    @ObjectClassDefinition(name = "ACS Commons Redirect Filter")
+    public @interface Configuration {
+        @AttributeDefinition(name = "Enable Redirect Filter", description = "Indicates whether the redirect filter is enabled or not.", type = AttributeType.BOOLEAN)
+        boolean enabled() default true;
+
+        @AttributeDefinition(name = "Rewrite Location Header", description = "Apply Sling Resource Mappings (/etc/map) to Location header. " +
+                "Use if Location header should rewritten using ResourceResolver#map", type = AttributeType.BOOLEAN)
+        boolean rewriteUrls() default true;
```
kwin, line 100: mapUrls instead of rewriteUrls as rewrite is a bit too generic

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
```diff
@@ -0,0 +1,508 @@
...
+        @AttributeDefinition(name = "Request Paths", description = "List of paths for which redirection is allowed", type = AttributeType.STRING)
+        String[] paths() default {"/content"};
+
+        @AttributeDefinition(name = "Preserve Query String", description = "Preserve query string in redirects", type = AttributeType.BOOLEAN)
+        boolean preserveQueryString() default true;
+
+        @AttributeDefinition(name = "Storage Path", description = "The path in the repository to store redirect configurations", type = AttributeType.STRING)
+        String storagePath() default DEFAULT_STORAGE_PATH;
+
+        @AttributeDefinition(name = "On Delivery Headers", description = "Optional On-Delivery headers in the name:value format,"+
```
kwin, line 114: "Additional redirect response headers"?

On Delivery is a bit vague for me....

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
```diff
@@ -0,0 +1,508 @@
...
+     *  /var/acs-commons/redirects --> Collection<RedirectRule>
+     *
+     * @param resource the parent resource containing redirect configurations
+     * @return a list of redirect configurations . Can be empty if no redirects are
+     * configured.
+     */
+    public static Collection<RedirectRule> getRules(Resource resource) {
+        Collection<RedirectRule> rules = new ArrayList<>();
+        for (Resource res : resource.getChildren()) {
+            if(REDIRECT_RULE_RESOURCE_TYPE.equals(res.getResourceType())){
```
kwin, line 253: Please support extended resource types as well via `ResourceResolver.isResourceType()` (https://sling.apache.org/apidocs/sling9/org/apache/sling/api/resource/ResourceResolver.html#isResourceType-org.apache.sling.api.resource.Resource-java.lang.String-)

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
```diff
@@ -0,0 +1,508 @@
...
+                || !(response instanceof SlingHttpServletResponse)) {
+            chain.doFilter(request, response);
+            return;
+        }
+
+        SlingHttpServletRequest slingRequest = (SlingHttpServletRequest) request;
+        SlingHttpServletResponse slingResponse = (SlingHttpServletResponse) response;
+
+        if (isEnabled() && doesRequestMatch(slingRequest) && handleRedirect(slingRequest, slingResponse)) {
+            return;
```
kwin, line 276: Some debug log to improve debugging possibilities would help here.

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
```diff
@@ -0,0 +1,508 @@
...
+                }
+
+                log.info("Redirecting {} to {}, statusCode: {}",
+                        resourcePath, location, redirectRule.getStatusCode());
+                slingResponse.setHeader("Location", location);
+                for(String name : onDeliveryHeaders.keySet()){
+                    String value = onDeliveryHeaders.get(name);
+                    slingResponse.setHeader(name, value);
+                }
+                slingResponse.setStatus(redirectRule.getStatusCode());
```
kwin, line 330: a global fallback for cache-control response headers would help here. You can overwrite then only when deviating from the standard in the redirect rule.

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
```diff
@@ -0,0 +1,508 @@
...
+    }
+
+    protected boolean rewriteUrls() {
+        return rewriteUrls;
+    }
+
+    /**
+     * @return whether redirection is enabled
+     */
+    protected boolean isEnabled() {
```
kwin, line 352: IMHO no need to map to individual fields, just expose the full configuration which has all necessary getters

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
```diff
@@ -0,0 +1,508 @@
...
+     * e.g. only *.html requests will be redirected. Same path with .json extension
+     * will <i>not</i> be redirected. This feature is disabled by default.</li>
+     * </ol>
+     *
+     * @param request the request to check
+     * @return whether redirection for the given is allowed
+     */
+    private boolean doesRequestMatch(SlingHttpServletRequest request) {
+        WCMMode wcmMode = WCMMode.fromRequest(request);
+        if (WCMMode.EDIT == wcmMode || WCMMode.PREVIEW == wcmMode || WCMMode.DESIGN == wcmMode) {
```
kwin, line 390: I think it would be more stable to just check for `WCMMode.DISABLED != wcmMode`

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
```diff
@@ -0,0 +1,508 @@
...
+        }
+
+        String method = request.getMethod();
+        if (!getMethods().contains(method)) {
+            log.trace("Request method [{}] does not match any of {}.", method, methods);
+            return false;
+        }
+
+        String ext = request.getRequestPathInfo().getExtension();
+        if (ext != null && !getExtensions().isEmpty() && !getExtensions().contains(ext)) {
```
kwin, line 402: `!getExtensions().isEmpty()` is redundant

bundle/src/main/java/com/adobe/acs/commons/redirects/models/SubstitutionElement.java
```diff
@@ -0,0 +1,121 @@
...
+                // Static text
+                String value = targetPath.substring(pos);
+                pos = targetPath.length();
+                elements.add(new StaticElement(value));
+            } else {
+                // $: back reference to rule
+                if (dollarPos + 1 == targetPath.length()) {
+                    String value = targetPath.substring(pos, dollarPos + 1);
+                    elements.add(new StaticElement(value));
+                    log.info("invalid back reference at pos({}): {}", dollarPos, targetPath);
```
kwin, line 93: info level seems wrong here, either warn/error or preferably just throw an exception

bundle/src/main/java/com/adobe/acs/commons/redirects/servlets/ImportRedirectMapServlet.java
```diff
@@ -0,0 +1,156 @@
...
+    protected void doPost(SlingHttpServletRequest request, SlingHttpServletResponse response)
+            throws ServletException, IOException {
+
+        String path = request.getParameter("path");
+        Resource storageRoot = request.getResourceResolver().getResource(path);
+        log.debug("Updating redirect maps at {}", storageRoot.getPath());
+        Map<String, RedirectRule> jcrRules = getRules(storageRoot)
+                .stream().collect(Collectors.toMap(RedirectRule::getSource, r -> r,
+                        (oldValue, newValue) -> oldValue, LinkedHashMap::new)); // rules stored in crx
+        Map<String, RedirectRule> xlsRules = readEntries(getFile(request))
```
kwin, line 76: try with resources on the returned input stream

bundle/src/main/java/com/adobe/acs/commons/redirects/MyLocationAdjuster.java
```diff
@@ -0,0 +1,18 @@
+package com.adobe.acs.commons.redirects;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.sling.api.SlingHttpServletRequest;
+import org.osgi.service.component.annotations.Component;
+
+@Component
+public class MyLocationAdjuster implements LocationHeaderAdjuster{
```
kwin, line 8: This implementation is not useful in the main code, please move to test classes

kwin (changes requested):
@YegorKozlov Thanks a lot for that PR, looks good overall. Just some minor remarks.

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
YegorKozlov: @kwin thanks for the speedy feedback. 
+1 to move it to /conf.
 Which will be better, /conf/redirects or /conf/acs-commons/redirects ? 
I'm inclined towards /conf/acs-commons as we can re-use it for other components.

YegorKozlov: @kwin thanks for the speedy feedback. 
+1 to move it to /conf.
 Which will be better, /conf/redirects or /conf/acs-commons/redirects ? 
I'm inclined towards /conf/acs-commons as we can re-use it for other components.

davidjgonzalez: We have `/conf/acs-commons/reports` in the 5.0.0 releae branch (this too had to get moved out of `/var`) so i vote for `/var/acs-commons/redirects` to keep `/conf` clean.

This is probably too much work (and might not be as good of an idea even) - but `/conf` is?/was? for CAConfig which is applied TO content trees. So following that paradigm Redirect configs could be created under `/conf/tenant-a` and then applied to content paths (ex. `/content/tenant-a` on a more granular level - vs having a global dashboard.
Or it could be put under `/conf/global` ..

anyhow - I digress, I'm not very passionate about the above since it doesn't seem like there is much rhyme or reason to what/how stuff is created/used in `/conf` anymore :)

https://github.com/Adobe-Consulting-Services/acs-aem-commons/tree/acs-aem-commons-5.0.0/ui.content/src/main/content/jcr_root/conf/acs-commons/reports

kwin: @davidjgonzalez Can you clarify
> so i vote for /var/acs-commons/redirects to keep /conf clean.

I am confused, so you want that to stay in /var (with the known issues with Cloud Manager Deployment) or was that a typo?

davidjgonzalez: Sorry @kwin  - i had critical typo :)

I vote for `/conf/acs-commons/redirects` to keep `/conf` clean and align with the move of `/conf/acs-commons/reports`

YegorKozlov: I updated the code to store redirects /conf/acs-commons/redirects

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
YegorKozlov: renamed to mapUrls()

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
YegorKozlov: renamed to  additionalHeaders()
```java
        @AttributeDefinition(name = "Additional Response Headers", description = "Optional response headers in the name:value format to apply on delivery,"
                + " e.g. Cache-Control: max-age=3600", type = AttributeType.STRING)
        String[] additionalHeaders() default {};
```

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
YegorKozlov: The redirection fact is logged in handleRedirect:
`
               log.info("Redirecting {} to {}, statusCode: {}",
                        resourcePath, location, redirectRule.getStatusCode());
`
people can aggregate it in Splunk and build reports of popular redirects and redirects per day, etc. There are other log.info and log.debug statements across the code which help with debugging.

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
YegorKozlov: Changed

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
YegorKozlov: No, it is not redundant. The list of extensions is optional. If you don't specify then any match, i.e. empty configuration is "allow any".  This semantics will change if we remove !getExtensions().isEmpty()

bundle/src/main/java/com/adobe/acs/commons/redirects/servlets/ImportRedirectMapServlet.java
YegorKozlov: Changed

bundle/src/main/java/com/adobe/acs/commons/redirects/MyLocationAdjuster.java
YegorKozlov: removed from the code.

bundle/src/main/java/com/adobe/acs/commons/redirects/models/SubstitutionElement.java
YegorKozlov: I changed the log level to warn.  Exception is too strict IMO.  In my implementation invalid back reference evaluates to an empty string which should be fine for most users.

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
YegorKozlov: I need for mocks in junit tests.  I'd like to keep accessors to individual fields for now.

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
YegorKozlov: Changed.

YegorKozlov:
Code Climate reports 2 issues :
1. replace Hashtable with HashMap. 
This is not a problem. Hashtable  is an argument in BundleContext#registerService and I can't replace it with HashMap. Code Climate is barking a wrong tree here,
2. Cognitive complexity  of RedirectFilter#handleRedirect
It is what it is. This is the core method and it does have cognitive complexity (just a bit IMO) . I'd like to leave the code as it is.

Other than that the patch is in a good shape. Please let me know if I can do anything else to get it applied.

Regards,
Yegor

bundle/src/main/java/com/adobe/acs/commons/redirects/servlets/ExportRedirectMapServlet.java
```diff
@@ -0,0 +1,137 @@
...
+        Collection<RedirectRule> rules = RedirectFilter.getRules(root);
+        XSSFWorkbook wb = export(rules);
+
+        response.setContentType(SPREADSHEETML_SHEET);
+        response.setHeader("Content-Disposition", "attachment;filename=\"acs-redirects.xlsx\" ");
+        wb.write(response.getOutputStream());
+    }
+
+    static XSSFWorkbook export(Collection<RedirectRule> rules) {
+        XSSFWorkbook wb = new XSSFWorkbook();
```
davidjgonzalez, line 87: @YegorKozlov wow! it's like your super familiar with apache POI or something! ;)

Been a long time - hope you've been well!

YegorKozlov: Hey @davidjgonzalez 

>  it's like your super familiar with apache POI or something! ;)

Yep, wearing my Apache POI committer hat on :)

kwin (approved):

PR #2515, opened 2021-01-25
RobotsServletTest: normalize line endings before asserting responses

YegorKozlov:
ACS Commons build fails on Windows because RobotsServletTest assumes unix line separator ('\n') . This small patch normalizes the line endings before passing strings to assertEquals.


kwin:
I would actually prefer if the `RobotsServlet` would always emit `\n` (even on Windows), although all line endings are valid according to https://developers.google.com/search/reference/robots_txt. That way one cannot derive the server's OS when looking at the line endings. Will try to come up with a PR.

kwin:
Always emitting `\n` requires more invasive code changes. Therefore I took over the first part of your patch in https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/2516. Converting the txt files containing the expected result is not necessary, as they always end with `\n` (on all platforms).

kwin:
Superseded by #2516.

PR #2533, opened 2021-02-20
Resolve ClientLib before trying to get it with the HtmlLibraryManager

### LINKED ISSUE #2528: Versioned ClientLibs not working with Resource Resolver Mapping
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.5.6
* [ ] ACS AEM Commons Version: 4.11.2

### Expected Behavior

When the project contains a Resource Resolver Mapping to hide the clientlib technical path, the service should be able to find the correct path before trying to obtain the Library.

### Actual Behavior

The service is relying on the path found in the html and not trying to resolve it first.

### Steps to Reproduce

[1] Setup and instance 6.5.6 with the ACS versioned clientLibs on.
[2] Add a JCR Resource Resolver Mapping like /etc/designs/myproject-/
[3] Check if the md5 was added to the request

### Links

https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/VersionedClientlibsTransformerFactory.java#L276
It's failing on this line when is doing the getLibrary() with the incorrect path

pcastelog:
Resolves #2528

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/VersionedClientlibsTransformerFactory.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/VersionedClientlibsTransformerFactory.java
@@ -216,7 +216,7 @@ public final class VersionedClientlibsTransformerFactory extends AbstractGuavaCa
             libraryPath = path.substring(contextPath.length());
         }
 
-        String versionedPath = this.getVersionedPath(libraryPath, libraryType, request.getResourceResolver());
+        String versionedPath = this.getVersionedPath(libraryPath, libraryType, request);
 
         if (StringUtils.isNotBlank(versionedPath)) {
             if(StringUtils.isNotBlank(contextPath)) {
@@ -231,7 +231,8 @@ public final class VersionedClientlibsTransformerFactory extends AbstractGuavaCa
         return newAttributes;
     }
 
-    private String getVersionedPath(final String originalPath, final LibraryType libraryType, final ResourceResolver resourceResolver) {
+    private String getVersionedPath(final String originalPath, final LibraryType libraryType,
+            final SlingHttpServletRequest request) {
         if (originalPath.startsWith(PROXY_PREFIX) && originalPath.contains(PROXIED_STATIC_RESOURCE_PATH)) {
             log.debug("Static resource accessed via the clientlib proxy: '{}'", originalPath);
             return null;
@@ -244,7 +245,7 @@ public final class VersionedClientlibsTransformerFactory extends AbstractGuavaCa
                 libraryPath = StringUtils.substringBeforeLast(libraryPath, ".");
             }
 
-            final HtmlLibrary htmlLibrary = getLibrary(libraryType, libraryPath, resourceResolver);
+            final HtmlLibrary htmlLibrary = getLibrary(libraryType, libraryPath, request);
 
             if (htmlLibrary != null) {
                 StringBuilder builder = new StringBuilder();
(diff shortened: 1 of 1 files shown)


bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/VersionedClientlibsTransformerFactory.java
```diff
@@ -272,10 +274,26 @@ private String getVersionedPath(final String originalPath, final LibraryType lib
...
+    private String resolvePath(LibraryType libraryType, String libraryPath, ResourceResolver resourceResolver) {
+        String clientLibraryPath = resolvePathIfJcrMapping(libraryPath, resourceResolver);
+        if (StringUtils.isEmpty(clientLibraryPath)) {
+            clientLibraryPath = resolvePathIfProxied(libraryType, libraryPath, resourceResolver);
+        }
+        return clientLibraryPath;
+    }
+
+    private String resolvePathIfJcrMapping(String libraryPath, ResourceResolver resourceResolver) {
+        Resource libraryResource = resourceResolver.resolve(libraryPath);
```
kwin, line 290: Please use the two argument version, as otherwise resolve may fail (as things like request headers cannot be taken into account): https://sling.apache.org/apidocs/sling11/org/apache/sling/api/resource/ResourceResolver.html#resolve-javax.servlet.http.HttpServletRequest-java.lang.String-

bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/VersionedClientlibsTransformerFactory.java
```diff
@@ -272,10 +274,26 @@ private String getVersionedPath(final String originalPath, final LibraryType lib
...
 
+    private String resolvePath(LibraryType libraryType, String libraryPath, ResourceResolver resourceResolver) {
+        String clientLibraryPath = resolvePathIfJcrMapping(libraryPath, resourceResolver);
+        if (StringUtils.isEmpty(clientLibraryPath)) {
+            clientLibraryPath = resolvePathIfProxied(libraryType, libraryPath, resourceResolver);
+        }
+        return clientLibraryPath;
+    }
+
+    private String resolvePathIfJcrMapping(String libraryPath, ResourceResolver resourceResolver) {
```
kwin, line 289: The method name is a bit off as you always do the mapping. I would prefer `resolvePath`, also I think the methods `resolvePath` and `resolvePathIfJcrMapping` should be merged into one.

kwin (changes requested):
Thanks for the PR. I added some remarks.

bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/VersionedClientlibsTransformerFactory.java
pcastelog: I passed the request through the methods till the new one that I added

bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/VersionedClientlibsTransformerFactory.java
pcastelog: Methods merged in one "resolvePath" that will take care of JCR resolve and If Proxied

davidjgonzalez (approved):

kwin (approved):

PR #2549, opened 2021-03-15
Issue 2548

YegorKozlov:
1. Declare RedirectFilter#urlAdjuster volatile to properly handle an osgi dynamic reference
2. RedirectFilter is diabled by default and requires an OSGi to start. If the OSGi configuration is missing then users will see a warning in the UI: 
image
they still can edit the rules through.

this was also reflected in the documentation in https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io/pull/208

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
@@ -43,6 +43,7 @@ import org.osgi.framework.BundleContext;
 import org.osgi.framework.ServiceRegistration;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.ConfigurationPolicy;
 import org.osgi.service.component.annotations.Deactivate;
 import org.osgi.service.component.annotations.Modified;
 import org.osgi.service.component.annotations.Reference;
@@ -101,7 +102,8 @@ import static org.osgi.framework.Constants.SERVICE_ID;
 /**
  * A request filter that implements support for virtual redirects.
  */
-@Component(service = {Filter.class, RedirectFilterMBean.class, EventHandler.class}, property = {
+@Component(service = {Filter.class, RedirectFilterMBean.class, EventHandler.class},
+        configurationPolicy = ConfigurationPolicy.REQUIRE,property = {
         SERVICE_DESCRIPTION + "=A request filter implementing support for virtual redirects",
         SLING_FILTER_SCOPE + "=" + EngineConstants.FILTER_SCOPE_REQUEST,
         SERVICE_RANKING + ":Integer=10000",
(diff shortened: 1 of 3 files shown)


bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
```diff
@@ -101,7 +102,8 @@
 /**
  * A request filter that implements support for virtual redirects.
  */
-@Component(service = {Filter.class, RedirectFilterMBean.class, EventHandler.class}, property = {
+@Component(service = {Filter.class, RedirectFilterMBean.class, EventHandler.class},
+        configurationPolicy = ConfigurationPolicy.REQUIRE,property = {
```
kwin, line 106: missing space

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
```diff
@@ -156,7 +158,7 @@
             policy = ReferencePolicy.DYNAMIC,
             policyOption = ReferencePolicyOption.GREEDY
     )
-    private LocationHeaderAdjuster urlAdjuster;
+    private volatile LocationHeaderAdjuster urlAdjuster;
```
kwin, line 161: usually a static policy means much less overhead, as those services should not change often....

ui.apps/src/main/content/jcr_root/apps/acs-commons/components/utilities/manage-redirects/manage-redirects.html
```diff
@@ -39,6 +39,15 @@
         </betty-titlebar>
     </div>
     <div class="content-container-inner" style="width:60%; margin:0 auto;">
+        <coral-alert size="L" variant="warning" data-sly-test=${context.disabled}>
+            <coral-alert-header>OSGi CONFIGURATION MISSING</coral-alert-header>
+            <coral-alert-content>RedirecFilter is disabled and requires an OSGi configuration to start.
```
kwin, line 44: RedirecFilter -> RedirectFilter

kwin (changes requested):

kwin (approved):

YegorKozlov:
Hey @davidjgonzalez 

> Out of curiosity, any idea what sort of time this adds to request processing per say, every 1000 redirect rule entires? This looks like O(N) where N is the number of redirect rules - just wondering how this scales. (i assume stream's matching is O(N)?
> 
> I assume performance should be ~linear, making it easy to project the overhead?
> 

The performance is O(1) + O(N)  where O(1)  is the cost to match a request by path and O(N) is the cost to match by wildcard and N is the number of wildcard rules.

Path-based redirects (` /content/path1 -> /content/path2` ) are stored in a hash map keyed by source, so matching a request is a Map#get(resourcePath) operation which is O(1).

Regex rules (`/content/path/(.*) -> /content/path2`) are iterated until a match is found . This gives us O(N), i.e.  a large number of wildcard rules *may* impact performance.

The logic is in RedirectFilter#match(String requestPath) , see
https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/853b2a955311d9c8af0ac68c154c2135b48a1293/bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java#L496

P.S. You might not be the only one asking about performance impact . I'm thinking of adding a section in the documentaion.

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
abhinavsinha347: @YegorKozlov please push the same fix on vanityurlservice class for VanityUrlAdjuster referencce. @kwin can you update the bug description.

PR #2579, opened 2021-04-29
Packagers always say that it's an ACL package, even when it isn't

shsteimer:
This file is shared across all packagers, so calling it an ACL package is misleading. There is probably a better way of doing this so that it indicates the type of package based on the packager used, but going with quick and dirty for now.


kwin (approved):

davidjgonzalez (approved):

davidjgonzalez:
THanks @shsteimer !

PR #2585, opened 2021-05-05
Added older-than milliseconds feature to workflowremover 

cornsen:
to be able to remove Workflows older than a defined time period.

This is required to set the OSGi configuration that it purges workflow instances which were started for more than XY Milliseconds ago.
The existing workflow.older-than property does only allow to define a static UTC-Timestamp which is useless in a scheduled service.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/mcp/form/NumberfieldComponent.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/mcp/form/NumberfieldComponent.java
@@ -17,10 +17,21 @@
  * limitations under the License.
  * #L%
  */
-/**
- * Form components for MCP
- */
-@Version("5.3.0")
 package com.adobe.acs.commons.mcp.form;
 
-import org.osgi.annotation.versioning.Version;
+import org.osgi.annotation.versioning.ProviderType;
+
+/**
+ * date picker field component
+ */
+@ProviderType
+public class NumberfieldComponent extends FieldComponent {
+    public NumberfieldComponent(){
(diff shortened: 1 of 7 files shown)


cornsen:
Hi @davidjgonzalez ,

could you please do a review as you fixed the WorkflowRemover olderThan short time ago. I just added an additional option to define a olderThanMilliseconds which is evaluating the delta between the WF starttime and the current time. Especially for the scheduled WF-Remover this is required to e.g. remove Workflows that are older than XY Days.
As this is my first PR to this repo i hope i got everything correct.

Best,
Cornelius

davidjgonzalez:
@cornsen generally looks good, 2 things:

1. The change to the public WorkflowRemover interface is a breaking change. Can you keep the original methods, but just have them construct a WorflowRemoveConfig object from them?
2. In practice this might not make a difference, but it seems like the "noop" value for millisAgo should be -1 rather than 0.  Seems like 0 should "technically" be a valid value whereas -1 doesn't make sense (since that's the future).

cornsen:
> @cornsen generally looks good, 2 things:
> 
> 1. The change to the public WorkflowRemover interface is a breaking change. Can you keep the original methods, but just have them construct a WorflowRemoveConfig object from them?
> 2. In practice this might not make a difference, but it seems like the "noop" value for millisAgo should be -1 rather than 0.  Seems like 0 should "technically" be a valid value whereas -1 doesn't make sense (since that's the future).

Thank you for the review @davidjgonzalez . I updated the code regarding to your comments and reverted the previous methods. Also updated the noop to -1 instead of 0. 

Best,
Cornelius

davidjgonzalez (approved):

Issue #2589, opened 2021-05-21
Commit fdd9364 breaks AEM 6.4 on restart


HitmanInWis:
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.4.8
* [x] ACS AEM Commons Version: 5.0.5-SNAPSHOT - commit fdd9364abb44415da9a5de742f7e06f9e0b9af2c
* [ ] Reproducible on Latest? yes/no

### Expected Behavior

AEM can be restarted after ACS Commons is installed.

### Actual Behavior

Upon installing ACS Commons, everything works fine until the server is attempted to be restarted.  After shutdown, starting up AEM fails - the user gets the following error in browser.

```
<h2>HTTP ERROR 503 AuthenticationSupport service missing. Cannot authenticate request.</h2>

URI: | /
-- | --
503
AuthenticationSupport service missing. Cannot authenticate request.
org.apache.felix.http.base.internal.dispatch.DispatcherServlet-3f3c3e07
```

### Steps to Reproduce

Install ACS AEM Commons built from master branch, then restart AEM.

davidjgonzalez:
Verified on vanilla.

I suspect the repo init..

YegorKozlov:
Does 6.4 use the same repo-init bundles as 6.5 ?  I don't have the 6.4 distro at hand and can't verify.

davidjgonzalez:
Looks like the issue is `set properties` is not supported on 6.4.8 .. trying out a fix ...

davidjgonzalez:
@YegorKozlov btw is it expected to see:

![image](https://user-images.githubusercontent.com/1451868/119183172-b6f3be80-ba41-11eb-8434-0514fae1abdf.png)

if you don't have a RedirectFilter OSGi component active (via mandatory OSGi config?)

YegorKozlov:
@davidjgonzalez no, it should load fine. I'll check.

davidjgonzalez:
Looks like the Sling Model requires the MBean to exist, however the MBean is tired to the RedirectFilter impl which requires an OSGi config

PR #2591, opened 2021-05-22
Prevent NPE in Redirect Manager UI when RedirectFilter is not configured

YegorKozlov:
Updated Sling Models in the Redirect Manager UI  to work if  RedirectFilter is not configured. Previously they threw NPE. 

Raised by @davidjgonzalez in https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2589

If OSGi config for pid  com.adobe.acs.commons.redirects.filter.RedirectFilter is missing then user will see a yellow warning the UI

image

No OSGi config does not prevent managing redirects. Users can still add/edit/replicate redirects rules.


davidjgonzalez:
@YegorKozlov should this commit include an update to the JSP/HTL to display the message? Or was that already in place from before?

YegorKozlov:
The message is already there. This update is only to ensure no NPE in the Sling models.

davidjgonzalez (approved):

Issue #2614, opened 2021-06-19
ReplicatorQueue should not implement the Replicator interface


joerghoh:
The ReplicatorQueue should not implement the Replicator interface.

See https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2604#issuecomment-864406901, thanks to @YegorKozlov for spotting it.

kwin:
Do we really want to track each required change for #2604 individually? IMHO for consumers it just makes a difference if all four underlying issues are fixed.

joerghoh:
@kwin I think that we cannot "fix" all problems of #2604 that easy as this one.

Issue #2604, opened 2021-06-10
Cloud Manager sonar analysis is failing due to critical issues related with CQRules:CQBP-84


silvamiguelez:
## Required Information

* AEM Version: 6.5.5
* ACS Commons Version: 5.0.4
* Reproducible on Latest? yes

### Expected Behavior

Cloud Manager code analysis is ok

### Actual Behavior

Cloud Manager code analysis is failing and breaking the build as a result due to Sonar critical issues

### Steps to Reproduce

Build a project including acs-commons library dependency (included in the "all" package) and suddenly since the last Cloud Manager release (today morning) builds are failing with the following critical sonar issues:

- The product interface com.day.cq.replication.Replicator annotated with @ProviderType should not be implemented by custom code. Detected in com.adobe.acs.commons.mcp.impl.processes.renovator.ReplicatorQueue contained in /apps/acs-commons/install/acs-aem-commons-bundle-5.0.4.jar. (Critical) CQRules:CQBP-84

- The product interface org.apache.sling.api.SlingHttpServletRequest annotated with @ProviderType should not be implemented by custom code. Detected in com.adobe.acs.commons.redirectmaps.impl.FakeSlingHttpServletRequest contained in /apps/acs-commons/install/acs-aem-commons-bundle-5.0.4.jar. (Critical) CQRules:CQBP-84

- The product interface org.apache.sling.api.request.RequestPathInfo annotated with @ProviderType should not be implemented by custom code. Detected in com.adobe.acs.commons.synth.impl.SynthesizedSlingHttpServletRequest$WrappedRequestPathInfo contained in /apps/acs-commons/install/acs-aem-commons-bundle-5.0.4.jar. (Critical) CQRules:CQBP-84

- The product interface com.adobe.granite.workflow.exec.WorkItem annotated with @ProviderType should not be implemented by custom code. Detected in com.adobe.acs.commons.workflow.synthetic.impl.granite.SyntheticWorkItem contained in /apps/acs-commons/install/acs-aem-commons-bundle-5.0.4.jar. (Critical) CQRules:CQBP-84

### Links
Related CQ rule documentation: https://experienceleague.adobe.com/docs/experience-manager-cloud-manager/using/how-to-use/custom-code-quality-rules.html?lang=en#product-apis-annotated-with-providertype-should-not-be-implemented-or-extended-by-customers

joerghoh:
[...]

PR #2615, opened 2021-06-20
Issue 2614: ReplicatorQueue should not implement the Replicator interface

YegorKozlov:
`com.adobe.acs.commons.mcp.impl.processes.renovator.ReplicatorQueue` does not need to implement the Replicator interface. This class is only used to count the number of activations and deactivations inside Renovator. We can safely remove `implements Replicator` from the code. 

Unit tests had a few assumptions that ReplicatorQueue  implemented Replicator . It was used to assert the number of published/unpublished pages. These we refactored to use Mockito ArgumentCaptor.

See https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2614 and https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2604#issuecomment-864406901


YegorKozlov:
I'd like to ignore the CodeClimate warning. Getting rid of it will require a deeper refactoring and involve more classes.

PR #2626, opened 2021-06-30
#2587 - Filter-in HTML extensions for AEM Env Indicator

davidjgonzalez:


kwin (approved):

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/AemEnvironmentIndicatorFilter.java
```diff
@@ -281,6 +291,21 @@ boolean accepts(final HttpServletRequest request) {
         return true;
     }
 
+    protected boolean isUnsupportedExtension(String requestURI) {
+        if (org.apache.commons.lang3.ArrayUtils.isEmpty(allowedExtensions)) {
```
HitmanInWis, line 295: If the import above is changed (as it is now) this can simply be `ArrayUtils`

bundle/src/test/java/com/adobe/acs/commons/wcm/impl/AemEnvironmentIndicatorFilterTest.java
```diff
@@ -320,7 +322,65 @@ public void testInnerFilter() throws IOException, ServletException {
...
         assertEquals(Boolean.FALSE, req.getAttribute(AemEnvironmentIndicatorFilter.INJECT_INDICATOR_PARAMETER));
+    }
+
+    @Test
+    public void isUnsupportedExtension_Defaults() {
+        context.currentResource("/content/test");
+
+        context.registerInjectActivateService(filter);
 
+        context.request().setPathInfo("/content/test.html");
```
HitmanInWis, line 333: I assume one of these is meant to be `test.htm` instead of `test.html`?

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/AemEnvironmentIndicatorFilter.java
```diff
@@ -281,6 +291,21 @@ boolean accepts(final HttpServletRequest request) {
...
+            return false;
+        }
+
+        final PathInfo pathInfo = new PathInfo(requestURI);
+        final String extension = pathInfo.getExtension();
+
+        if (StringUtils.isBlank(extension)) {
+            return false;
```
HitmanInWis, line 303: If the request extension is not specified at all, shouldn't this return `true`?

PR #2631, opened 2021-07-01
Run builds of PRs from forks against the origin's workflow

### LINKED ISSUE #2629: Coveralls Github action only works when PR is local to acs/acs-aem-commons repo
It appears that the updated to Coveralls Github integration only works if the PR is made from a branch on: `Adobe-Consulting-Services/acs-aem-commons`

If the PR is made from a fork's branch the coveralls submission fails.

I believe this is because `steps.refs.outputs.branch_name` evaluates to a detached state (uuid/hash) when the PR is from a fork - whereas when its a "local" PR, it evaluates to the proper named branch.

steps.refs.outputs.branch_name

/cc @kwin

kwin:
This provides the action access to all read/write tokens and secrets
This closes #2629

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,6 +1,8 @@
 name: Java CI
 
-on: [push, pull_request]
+# Run PRs from forks against the origins workflow (with access to secrets)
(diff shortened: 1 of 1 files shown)


kwin:
This won't help as it even takes the code from the origin instead of from the fork (not only the workflow).

PR #2633, opened 2021-07-07
Attempt CodeCov integration

davidjgonzalez:


davidjgonzalez:
@kwin I think the CodeCov works on this (removed coveralls) - can you check the SNAPSHOT to OSSRH piece (which should only trigger on master)

davidjgonzalez:
I assume the CodeCov complaining about "no report found to compare against" gets resolved when this gets into master and branches are made off that.

.github/workflows/maven.yml
```diff
@@ -34,30 +34,51 @@ jobs:
...
+      - name: Upload code coverage to CodeCov (Main build)
+        if: ${{ matrix.isMainBuildEnv }}
+        run: bash <(curl -s https://codecov.io/bash)
+        
+      - name: Deploy SNAPSHOT to OSSRH     
+        if: ${{ matrix.isMainBuildEnv && github.ref == 'refs/heads/main' }}
+        env:
+          MAVEN_USERNAME: ${{ secrets.OSSRH_USERNAME }}
+          MAVEN_PASSWORD: ${{ secrets.OSSRH_TOKEN }}
+        run: mvn clean deploy
```
kwin, line 84: `mvn clean deploy` encompasses a regular Maven build.

kwin:
I think we should keep the approach of handling differences between branches in environment variables. Just add the coveralls upload step conditionally in addition to the Maven build leveraging env variables

.github/workflows/maven.yml
davidjgonzalez: you thinking just do a `mvn deploy:deploy --batch-mode` ?

kwin: No, just leave as before this PR and do the coveralls upload post deployment.

davidjgonzalez: Your master still have coveralls in it right? So that needs to come out

PR #2646, opened 2021-07-13
#2645: Redirect Manager: Default CaConfig Configuration not showing up 

### LINKED ISSUE #2645: Redirect Manager: Default CaConfig Configuration not showing up
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: __6.5_____
* [ ] ACS AEM Commons Version: __5.0.7-SNAPSHOT___

### Expected Behavior

The default CaConfig configuration is created by a repo-init script and should show up .

### Actual Behavior

image

It's a regression caused by #2589 
The repo-init script used to set sling:resourceType and it was commented out:
https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/7c358e8ae6451e871823111d1439973f5190a166/ui.apps/src/main/content/jcr_root/apps/acs-commons/config/org.apache.sling.jcr.repoinit.RepositoryInitializer-aem-cs.config#L12

The fix is coming

YegorKozlov:
Fix #2645

The default global configuration  created by the repo-init script does not set sling:resourceType which is required to show up in the UI.

It's a regression caused by fixing #2589 

I don't have a better idea than change the JCR-SQL that selects CaConfig configurations :

```
String sql = "SELECT * FROM [nt:unstructured] AS s WHERE ISDESCENDANTNODE([/conf]) "
+ "AND (s.[sling:resourceType]='" + REDIRECTS_RESOURCE_TYPE + "' OR s.[jcr:path]='/conf/global/settings/redirects')";
```

this way it selects the default global config (/conf/global/settings/redirects) and other nodes having the resource type. 

I know it smells :)

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/redirects/models/Configurations.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/redirects/models/Configurations.java
@@ -57,7 +57,7 @@ public class Configurations {
 
     public Collection<RedirectConfiguration> getConfigurations() {
         String sql = "SELECT * FROM [nt:unstructured] AS s WHERE ISDESCENDANTNODE([/conf]) "
-                + "AND s.[sling:resourceType]='" + REDIRECTS_RESOURCE_TYPE + "'";
+                + "AND (s.[sling:resourceType]='" + REDIRECTS_RESOURCE_TYPE + "' OR s.[jcr:path]='/conf/global/settings/redirects')";
         log.debug(sql);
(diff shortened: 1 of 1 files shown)


davidjgonzalez:
@YegorKozlov  could we deploy that node/property using a content package? I cant remember if you can specify a specific property to deploy.

YegorKozlov:
Yes, we can deploy it using a content package. 
My bad, I was testing using ui.apps and didn't see that node in ui.content

Issue #2650, opened 2021-07-21
Avoid circular reference warning with DiscoveryServiceHelper


stefan-egli:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: _______
* [x] ACS AEM Commons Version: 5.0.4
* [ ] Reproducible on Latest? yes/no

### Expected Behavior

Startup of AEM with acs-aem-commons deployed should not issue a log.error about a circular reference with DiscoveryServiceHelper (and OakDiscoveryService)

### Actual Behavior

Startup of AEM with acs-aem-commons (seen with 5.0.4 but presumably also happens with latest as that code hasn't changed) issues the following log.error:

```
ERROR [FelixLogListener] ROOT bundle org.apache.felix.scr:2.1.26 (44) Circular reference detected trying to get service [org.apache.sling.discovery.DiscoveryService, org.apache.sling.discovery.oak.OakDiscoveryService]
 stack of references: ServiceReference: [org.apache.sling.discovery.DiscoveryService, org.apache.sling.discovery.oak.OakDiscoveryService]
ServiceReference: [org.apache.sling.discovery.TopologyEventListener]
 (java.lang.Exception: stack trace)
java.lang.Exception: stack trace
	at org.apache.felix.scr.impl.ComponentRegistry.enterCreate(ComponentRegistry.java:493) [org.apache.felix.scr:2.1.26]
	at org.apache.felix.scr.impl.BundleComponentActivator.enterCreate(BundleComponentActivator.java:717) [org.apache.felix.scr:2.1.26]
	at org.apache.felix.scr.impl.manager.SingleComponentManager.getService(SingleComponentManager.java:899) [org.apache.felix.scr:2.1.26]
	at org.apache.felix.framework.ServiceRegistrationImpl.getFactoryUnchecked(ServiceRegistrationImpl.java:348)
<cut>
	at org.apache.felix.scr.impl.manager.AbstractComponentManager.registerService(AbstractComponentManager.java:984) [org.apache.felix.scr:2.1.26]
```
and
```
[...]

PR #2651, opened 2021-07-21
Rely on TopologyEvent only instead of refering to DiscoveryService which c…

stefan-egli:
…auses circular reference (#2650)

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/util/impl/DiscoveryServiceHelper.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/util/impl/DiscoveryServiceHelper.java
@@ -19,7 +19,6 @@
  */
 package com.adobe.acs.commons.util.impl;
 
-import org.apache.sling.discovery.DiscoveryService;
 import org.apache.sling.discovery.TopologyEvent;
 import org.apache.sling.discovery.TopologyEventListener;
 import org.apache.sling.discovery.TopologyView;
@@ -28,7 +27,6 @@ import org.osgi.framework.ServiceRegistration;
 import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Deactivate;
-import org.osgi.service.component.annotations.Reference;
 
 import com.adobe.acs.commons.util.ClusterLeader;
 
@@ -41,9 +39,6 @@ import com.adobe.acs.commons.util.ClusterLeader;
  */
 @Component
 public class DiscoveryServiceHelper implements TopologyEventListener {
-
-    @Reference
-    DiscoveryService discoveryService;
 
     private BundleContext bundleContext;
 
@@ -64,9 +59,6 @@ public class DiscoveryServiceHelper implements TopologyEventListener {
     @Activate
     public void activate(BundleContext bundleContext) {
         this.bundleContext = bundleContext;
-        if (discoveryService.getTopology().getLocalInstance().isLeader()) {
-            registerClusterLeader();
-        }
     }
 
     @Deactivate
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 <!-- Keep this up to date! After a release, change the tag name to the latest release -->
 [unreleased changes details]: https://github.com/Adobe-Consulting-Services/acs-aem-commons/compare/acs-aem-commons-5.0.4...HEAD
 
+## 5.0.8
+
+### Fixed
+- #2650 - Rely on TopologyEvent only instead of refering to DiscoveryService which causes circular reference errors
+
 ## 5.0.6 - 2021-06-12
 
 ### Changed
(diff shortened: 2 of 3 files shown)


stefan-egli:
@kwin , could you please have a look at this PR? It avoids a circular reference error that can be avoided: the DiscoveryServiceHelper can rely on the TopologyEvent (which it is guaranteed to get as per discovery API) - instead of also having a reference to DiscoveryService (which causes osgi to issue the circular reference log.error). Thx!

bundle/src/main/java/com/adobe/acs/commons/util/impl/DiscoveryServiceHelper.java
```diff
@@ -64,9 +59,6 @@ public void handleTopologyEvent(TopologyEvent event) {
     @Activate
     public void activate(BundleContext bundleContext) {
         this.bundleContext = bundleContext;
-        if (discoveryService.getTopology().getLocalInstance().isLeader()) {
```
kwin, line 67: How often are relevant topology events send out? How much does this change potentially defer the `ClusterLeader` service from becoming active on a leader...

stefan-egli: They are sent as soon as a topology change is detected and has settled. 
There is no difference between getting the event vs asking the DiscoveryService - the underlying state is the same.

The problem with this activate method in addition btw is that the topology might not be defined yet - in which case TopologyView.isCurrent() would return false - ie the leader flag would be based on stale/not current information. If you rely on TopologyEventHandler this would not happen.

kwin: I am not following here: In case `TopologyView.isCurrent()` would return false no `ClusterLeader` service is registered so I don't think there is a problem here.

> They are sent as soon as a topology change is detected and has settled.
The problem is rather that this service might start after a stable topology has been established already. In that case the `ClusterLeader` service would only be registered once a new topology change event is received.

The whole idea of registering the service eagerly in activate() is that otherwise it might take too long until this service figured out whether it runs on the leader or not.

stefan-egli: I was mentioning `TopologyView.isCurrent()` not `InstanceDescription.isLeader()`

stefan-egli: > The problem is rather that this service might start after a stable topology has been established already. In that case the ClusterLeader service would only be registered once a new topology change event is received.

Just to make sure we're talking about the same thing: there's both TOPOLOGY_CHANGED - but there's also TOPOLOGY_INIT. TOPOLOGY_INIT is for exactly this purpose: it sends the very first state of the topology (once it is defined) to the listener right/soon-after the listener was registered. Ie the listener doesn't have to wait for a topology change.

> The whole idea of registering the service eagerly in activate() is that otherwise it might take too long until this service figured out whether it runs on the leader or not.

Ideally we're talking a few millis

kwin: Ok, but that could be easily fixed by adding another condition. Your change affects functionality in a bad way as it defers registering `ClusterLeader`, I am just not sure by how much...

kwin: > TOPOLOGY_INIT is for exactly this purpose: it sends the very first state of the topology (once it is defined) to the listener right/soon-after the listener was registered. Ie the listener doesn't have to wait for a topology change.

Thanks, indeed that was the missing part. Then the change looks good to me.

kwin (approved):

kwin:
@stefan-egli Can you add a changelog entry?

bundle/src/main/java/com/adobe/acs/commons/util/impl/DiscoveryServiceHelper.java
stefan-egli: Ok great. I was gonna add that the difference is miniscule. In case of a standalone instance it would be merely one
[thread-switch](https://github.com/apache/sling-org-apache-sling-discovery-commons/blob/a484a7c9e70847831a41ac3b1e6168c786fd6d0c/src/main/java/org/apache/sling/discovery/commons/providers/base/ViewStateManagerImpl.java#L203) (as TopologyEvents are always sent asynchronously to avoid blocking) - and in case of a cluster it is likely that at activation time the topology hasn't even settled, so no time would be lost at all.

stefan-egli:
sure, let me find out where that is

stefan-egli:
done : https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/2651/commits/8ee1785d0253642f47969552dbfabf04622618d8

stefan-egli:
@kwin , is that code coverage -0.01% a worry?

kwin:
> is that code coverage -0.01% a worry?

That is no problem, we still need to tweak a bit the required checks after migrating from Travis to GitHub actions.

PR #2654, opened 2021-07-21
Run PRs from forked repos with readwrite token

### LINKED ISSUE #2653: GitHub Actions CI: Status not reported from forked PRs
In case the PR was raised from a forked repository the Github action run after previous approval but doesn't report the status (necessary for the required checks). This can be seen in https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/2651 where the GitHub action was successfully executed in https://github.com/Adobe-Consulting-Services/acs-aem-commons/actions/runs/1052117286 but the status not reported back.

The same is true for the Codeclimate check.

kwin:
This closes #2653


kwin:
This does no longer trigger the action nor does it checkout the right code: https://github.com/actions/checkout/issues/518

kwin:
I still don't know why the workflow is no longer triggered with the updated events...
Seems like a common issue though: https://github.community/t/pull-request-target-not-triggering-workflow/163924

kwin:
This will only fire, once it hits the master branch, as action `pull_request_target` uses the GH action of the target branch.

.github/workflows/maven.yml
```diff
@@ -26,8 +44,18 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
+      - name: Debug Event
```
kwin, line 47: This is just temporary in order to apply some post merge fixes more easily.

davidjgonzalez (approved):

PR #2667, opened 2021-08-11
Remove HALT option from MCP as it can result in corruption

davidjgonzalez:
2021-08-11 at 5 57 PM

Removed Halt from MCP process modal as this can result in repo corruption sometimes (per Leo B.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 <!-- Keep this up to date! After a release, change the tag name to the latest release -->
 [unreleased changes details]: https://github.com/Adobe-Consulting-Services/acs-aem-commons/compare/acs-aem-commons-5.0.4...HEAD
 
+# 5.0.10
+
+- #2660 - Remove Halt button from MCP as it's use can result in repository corruption.
+
 ## 5.0.8
 
 ### Fixed
--- a/ui.apps/src/main/content/jcr_root/apps/acs-commons/components/utilities/manage-controlled-processes/clientlibs/js/scriptRunner.js
+++ b/ui.apps/src/main/content/jcr_root/apps/acs-commons/components/utilities/manage-controlled-processes/clientlibs/js/scriptRunner.js
@@ -292,8 +292,15 @@ var ScriptRunner = {
                         innerHTML: response
                     },
                     footer: {
+                        innerHTML: '<button id="okButton" is="coral-button" variant="default" coral-close>Close</button>'
+
+                        /*
+                        - https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2660
+                        - Remove Halt button as this is reported to result in Oak repo corruption sometimes
+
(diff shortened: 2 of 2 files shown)


badvision:
Recommend you also set the watchdog in FAM default to -1 (meaning no watchdog, never terminate.)

It is possible to have watchdog and halt features as long as you're not breaking a commit in progress -- I advise looking at the current stack of worker threads to see if they are not in a commit somehow, and if not in a commit then allow the halt.  It's a bit tricky but ultimately it could prevent other disasters as it was intended to do when I added that feature.  Say someone starts something and they want to halt it because they meant to only do a dry run, and they catch it during the analysis phase so no changes are committed, etc.

davidjgonzalez:
@badvision Forced task.timeout to `-1` 

For now, more pressing issues than trying to inspect stack traces to safely terminate. Hopefully, the processes don't get stuck, and if there are locks in AEM, Support can help understand the cause and remediate.

kwin:
Isn't this related to https://jackrabbit.apache.org/oak/docs/dos_and_donts.html#Dont_use_Thread.interrupt?

davidjgonzalez:
@kwin  yup

badvision:
Yes more or less.

kwin:
Shouldn't we rather remove all usages of Thread.interrupt() instead of just disabling triggering it via UI?

leoberliant:
@davidjgonzalez thanks for removing the halt button. While it's a useful feature, the UI change helps to avoid unnecessary problems with the renovator tool

davidjgonzalez:
@kwin yup - should almost be done, might have one more unit test to fix.. working on it.

bundle/src/main/java/com/adobe/acs/commons/throttling/RequestThrottler.java
```diff
@@ -225,6 +225,14 @@ protected void delay(long ms) {
         try {
             Thread.sleep(ms);
         } catch (InterruptedException e) {
+            // We no longer perform thread interrupts, so something else in AEM is causing this interruption.
```
davidjgonzalez, line 228: @badvision @kwin 
this is the best way to handle this right - since w should either be restoring the interrupted status or throwing the exception (i really dont want to get into changing signatures w/ this fix)

davidjgonzalez: and sorry - that was a question, not a statement :P  ... thanks kwin!

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 <!-- Keep this up to date! After a release, change the tag name to the latest release -->
 [unreleased changes details]: https://github.com/Adobe-Consulting-Services/acs-aem-commons/compare/acs-aem-commons-5.0.4...HEAD
 
+# 5.0.10
```
kwin, line 12: IMHO next release is 5.0.8!

davidjgonzalez:
```suggestion
# 5.0.8
```

Issue #2673, opened 2021-08-20
Redirect Manager Null Pointer


hoosierPadre:
## Required Information

* [ ] AEM Version - 2021.8.5743.20210817T230726Z (ACS)
* [ ] ACS AEM Commons Version: 5.06
* [ ] Reproducible on Latest? yes

### Expected Behavior

When editing redirects in the redirect manager changes should be saved and then executed via the com.adobe.acs.commons.redirects.filter.RedirectFilter when called at the page level

### Actual Behavior 

When editing redirects in the redirect manager changes are not being saved (NPE are being thrown). Consequently redirects aren't executed when accessing paths listed at the page level filter.

20.08.2021 15:13:47.778 *WARN* [sling-oak-observation-10] org.apache.jackrabbit.oak.jcr.observation.ChangeProcessor Error while dispatching observation events for //*[11111b]@org.apache.sling.jcr.resource.internal.JcrResourceListener
java.lang.NullPointerException: null
	at com.adobe.acs.commons.redirects.filter.RedirectFilter.onChange(RedirectFilter.java:273) [com.adobe.acs.acs-aem-commons-bundle:5.0.6]
	at org.apache.sling.resourceresolver.impl.observation.BasicObservationReporter.reportChanges(BasicObservationReporter.java:211) [org.apache.sling.resourceresolver:1.7.2]
	at org.apache.sling.jcr.resource.internal.JcrResourceListener.onEvent(JcrResourceListener.java:155) [org.apache.sling.jcr.resource:3.0.22]
	at org.apache.jackrabbit.commons.observation.ListenerTracker$1.onEvent(ListenerTracker.java:190) [org.apache.jackrabbit.jackrabbit-jcr-commons:2.20.0]
	at org.apache.jackrabbit.oak.jcr.observation.ChangeProcessor.contentChanged(ChangeProcessor.java:513) [org.apache.jackrabbit.oak-jcr:1.39.0.R1887543]
	at org.apache.jackrabbit.oak.plugins.observation.FilteringDispatcher.contentChanged(FilteringDispatcher.java:52) [org.apache.jackrabbit.oak-core:1.39.0.R1887543]
	at org.apache.jackrabbit.oak.spi.commit.BackgroundObserver$1$1.call(BackgroundObserver.java:127) [org.apache.jackrabbit.oak-store-spi:1.39.0.R1887543]
[...]

PR #2674, opened 2021-08-23
#2673 NPE In Redirect Manager

YegorKozlov:
The patch is to prevent NPE when user edit redirect rules, but the filter is disabled in the OSGi configuration, i.e. OSGi config exists, but the `enable` flag is not set.,


YegorKozlov:
The build succeeds, but Java CI fails to publish the build status.

davidjgonzalez (approved):

Issue #2679, opened 2021-08-25
Renovator 5.0.6 Not Compatible with AEMaaCS


scottk0000:
## Required Information

* [ ] AEM Version 2021.8.5755.20210819T230729Z (ACS)
* [ ] ACS AEM Commons Version: 5.0.6
* [ ] Reproducible on Latest? yes

### Expected Behavior

Renovator to should be able to move pages on AEM CS from /content/pathA to /content/pathB.
This was working fine with ACS Commons 5.0.4.

### Actual Behavior

An exception is thrown
```
17.08.2021 15:13:54.900 *ERROR* [pool-19-thread-4] com.adobe.acs.commons.fam.impl.ActionManagerImpl Caught exception in task: Can not set final com.day.cq.replication.Replicator field com.day.cq.wcm.core.impl.page.PageManagerImpl.replicator to com.adobe.acs.commons.mcp.impl.processes.renovator.ReplicatorQueue
java.lang.IllegalArgumentException: Can not set final com.day.cq.replication.Replicator field com.day.cq.wcm.core.impl.page.PageManagerImpl.replicator to com.adobe.acs.commons.mcp.impl.processes.renovator.ReplicatorQueue
	at java.base/jdk.internal.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(UnsafeFieldAccessorImpl.java:167)
	at java.base/jdk.internal.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(UnsafeFieldAccessorImpl.java:171)
	at java.base/jdk.internal.reflect.UnsafeQualifiedObjectFieldAccessorImpl.set(UnsafeQualifiedObjectFieldAccessorImpl.java:83)
	at java.base/java.lang.reflect.Field.set(Field.java:780)
	at org.apache.commons.lang.reflect.FieldUtils.writeField(FieldUtils.java:523) [org.apache.commons.lang:2.6.0]
	at org.apache.commons.lang.reflect.FieldUtils.writeField(FieldUtils.java:500) [org.apache.commons.lang:2.6.0]
	at com.adobe.acs.commons.mcp.impl.processes.renovator.MovingPage.move(MovingPage.java:74) [com.adobe.acs.acs-aem-commons-bundle:5.0.6]
[...]

PR #2693, opened 2021-09-08
#2679 wrap ReplicatorQueue in Replicator via java.lang.reflect.Proxy

YegorKozlov:
A hackish attempt to fix https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2679

The idea is to use Java Reflection API to create a dynamic proxy implementing `com.day.cq.replication.Replicator`

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/renovator/MovingPage.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/renovator/MovingPage.java
@@ -21,6 +21,9 @@ package com.adobe.acs.commons.mcp.impl.processes.renovator;
 
 import com.adobe.acs.commons.fam.actions.Actions;
 
+import com.day.cq.replication.ReplicationActionType;
+import com.day.cq.replication.ReplicationOptions;
+import com.day.cq.replication.Replicator;
 import com.day.cq.wcm.api.PageManager;
 import com.day.cq.wcm.api.PageManagerFactory;
 import com.day.cq.wcm.api.WCMException;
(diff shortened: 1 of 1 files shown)


davidjgonzalez (approved):
AFIAK this is the only way to work around this other than getting the product team to allow implementations of the interface. Injecting the RQ is its own can of worms :) ... but i dont think we'll solve that without some significant re-thinking of this tool.

royteeuwen:
@davidjgonzalez  is this something you can request, to change this at product team side? Because for me this is the same issue as having your own SlingHttpServletRequest implementation. We should be able to provide this, and not have to hack it by just implementing HttpServletRequest instead of SlingHttpServletRequest. The "feature" of not being able to implement ProviderType's is not a great feature

davidjgonzalez:
@royteeuwen ill make the request - but I would not get your hopes up. :T 

For SlingHttpServletRequest we might have to start requiring/shipping with Sling Servlet Helpers [1] which does alot of the work of supporting synthetic/internal HTTP Requests (which is often the use case with implementing SlingHttpServletRequest) ... also, [1] is included OOTB In AEM CS last time i checked (though dont think its in 6.5, but havent checked lately)

[1] https://sling.apache.org/documentation/bundles/servlet-helpers.html

cziegeler:
Do you really want to have code in a common library that relies on implementation details? I can only suggest to not do this. Reflection, setting private fields are no-gos! If you want to achieve something that is not possible with the API today, start the conversation to enhance the API - but do not hack around it
Please also note that once you go this route and implement Replicator (a provider type which you shouldn't implement), then your code will break with every update to that interface. That's one of the reasons why the check exists in the first place

YegorKozlov:
> Do you really want to have code in a common library that relies on implementation details? I can only suggest to not do this. Reflection, setting private fields are no-gos! If you want to achieve something that is not possible with the API today, start the conversation to enhance the API - but do not hack around it
> Please also note that once you go this route and implement Replicator (a provider type which you shouldn't implement), then your code will break with every update to that interface. That's one of the reasons why the check exists in the first place

I agree that using reflection smells. You can consider it as a short-term fix only. 
Extending the WCM API, to which PageManager belongs, may take months or even years and we need to balance between "making it work" and "doing it right".  Even if we extend the API, it will only be available on new versions of AEM .

PR #2700, opened 2021-09-22
dependencies for AEM 6.4, update mock libraries, build with Java 17

kwin:


davidjgonzalez:
Just throwing this out there - but we should eventually get Azul builds  as this looks like the future for AEM due to all of Oracles licensing hoops  (and already supported https://experienceleague.adobe.com/docs/experience-manager-65/deploying/introduction/technical-requirements.html)

kwin:
IMHO we should build with the most common freely available JDK distribution in our CI to make sure most developers can build the current source code  locally (to ease contribution). I would argue this is rather Adoptium than Zulu. ACS AEM Commons is never build with Cloud Manager and the build distribution and the execution distribution don't have to be equal. @davidjgonzalez Any reason why you think that CI build and execution should use the same Java distribution?

davidjgonzalez:
Just that I expect a push/recommendation/whatever for AEM developers to use/prefer Azul -- so over time, I expect Azul will be "the JDK" used on most AEM developer machines (which I assume will then also be used to develop ACS AEM Commons).

kwin:
Currently there are ~200 test failures with Java 17 falling into one of these categories:

1. `java.lang.RuntimeException: Unable to initialize JCR_OAK resource resolver factory: java/security/acl/Group
Caused by: java.lang.NoClassDefFoundError: java/security/acl/Group
Caused by: java.lang.ClassNotFoundException: java.security.acl.Group`
2. `java.lang.reflect.InaccessibleObjectException: Unable to make protected native java.lang.Object java.lang.Object.clone() throws java.lang.CloneNotSupportedException accessible: module java.base does not "opens java.lang" to unnamed module @755c9148
	at com.adobe.acs.commons.errorpagehandler.cache.impl.PowerMockErrorPageCacheImplTest.setUp(PowerMockErrorPageCacheImplTest.java)`
3. `java.lang.NoClassDefFoundError: Could not initialize class org.apache.commons.lang.reflect.MemberUtils`

kwin:
Putting back to draft to wait for https://issues.apache.org/jira/browse/SLING-10972 to be able to embed the new Guava version again.

kwin:
@davidjgonzalez IMHO now would be a good timing for merge. Although I needed to touch quite some files, this is mostly tests. The changes on the implementation are actually quite few. Can you have a look at this soon?

bundle/src/main/java/com/adobe/acs/commons/errorpagehandler/cache/impl/ErrorPageCacheImpl.java
```diff
@@ -39,13 +36,16 @@
 import javax.management.openmbean.TabularDataSupport;
 import javax.management.openmbean.TabularType;
 
-import java.util.Date;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
+import org.apache.commons.lang.StringUtils;
```
davidjgonzalez, line 39:
```suggestion
import org.apache.commons.lang3.StringUtils;
```

davidjgonzalez (approved):
LGTM! Squash and merge away!

PR #2728, opened 2021-10-20
Providing a way to store SCP in experience fragments

Derthenier:
create the sling:OrderedFolder with jcr:content node to allow for Shared Component Properties


HitmanInWis:
Please merge in latest and resolve codeclimate issues

HitmanInWis (changes requested):
This PR is including more than one issue (both the localization of generic lists and SCP support for XF).  Please submit separate PRs for each feature, as they are unrelated.

PR #2706, opened 2021-09-28
Codecov configuration

davidjgonzalez:

PR #2731, opened 2021-10-21
More optimizations for Shared Component Properties

adamcin:
This patch includes changes to leverage standard Sling object patterns to greatly improve performance of the SharedValueMapValueInjector and the SharedComponentPropertiesBindingsValuesProvider when rendering a complex page.

* SharedComponentPropertiesBindingsValuesProvider uses an internal request attribute cache type for collecting Bindings computed different levels of shared properties according to their resource paths. This optimizes the BVP behavior by keeping the SCP results around for reuse across sling includes.
* The SharedValueMapValueInjector will now attempt to retrieve results from the SlingBindings when the model adaptable is a request. This greatly optimizes requests for the simple case where a sling model used within an HTL script can access any SCP-provided properties from the SlingBindings, rather having to evaluate the adaptable request or resource again. 
* For more complicated scenarios where the injector is evaluating SCP for a resource adaptable, such as when it is referenced by another model using the @ChildResource annotation, and the bindings provided by BVP are not present or are not relevant for the child resource, the injector will adapt the resource to a SharedValueMapResourceAdapter, which implicitly leverages the SlingAdaptable adapters cache so that the SCP calculation for the injector is only evaluated once. This is similar to the traditional behavior of Resource.getValueMap() and originally Resource.adaptTo(ValueMap.class).

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/SharedValueMapValueInjector.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/SharedValueMapValueInjector.java
@@ -108,46 +108,79 @@ public class SharedValueMapValueInjector implements Injector {
         // first attempt to retrieve from bindings
         final SlingBindings bindings = getBindings(adaptable);
         if (bindings != null) {
-            if (valueType == SharedComponentProperties.ValueTypes.MERGED
-                    && resource.getPath().equals(bindings.get(SharedComponentProperties.MERGED_PROPERTIES_PATH))) {
-                return (ValueMap) bindings.get(SharedComponentProperties.MERGED_PROPERTIES);
-            }
-
-            if (rootPagePath.equals(bindings.get(SharedComponentProperties.SHARED_PROPERTIES_PAGE_PATH))) {
-                final ValueMap globalVmBound = (ValueMap) bindings.get(SharedComponentProperties.GLOBAL_PROPERTIES);
-                if (valueType == SharedComponentProperties.ValueTypes.GLOBAL) {
-                    return globalVmBound;
-                }
-
-                final String sharedPropertiesPath = sharedComponentProperties.getSharedPropertiesPath(resource);
-                if (valueType == SharedComponentProperties.ValueTypes.SHARED && sharedPropertiesPath == null) {
-                    return null;
-                }
-
-                if (sharedPropertiesPath != null
-                        && sharedPropertiesPath.equals(bindings.get(SharedComponentProperties.SHARED_PROPERTIES_PATH))) {
-                    final ValueMap sharedVmBound = (ValueMap) bindings.get(SharedComponentProperties.SHARED_PROPERTIES);
-                    if (valueType == SharedComponentProperties.ValueTypes.SHARED) {
-                        return sharedVmBound;
-                    } else if (valueType == SharedComponentProperties.ValueTypes.MERGED) {
-                        return sharedComponentProperties.mergeProperties(globalVmBound, sharedVmBound, resource);
-                    }
-                } else if (sharedPropertiesPath == null && valueType == SharedComponentProperties.ValueTypes.MERGED) {
-                    return sharedComponentProperties.mergeProperties(globalVmBound, null, resource);
-                }
+            final ValueMap fromBindings = getValueMapFromBindings(bindings, valueType, resource, rootPagePath);
+            if (fromBindings != null) {
+                return fromBindings;
             }
         }
 
         switch (valueType) {
             case GLOBAL:
-                return getGlobalProperties(rootPagePath, resource);
+                return getGlobalProperties(resource);
             case SHARED:
-                return getSharedProperties(rootPagePath, resource);
+                return getSharedProperties(resource);
             case MERGED:
-                return getMergedProperties(rootPagePath, resource);
+                return getMergedProperties(resource);
+            default:
+                return null;
+        }
+    }
+
+    ValueMap getValueMapFromBindings(final SlingBindings bindings,
+                                     final SharedComponentProperties.ValueTypes valueType,
+                                     final Resource resource,
+                                     final String rootPagePath) {
+        // if the merged path in bindings matches the resource path, just assume that the merged properties in
+        // bindings are sufficient
+        if (valueType == SharedComponentProperties.ValueTypes.MERGED
+                && resource.getPath().equals(bindings.get(SharedComponentProperties.MERGED_PROPERTIES_PATH))) {
+            return Optional.ofNullable((ValueMap) bindings.get(SharedComponentProperties.MERGED_PROPERTIES))
+                    .orElse(ValueMap.EMPTY);
         }
 
-        return null;
+        // next check that the root page path matches the shared properties page path in bindings
+        // this might not match if a request wrapper was constructed for a non-child resource of the original request
+        // but which was not subjected to its own script bindings phase
+        if (!rootPagePath.equals(bindings.get(SharedComponentProperties.SHARED_PROPERTIES_PAGE_PATH))) {
+            // return null to indicate that the Injector should construct the value map
+            return null;
+        }
+
+        final ValueMap globalVmBound = (ValueMap) bindings.get(SharedComponentProperties.GLOBAL_PROPERTIES);
+        // if GLOBAL is requested, just return it from bindings
+        if (valueType == SharedComponentProperties.ValueTypes.GLOBAL) {
+            return Optional.ofNullable(globalVmBound).orElse(ValueMap.EMPTY);
+        }
+
+        final String sharedPropertiesPath = sharedComponentProperties.getSharedPropertiesPath(resource);
+        // it is possible for getSharedPropertiesPath to return null for a resource if its resource type is invalid
+        // i.e., String.EMPTY, sling:nonexisting, nt:unstructured, or /var/absolute/path/not/in/resolver/search/path
+        if (sharedPropertiesPath == null) {
+            // if SHARED is requested in this case, return ValueMap.EMPTY
+            if (valueType == SharedComponentProperties.ValueTypes.SHARED) {
+                return ValueMap.EMPTY;
+            } else { // otherwise, pass null for sharedProperties and merge with global and resource
+                return sharedComponentProperties.mergeProperties(globalVmBound, null, resource);
+            }
+        }
+
+        // it is also possible for the shared properties path to differ from the path in bindings for the same
+        // reason that shared properties page path may differ (a wrapped request for a child resource of a
+        // different resource type)
+        final ValueMap sharedVmBound;
+        if (sharedPropertiesPath.equals(bindings.get(SharedComponentProperties.SHARED_PROPERTIES_PATH))) {
+            sharedVmBound = (ValueMap) bindings.get(SharedComponentProperties.SHARED_PROPERTIES);
+        } else {
+            sharedVmBound = Optional.ofNullable(resource.getResourceResolver().getResource(sharedPropertiesPath))
+                    .map(Resource::getValueMap).orElse(ValueMap.EMPTY);
+        }
+
+        // now that shared properties have been retrieved either return them directly if SHARED is requested
+        if (valueType == SharedComponentProperties.ValueTypes.SHARED) {
+            return Optional.ofNullable(sharedVmBound).orElse(ValueMap.EMPTY);
+        } else { // or return everything merged
+            return sharedComponentProperties.mergeProperties(globalVmBound, sharedVmBound, resource);
+        }
     }
 
     private SharedComponentProperties.ValueTypes getValueType(final AnnotatedElement element) {
@@ -166,7 +199,7 @@ public class SharedValueMapValueInjector implements Injector {
     /**
      * Get shared properties ValueMap the current resource.
      */
-    protected ValueMap getSharedProperties(String rootPagePath, Resource resource) {
+    protected ValueMap getSharedProperties(Resource resource) {
         return Optional.ofNullable(sharedComponentProperties.getSharedPropertiesPath(resource))
                 .map(resource.getResourceResolver()::getResource)
                 .map(Resource::getValueMap)
@@ -176,7 +209,7 @@ public class SharedValueMapValueInjector implements Injector {
     /**
      * Get global properties ValueMap for the current resource.
      */
-    protected ValueMap getGlobalProperties(String rootPagePath, Resource resource) {
+    protected ValueMap getGlobalProperties(Resource resource) {
         return Optional.ofNullable(sharedComponentProperties.getGlobalPropertiesPath(resource))
                 .map(resource.getResourceResolver()::getResource)
                 .map(Resource::getValueMap)
@@ -186,10 +219,10 @@ public class SharedValueMapValueInjector implements Injector {
     /**
      * Get merged properties ValueMap for the current resource.
      */
-    protected ValueMap getMergedProperties(String rootPagePath, Resource resource) {
+    protected ValueMap getMergedProperties(Resource resource) {
         return sharedComponentProperties.mergeProperties(
-                getGlobalProperties(rootPagePath, resource),
-                getSharedProperties(rootPagePath, resource),
+                getGlobalProperties(resource),
+                getSharedProperties(resource),
                 resource);
     }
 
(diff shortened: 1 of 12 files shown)


bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/SharedValueMapValueInjector.java
```diff
@@ -21,106 +21,234 @@
...
+            } else { // otherwise, pass null for sharedProperties and merge with global and resource
+                return sharedComponentProperties.mergeProperties(globalVmBound, null, resource);
+            }
+        }
+
+        // it is also possible for the shared properties path to differ from the path in bindings for the same
+        // reason that shared properties page path may differ (a wrapped request for a child resource of a
+        // different resource type)
+        final ValueMap sharedVmBound;
+        if (sharedPropertiesPath.equals(bindings.get(SharedComponentProperties.SHARED_PROPERTIES_PATH))) {
```
HitmanInWis, line 196: Would it have been a simpler system to use a property like "shared-propertied-<path>" so that we would only need a single property instead of both `SHARED_PROPERTIES_PATH` and `SHARED_PROPERTIES` and then technically could have bindings for multiple resources?

HitmanInWis (changes requested):

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedPropertiesRequestCache.java
```diff
@@ -0,0 +1,68 @@
...
+import javax.script.SimpleBindings;
+import javax.servlet.ServletRequest;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Consumer;
+
+/**
+ * Simple cache for global and shared properties bindings keyed by path and persisted in a request attribute.
+ */
+final class SharedPropertiesRequestCache {
```
HitmanInWis, line 33: Is this class purposely package protected?

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedPropertiesRequestCache.java
```diff
@@ -0,0 +1,68 @@
...
+            cache = new SharedPropertiesRequestCache();
+            cache.toRequest(req);
+        }
+        return cache;
+    }
+
+    public SharedPropertiesRequestCache toRequest(ServletRequest req) {
+        SharedPropertiesRequestCache prev = (SharedPropertiesRequestCache) req.getAttribute(REQUEST_ATTRIBUTE_NAME);
+        req.setAttribute(REQUEST_ATTRIBUTE_NAME, this);
+        return prev;
```
HitmanInWis, line 66: What is the purpose of returning `prev` here?

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedValueMapResourceAdapterImpl.java
```diff
@@ -0,0 +1,58 @@
+/*
+ * #%L
+ * ACS AEM Commons Bundle
+ * %%
+ * Copyright (C) 2016 Adobe
```
HitmanInWis, line 5: nit: copyrights on new files should have current year

HitmanInWis (changes requested):

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesBindingsValuesProvider.java
```diff
@@ -57,107 +55,81 @@
...
     private static final Logger log = LoggerFactory.getLogger(SharedComponentPropertiesBindingsValuesProvider.class);
 
-    @Reference
-    private PageRootProvider pageRootProvider;
-
-    @Reference
+    /**
+     * Bind if available, check for null when reading.
+     */
+    @Reference(policyOption = ReferencePolicyOption.GREEDY, cardinality = ReferenceCardinality.OPTIONAL_UNARY)
```
HitmanInWis, line 61: Was this a change so that using a SCP annotation on a model w/o SCP configured wont blow things up?

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesBindingsValuesProvider.java
```diff
@@ -57,107 +55,81 @@
...
+    public void addBindings(final Bindings bindings) {
+        final SlingHttpServletRequest request = (SlingHttpServletRequest) bindings.get(SlingBindings.REQUEST);
+        final Resource resource = (Resource) bindings.get(SlingBindings.RESOURCE);
+        if (request != null && resource != null) {
+            final SharedPropertiesRequestCache cache = SharedPropertiesRequestCache.fromRequest(request);
+            if (sharedComponentProperties != null) {
+                setSharedProperties(bindings, resource, cache);
             } else {
-                log.debug("Page Root Provider must be configured for shared component properties to be supported");
+                log.debug("Shared Component Properties must be configured enable this provider");
```
HitmanInWis, line 73: nit: "to" enable this provider

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesBindingsValuesProvider.java
```diff
@@ -57,107 +55,81 @@
...
     @Override
-    public void addBindings(Bindings bindings) {
-        Resource resource = (Resource) bindings.get("resource");
-        if (resource != null) {
-            if (pageRootProvider != null) {
-                setSharedProperties(bindings, resource);
+    public void addBindings(final Bindings bindings) {
+        final SlingHttpServletRequest request = (SlingHttpServletRequest) bindings.get(SlingBindings.REQUEST);
+        final Resource resource = (Resource) bindings.get(SlingBindings.RESOURCE);
+        if (request != null && resource != null) {
```
HitmanInWis, line 68: Previous code only checked for `resource != null` but now we're also requiring `request` to be non-null.  Does this add any risks?

HitmanInWis (changes requested):

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/SharedValueMapValueInjector.java
```diff
@@ -21,106 +21,234 @@
...
     }
 
     /**
      * Get shared properties ValueMap the current resource.
      */
-    protected ValueMap getSharedProperties(String rootPagePath, Resource resource) {
-        String sharedPropsPath = rootPagePath + "/" + JcrConstants.JCR_CONTENT + "/" + SharedComponentProperties.NN_SHARED_COMPONENT_PROPERTIES + "/" + resource.getResourceType();
-        Resource sharedPropsResource = resource.getResourceResolver().getResource(sharedPropsPath);
-        return sharedPropsResource != null ? sharedPropsResource.getValueMap() : ValueMap.EMPTY;
+    protected ValueMap getSharedProperties(Resource resource) {
```
HitmanInWis, line 227: Should `getSharedProperties` and `getGlobalProperties` be consolidated in the new `SharedComponentPropertiesImpl` since we also need it there for `getSharedValueMapResourceAdapter`?

HitmanInWis (changes requested):

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesBindingsValuesProvider.java
adamcin: No. A BindingsValuesProvider is only executed in the context of a Sling request, and should have access to all property names defined by the SlingBindings interface. But we still want to check for null, in case, you know, whatever.

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesBindingsValuesProvider.java
adamcin: Yes, exactly. I want the annotation behavior to "failover gracefully" to the default bindings, like I'm doing with @SharedValueMapValue behaving like @ValueMapValue.

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedPropertiesRequestCache.java
adamcin: Just followed the pattern of WCMMode.toRequest(). Didn't have any other reason.

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedPropertiesRequestCache.java
adamcin: No. I think the IDE does that when I move an inner class up a level. I'll make it public.

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/SharedValueMapValueInjector.java
```diff
@@ -21,106 +21,234 @@
...
+            case MERGED:
+                return Optional.ofNullable(adapter)
+                        .map(SharedValueMapResourceAdapter::getMergedProperties)
+                        .orElseGet(() -> getMergedProperties(resource));
+            default:
+                return ValueMap.EMPTY;
+        }
+    }
+
+    ValueMap getValueMapFromBindings(final SlingBindings bindings,
```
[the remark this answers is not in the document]
adamcin: I'll correct that.

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/SharedValueMapValueInjector.java
adamcin: I suppose that would work for how we're using the bindings with the SCP API, but it would be a bit unexpected for implicit script variable names, which are taken from the binding property name. I assume they should be legal java identifiers. I'm not sure what would happen with hyphens or slashes.

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedValueMapResourceAdapterImpl.java
adamcin: they're not copyrights. they're copypastes.

bundle/src/main/java/com/adobe/acs/commons/wcm/properties/shared/impl/SharedComponentPropertiesBindingsValuesProvider.java
adamcin: will correct.

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/SharedValueMapValueInjector.java
adamcin: done.

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/SharedValueMapValueInjector.java
HitmanInWis: Ok, my only reason for thinking this was in case it somehow enabled better caching of bindings for all the things w/o overwriting each other back and forth for alternating components of different resource types.  I haven't yet run the code, so I dont know if that's applicable or not.  I just see code validating that current bindings are for the current resource, which made me think that it might be possible that bindings are flip flopping, which could be prevented via naming.

Perhaps another way we could resolve would be to put the bindings into a map object, keyed by resource path.

adamcin: It's not so much that they are flip-flopping, because if that happens the caching is still being managed by BVP at the start of the servlet execution for a requested resource. The check for the path is to protect against the case where the cached bindings AREN'T overwritten, when one might assume they should be. I.E. when a Request-adaptable model has a @ChildResourceWithRequest-annotated model member that has its own @SharedValueMapValue annotations. The BVP doesn't have a chance to execute for the request wrapper we inject for the child model, but the parent bindings are still present because we carry along the SlingBindings request attribute in the wrapper class. This means that if we naively check for the presence of a "sharedProperties" binding, it will be for the parent resource, not the current resource. We haven't actually created a new sharedProperties for the child because we are skipping the normal BindingsValuesProvider phase with the wrapped request, so there is no relevant cache value to look up.

HitmanInWis: In that situation, does the changes to the bind values from the child object affect the original request pointing at the parent object?

adamcin: I'm saying those binding values don't change at all for the child resource. The Injector operating on the child doesn't push any changes back to the request bindings. It only decides whether or not to use the bindings based on whether the path bindings match the values that would be expected for child resource. If not, it ignores the bindings altogether.

HitmanInWis: ah k, got it

HitmanInWis:
Code all looks good to me - I just need to find time to test.

HitmanInWis:
Currently testing, hoping to get thru it today :)

HitmanInWis:
Ok, I've tested this very thoroughly thru all the different code branches on AEM 6.5.  Unfortunately won't be able to get Cloud Service and 6.4 until early next week, but feeling pretty good about these optimizations being backward compatible and safe.

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/SharedValueMapValueInjector.java
```diff
@@ -21,106 +21,234 @@
...
+        // we always need a resource, if only to determine if cached global properties are valid for the resource path
+        if (resource == null) {
+            return ValueMap.EMPTY;
+        }
+
+        // the root page path is used to test the validity of bindings
+        final String rootPagePath = sharedComponentProperties.getSharedPropertiesPagePath(resource);
+        if (rootPagePath == null) {
+            // when we have a resource but no root page path, we can at least satisfy MERGED with the resource valuemap
+            return valueType == SharedComponentProperties.ValueTypes.MERGED ? resource.getValueMap() : null;
```
HitmanInWis, line 121: edge case bug here, where SCP is configured but there is no page root - the `else` case returns `null` but needs to instead return `ValueMap.EMPTY` if we dont want model objects pulling `GLOBAL` or `SHARED` `valueType` to blow up.

HitmanInWis (changes requested):

HitmanInWis (approved):

adamcin:
@davidjgonzalez I think we need to upgrade the codecov upload step in the github workflow. My fa2f2a2 commit above is unfortunately not going to affect the active workflow defined in the master branch until it is merged.

davidjgonzalez:
yeah - the github actions broke the ability to send to Codecov on PRs - was looking into it a while ago (think theres an open PR i made to test it out) .. iIRC we might have to add some override vars to get it to work? Or maybe Codecov has made updates on their side since then.

kwin:
> @davidjgonzalez any idea why the `codecov/project` test never comes back? Looks like it's preventing me from merging this for Mark.

Not sure, seems to be an edge case with PRs. We are trying out a new GH action in https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/2706/files. Feel free to comment. For the time being I no longer made that status required.

.github/workflows/maven.yml
```diff
@@ -93,12 +93,14 @@ jobs:
 
       - name: Upload code coverage to CodeCov (Main build)
         if: matrix.isMainBuildEnv
-        run: bash <(curl -s https://codecov.io/bash)
+        uses: codecov/codecov-action@v2
```
kwin, line 96: @adamcin Please revert this commit. We have https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/2706/files for that.

adamcin: @kwin reverted

Issue #2735, opened 2021-10-26
Redirect Manager does not preserve query string in external redirects


YegorKozlov:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.5.10_______
* [ ] ACS AEM Commons Version: _5.0.14____
* [ ] Reproducible on Latest? yes

### Expected Behavior
Redirect Manager should preserve query string in external redirects. For example, given a redirect
```
/content/my-site/page1 => https://www.my-site.com/targetUrl
```

a request with a query string should get redirected to ` https://www.my-site.com/targetUrl` and the query string should be preserved,.e.g.
```
http://localhost:4503/content/my-site/page1.html?a=1&b=2
should redirect to 
https://www.my-site.com/targetUrl?a=1&b=2
```
### Actual Behavior

query string in external redirects is not preserved and the example above redirects to ` https://www.my-site.com/targetUrl`

PR #2736, opened 2021-10-26
Redirect Manager:  preserve query string in external redirects

YegorKozlov:
Fix https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2735


bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
```diff
@@ -419,16 +419,16 @@ String evaluate(RedirectMatch match, SlingHttpServletRequest slingRequest){
...
-                String queryString = slingRequest.getQueryString();
-                if (queryString != null) {
-                    location = preserveQueryString(location, queryString);
-                }
-            }
             if(urlAdjuster != null){
                 location = urlAdjuster.adjust(slingRequest, location);
             }
         }
+        if(preserveQueryString) {
```
kwin, line 426: nit: missing space

kwin (approved):
LGTM, but lacks a changelog entry

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
davidjgonzalez:
```suggestion
        if (preserveQueryString) {

```

davidjgonzalez:
```suggestion
        if (preserveQueryString) {

```

davidjgonzalez (approved):
I will add changelog

PR #2740, opened 2021-11-10
add additional options.

wildone:
The idea behind these extensions is to allow replicating parts of a page without replicating the whole page to the publisher and invalidate the page path with the dispatcher. This allows updates of metadata on a page without updating whole page.

An example is if you have `jcr:content/seo` metadata that needs to be updated without replicating the whole page.

This extended version will test if the current payload has a valid child that needs to be replicated if the not current path will be replicated. If flush dispatcher has been enabled this will also replicate this or parent path as payload. Parent payload path is found by resource type that is specified.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/ReplicateWithOptionsWorkflowProcess.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/ReplicateWithOptionsWorkflowProcess.java
@@ -28,6 +28,7 @@ import com.adobe.acs.commons.util.WorkflowHelper;
 import com.adobe.acs.commons.util.visitors.ContentVisitor;
 import com.adobe.acs.commons.util.visitors.ResourceRunnable;
 import com.adobe.acs.commons.workflow.WorkflowPackageManager;
+import org.apache.jackrabbit.JcrConstants;
 import com.day.cq.replication.ReplicationActionType;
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 
 ### Fixed
 
+- #2740 - Add additional options to Replicate With Options workflow process.
```
kwin, line 16: Please add to line 13ff, 5.0.4 has been released already.

bundle/src/main/java/com/adobe/acs/commons/workflow/process/impl/ReplicateWithOptionsWorkflowProcess.java
```diff
@@ -188,5 +280,57 @@ public boolean isThrottle() {
             return throttle;
         }
 
+        public ReplicationOptions getReplicationOptionsDispatcher(Resource content) {
+            ReplicationOptions replicationOptionsDispatcher = new ReplicationOptions();
+            replicationOptionsDispatcher.setSynchronous(
+                replicationOptionsDispatcher.isSynchronous()
```
kwin, line 286: I don't get this line here? Is source and target really both the new object `replicationOptionsDispatcher`?

kwin (changes requested):

Issue #2742, opened 2021-11-11
Generic List Localization


Derthenier:
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: latest AEM SDK
* [x] ACS AEM Commons Version: 5.0.0
* [x] Reproducible on Latest? yes

### Expected Behavior

When I go to author a generic list, I should be able to specify localized titles.

### Actual Behavior

Only a single title field is available. This only allows for a single language to be saved.

### Steps to Reproduce

1. Create a new Generic List
2. Click on Add to add an item to the list.
3. Verify that only `Title` is available with no opportunity to specify localized values.

davidjgonzalez:
Fixed in 5.1.0

saurabh1140:
Hello @davidjgonzalez , This has made list very long and complex to author, what to do if we want to disable it? we want just Title and value. no localization.

ericvangeem:
@davidjgonzalez Echoing what @saurabh1140 said. This has severely degraded the authoring experience of Generic Lists.

Derthenier:
@ericvangeem / @saurabh1140  -  Have you tried overlaying `wcm/core/resources/languages` and seeing if that helps the authoring experience?
I understand the issue.
What would you think about providing a configuration for whether the generic list should be localized?

ericvangeem:
I would agree with the idea of defining a configuration to enable
localization of the titles.

SachinMali:
I think apart from 'Authorability' concerns outlined above, this Localization feature has caused the regression and specifically two issues, see https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2783, https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2765

And @Derthenier your above suggestion of overlaying 'wcm/core/resources/languages' won't help as the path is hardcoded @ https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/mcp/util/AnnotatedFieldDeserializer.java#L228

 Would you mind fixing it, and I think two ways..
1. Quick/Dirty + Hotfix: Disable the Localization support by toggling the annotation to 'false' @ https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/genericlists/GenericList.java#L87

[...]

PR #2743, opened 2021-11-11
#2742 - Providing a way to localize list titles

Derthenier:
Available languages to use for localization are taken from `wcm/core/resources/languages`. Users can customize what languages are needed by overlaying `/libs/wcm/core/resources/languages` into `/apps/wcm/core/resources/languages' and only storing the languages they need.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/mcp/form/package-info.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/mcp/form/package-info.java
@@ -20,7 +20,7 @@
 /**
  * Form components for MCP
  */
-@Version("5.3.0")
+@Version("6.0.0")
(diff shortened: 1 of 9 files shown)


bundle/src/main/java/com/adobe/acs/commons/mcp/form/package-info.java
```diff
@@ -20,7 +20,7 @@
 /**
  * Form components for MCP
  */
-@Version("5.3.0")
+@Version("6.0.0")
```
kwin, line 23: Isn't this a backwards compatible change? If not, can we convert it to one?

HitmanInWis:
Will need unit tests updated to pass the codecov test

Issue #2488, opened 2020-11-30
MCP Process manager not tracking jobs


mcihon-inwk:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.5.6.0
* [ ] ACS AEM Commons Version: 4.8.4
* [ ] Reproducible on Latest? yes

### Expected Behavior

When I go to Managed Control Processes and start a new job, I am expecting the job to show up in the process manager, where it would show the details of the process (name, description, status, errors, etc.)

### Actual Behavior

When I go to Managed Control Processes and start a new job, the job starts and runs to completion (or error), however the process manager never updates.  I can find the jobs using CRXDE Lite and check the status that way, however nothing shows up in the process manager.  I have checked the page source to see if it was a rendering issue, however there is no indication that the process manager contains any job/process information when viewing the source of the MCP pages.

### Steps to Reproduce

Browse to ACS AEM Commons, select Managed Control Processes, and start a new job.  This has occurred for all jobs I have created.

badvision:
If you're running through a dispatcher, make sure it's not caching the JSON
url used by the UI.  Ran into that recently.

mcihon-inwk:
Thanks for the suggestion.  I'm a bit new to AEM and ACS Commons.  Would you be able to point me in the direction of any relevant documentation to check this?

badvision:
Open your browser's javascript debug panel.  Look to see if you're getting any javascript errors or if it's loading a .json file correctly in the web requests.  If it's getting the same json file, then the dispatcher needs to be configured not to cache that file (we should get a PR in to fix that from MCP, too) -- if you're getting any other kind of JS error let me know.  If you're not getting any errors but still not seeing anything then you're using an old version of MCP and this is a browser bug I've already fixed; install a newer version of ACS Commons.

mcihon-inwk:
Ok, I'm not seeing any errors.  I started a dry run job and I do see an updated json file signifying the start of the job called jcr:content.start.son.  There is also another json file titled jcr:content.list.json which is always empty on every page load.  I'm guessing this is where the list of completed jobs _should_ populate and this is what you are referring to when you say the dispatcher is caching the json file.  Is this correct?

**jcr:content.start.json**
![image](https://user-images.githubusercontent.com/74873818/100796428-4222be80-33e6-11eb-832b-ea080452dfc7.png)

**jcr:content.list.json**
[...]

PR #2751, opened 2021-12-05
Fix #2749 and #2488: Manage Controlled Processes does not show any process

### LINKED ISSUE #2749: Manage Controlled Processes does not show any process
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.5.6.0
* [x] ACS AEM Commons Version: 4.12.0
* [ ] Reproducible on Latest? yes/no

### Expected Behavior

Shows a list of executed process

### Actual Behavior

Does not show any process

### Steps to Reproduce
Go to /apps/acs-commons/content/manage-controlled-processes.html
This is currently only happenning on PROD and the only way to see the list is accesing the instance through Ip

### Pictures

linkview2
ipView2

YegorKozlov:
Fix #2749 and #2488:

When serializing data using

```
gson.toJson(object, writer);
```
we need to ensure response.getWriter() is closed in the end, otherwise it might return an empty json. 

see the explanation in https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2749#issuecomment-986244291

### DIFF
--- a/bundle/src/test/java/com/adobe/acs/commons/mcp/impl/ControlledProcessManagerServletTest.java
+++ b/bundle/src/test/java/com/adobe/acs/commons/mcp/impl/ControlledProcessManagerServletTest.java
@@ -19,19 +19,48 @@
  */
 package com.adobe.acs.commons.mcp.impl;
 
+import com.adobe.acs.commons.mcp.ControlledProcessManager;
+import com.adobe.acs.commons.mcp.ProcessDefinition;
+import com.adobe.acs.commons.mcp.ProcessInstance;
+import com.adobe.acs.commons.mcp.model.ManagedProcess;
+import com.adobe.acs.commons.mcp.model.impl.ArchivedProcessInstance;
+import com.google.common.reflect.TypeToken;
 import com.google.gson.Gson;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStreamWriter;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.io.Writer;
+import java.lang.reflect.Type;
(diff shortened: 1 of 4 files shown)


kwin:
The servlet's response is automatically closed by the servlet engine. Closing explicitly might cause issues with filters https://stackoverflow.com/a/1159244. The issue must be a different one...

kwin:
Usually those issues arise if a BufferedWriter is used under the hood which is not closed, maybe the original response writer is wrapped in a filter here and the filter forgets to properly close the BuffetexWriter?

YegorKozlov:
> Usually those issues arise if a BufferedWriter is used under the hood which is not closed, maybe the original response writer is wrapped in a filter here and the filter forgets to properly close the BuffetexWriter?

I don't see any filters that could do that. If explicit close() can cause issues then we can try a different strategy: instead of passing response.getWriter() to Gson we can first serialize the entire payload into a string and then pass that string to response.getWriter() . That will work too. 

What puzzles me is why it works on local and when connected to remote IP and does not work when connected via host name. I  suspect it could be HTTP/1.1 vs HTTP2  .

bundle/src/test/java/com/adobe/acs/commons/mcp/impl/ControlledProcessManagerServletTest.java
```diff
@@ -164,6 +164,7 @@ public void test2749() throws IOException, ServletException {
 
         servlet.doGet(request, response);
 
+        writer.close(); // the servlet engine closes the response stream
```
kwin, line 167: try with resources on bos and writer instead?

davidjgonzalez:
Ill add changlog and update test fo try-resources

PR #2761, opened 2021-12-14
Switch from m-scr-p and m-bundle-p to bnd-m-p

### LINKED ISSUE #1718: Use bnd-maven-plugin instead of maven-bundle-plugin
The `bnd-maven-plugin` (https://github.com/bndtools/bnd/tree/master/maven/bnd-maven-plugin) and the `bnd-baseline-maven-plugin` (https://github.com/bndtools/bnd/tree/master/maven/bnd-baseline-maven-plugin) are nowadays better maintained than the `maven-bundle-plugin`and are always released along with the underlying `bnd` (which is also the foundation for `maven-bundle-plugin`). Also the former is designed much more in line with the Maven build life cycle (further information in https://blogs.paremus.com/2015/03/announcing-the-bnd-maven-plugin/).
Also Apache Sling tries to move away from `maven-bundle-plugin` (http://apache-sling.73963.n3.nabble.com/hackathon-bnd-maven-plugin-vs-maven-bundle-plugin-td4086273.html).

There are a few important differences to the `maven-bundle-plugin` to consider while migrating:

1. Packages are no longer automatically exported (but there is a handy directive for exporting all versioned packages)
1. In-POM instructions are supported but need to be enclosed in a `CDATA` section
1. Certain pom elements no longer end up automatically as manifest attributes (compare with https://github.com/bndtools/bnd/issues/2447, supposed to be fixed with version 4.2)

This was also mentioned in https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/1560#issuecomment-456720819

kwin:
This closes #1718


bundle/src/test/java/com/adobe/acs/commons/it/build/ScrMetadataIT.java
```diff
@@ -275,95 +229,137 @@ private DescriptorList getDescriptorsFromCurrent() throws Exception {
...
+                        metatypeDescriptors.add(parseMetatype(new InputStreamFacade(zis), entry.getName()));
                     } else if (entry.getName().startsWith("OSGI-INF/")) {
                         result.merge(parseScr(new InputStreamFacade(zis), entry.getName(), checkNs));
                     }
                 }
-
                 entry = zis.getNextEntry();
             }
         }
+        // metatype descriptors must come last (after component descriptions)
```
kwin, line 246: @justinedelson Any reason why you think that metatype attributes should overwrite the value of same named component properties? IMHO those two are not directly related. Currently it feels a bit like we a comparing apples and pears. The correct way would be to compare SCR descriptors and metatype descriptors separately, as on affects the default properties of DS components, the other one only how the Felix Web Console renders a UI for setting custom properties.

kwin:
@davidjgonzalez Can you review? I would like to merge soon...

davidjgonzalez (approved):

PR #2788, opened 2022-02-17
Defect/localize generic lists/configuarble option

SachinMali:
The recent Localization support for MCP based 'FormField' does not give the ability to consumers of ACS Commons to decide whether they want to enable it and secondly, the generated dialogs display ALL the language-based (from /libs/wcm/core/resources/languages) localized fields thus suboptimal authoring experience, for e.g. authorability of Generic List is hampered. And in case of Generic List having > 8 items throws error w/ NPE or Too Many Calls.

This fix will help to suppress the Localization behavior of MCP-based FormFields at the global level. And if someone is interested in localized fields, they can overlay the desired languages from /libs/wcm/core/resources/languages..


SachinMali:
To clarify how this will work
By default localization of MCP-based dialogs is disabled, so Generic List authoring will be normal like before.
Screen Shot 2022-02-23 at 10 59 20 AM

And whoever is interested in Localized MCP-based dialogs they can enable at Global Level via OSGi config
Screen Shot 2022-02-23 at 11 05 15 AM

Overlayed languages structure
Screen Shot 2022-02-23 at 11 06 11 AM

Then Localized Generic List authoring will look like

Screen Shot 2022-02-23 at 11 00 00 AM

Issue #2794, opened 2022-03-02
Feature review: Redirect Configuration: 'context prefix' per configuration


senn:
As redirect rules are often created and maintained by authors, there are a lot of cases where a (big) part of the path is useless and even confusing.
For example the following rules:

- /content/mysite/about/contact-us => /content/mysite/about/contact-them
- /content/mysite/blog => /content/mysite/articles

The "/content/mysite" part is often irrelevant to authors. And also tedious when they are typing out the paths instead of selecting.

The idea is to have a "context prefix" option that can be set per _main_ configuration.  This would allow all redirect rules to omit the prefix.  The above redirect rules would then be 

- /about/contact-us => /about/contact-them
- /blog => /articles

Which is much more readable and understandable for authors.  Especially in more complex cases.

The "context prefix" is optional and when not set, the rules are evaluated as normal.

YegorKozlov:
Thanks for the patch, but this functionality can already be achieved with Sling Mappings.

When enabled, Sling Mappings are used for two things:

1. To rewrite Location header, so that instead 
```Location: /content/mysite/about/contact-us```
user will get 
```Location: /about/contact-us```

2. To match ResourceResolver#map'ed path as a second try. 

RedirectFilter will first try to match  the full JCR Path from the request, e.g. /content/mysite/about/contact-us and , if it didn't match, try ResourceResolver#map'ed path (/about/contact-us) 

Example:

Given a Sling Mappings configuration  /etc/map/https/www.mysite.com

```
{
  "jcr:primaryType": "sling:Folder",
  "www.mysite.com": {
    "jcr:primaryType": "sling:Mapping",
    "sling:internalRedirect": [
      "/content/mysite"
    ]
  }
}
```

and a redirect configuration 
```
/about/contact-us => /about/contact-them
```

When evaluating http://localhost:4502/content/mysite/about/contact-us.html , RedirectFilter will try to match the full JCR path which will return null. Then it will  ResourceResolver#map it and try to match `/about/contact-us` which will match.


See https://adobe-consulting-services.github.io/acs-aem-commons/features/redirect-manager/mappings.html

royteeuwen:
@YegorKozlov  I don't see how this would work if you have multiple domains on your AEM? They all want their own redirect manager configuration, which wouldn't work if you do it like you state (unless I'm mistaken)

YegorKozlov:
@royteeuwen I don't see a conflict here. 
[...]

PR #2795, opened 2022-03-02
first version of working context-prefix mechanism

senn:
Here is a working version (PoC if you will), of the "context prefix" feature, see #2794.

- The prefix can be set, unset and updated using the UI (dialog)
image
image

- The prefix is a string property "contextprefix" on the config resource (the parent node of the individual rules)
image

- The RedirectFilter takes this prefix into account when resolving

Remark: no unit test yet as I seem to be unable to mock the configResource with a contextPrefix value.  Will have to look into that later.  Pointers welcome.


bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
```diff
@@ -409,8 +415,17 @@ boolean handleRedirect(SlingHttpServletRequest slingRequest, SlingHttpServletRes
      * {@link ResourceResolver#map(HttpServletRequest, String)}
      */
     String evaluate(RedirectMatch match, SlingHttpServletRequest slingRequest){
+        //fetches optional contextPrefix
+        final Resource configResource = configResolver.getResource(slingRequest.getResource(), config.bucketName(), config.configName());
+        final ValueMap properties = ResourceUtil.getValueMap(configResource);
+        final String contextPrefix = Optional.ofNullable(properties.get(Redirects.CFG_PROP_CONTEXT_PREFIX))
+                .orElse("").toString();
```
YegorKozlov, line 422: a better way to express this is
``` java
      ValueMap properties = configResource.getValueMap();
      String contextPrefix = properties.get(Redirects.CFG_PROP_CONTEXT_PREFIX, "");
```
Also, the finals are not necessary here.

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
```diff
@@ -549,6 +563,14 @@ RedirectMatch match(SlingHttpServletRequest slingRequest) {
                 return cfg == null ? RedirectConfiguration.EMPTY : cfg;
             });
             String resourcePath = slingRequest.getRequestPathInfo().getResourcePath(); // /content/mysite/en/page.html
+
+            final ValueMap properties = ResourceUtil.getValueMap(configResource);
+            final String contextPrefix = Optional.ofNullable(properties.get(Redirects.CFG_PROP_CONTEXT_PREFIX))
+                    .orElse("").toString();
```
YegorKozlov, line 569: a better way to express this is
``` java
      ValueMap properties = configResource.getValueMap();
      String contextPrefix = properties.get(Redirects.CFG_PROP_CONTEXT_PREFIX, "");
```
Also, the finals are not necessary here.

bundle/src/main/java/com/adobe/acs/commons/redirects/models/Redirects.java
```diff
@@ -54,6 +61,10 @@ protected void init() {
         List<Resource> all = new ArrayList<>();
         configResource.listChildren().forEachRemaining(all::add);
         pages = Lists.partition(all, pageSize);
+
+        final ValueMap properties = ResourceUtil.getValueMap(configResource);
+        contextPrefix = Optional.ofNullable(properties.get(CFG_PROP_CONTEXT_PREFIX))
+                .orElse("").toString();
```
YegorKozlov, line 67: a better way to express this is
``` java
      ValueMap properties = configResource.getValueMap();
      String contextPrefix = properties.get(Redirects.CFG_PROP_CONTEXT_PREFIX, "");
```
Also, the finals are not necessary here.

YegorKozlov:
The initial version looks good

It would be nice to cover it with unit tests. Something. like this :

```java
    @Test
    public void testContextPrefix() throws Exception {

        withRules(
                new RedirectRule("/en/one", "/en/two",
                        302, null, null));

        Resource caconfig = context.resourceResolver().getResource(redirectStoragePath);
        caconfig.adaptTo(ModifiableValueMap.class).put(Redirects.CFG_PROP_CONTEXT_PREFIX, "/content/geometrixx");

        MockSlingHttpServletResponse response = navigate("/content/geometrixx/en/one.html");

        assertEquals(302, response.getStatus());
        assertEquals("/content/geometrixx/en/two.html", response.getHeader("Location"));
        verify(filterChain, never())
                .doFilter(any(SlingHttpServletRequest.class), any(SlingHttpServletResponse.class));
    }

```

Please cover all possible cases, in particular
 - all redirects are expressed in full JCR paths
 - all redirects omit the prefix
 - there are mixed rules, some with the prefix, some without (this should be supported IMO)

YegorKozlov:
Do you mind moving the  prefix to the Create Context Aware Redirect Configuration dialog ?

image

If prefix is set we can display a message on the edit page instead of the current button :

image

YegorKozlov:
Overall looks good to me.  I'll give it a close-up shot and provide my feedback later this week. 
It would be nice to drop a few lines in the [Documentation](https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io/tree/master/_acs-aem-commons/features/redirect-manager) in another PR.

YegorKozlov:
the new feature does not work with absolute target urls, e.g.

`/en/one ==> https://adobe-consulting-services.github.io/acs-aem-commons/`

below is a breaking test:

```
    @Test
    public void testContextPrefixAbsoluteUrl() throws Exception {
        withRules(
                new RedirectRule("/en/one", "https://adobe-consulting-services.github.io/acs-aem-commons/",
                        302, null, null));

        Resource configResource = context.resourceResolver().getResource(redirectStoragePath);
        configResource.adaptTo(ModifiableValueMap.class).put(Redirects.CFG_PROP_CONTEXT_PREFIX, "/content/geometrixx");

        MockSlingHttpServletResponse response = navigate("/content/geometrixx/en/one.html");

        assertEquals(302, response.getStatus());
        assertEquals("https://adobe-consulting-services.github.io/acs-aem-commons/", response.getHeader("Location"));
        verify(filterChain, never())
                .doFilter(any(SlingHttpServletRequest.class), any(SlingHttpServletResponse.class));

    }
```

YegorKozlov:
@senn 

> The prefix assumes it will be needed in all sources or targets within that configuration. So when it's not present in the redirect rule, the prefix is concatenated to the source or target. If not, nothing happens.

Pardon me, but it wasn't obvious. It might make sense to update the "This prefix can be omitted from the redirect rules." message to reflect it. 

> I feel there's only 2 good options here:
> 
> Don't support this, so when using a context prefix ALL sources and targets should be for the same prefix (no absolute urls and no different start paths)
> Add a way to "escape" the current context prefix by using a special character (maybe "!"). This way, absolute urls and other start paths can be entered like this

If possible, I'd go for option #2. I believe there are ways to tell absolute urls from relative ones. My be a simple heuristic to check if url starts with https?:// or // will do. If it is not to be relied on then we can leave it as it is, but users need to understand that omitting prefix makes the tool a little bit less flexible.

BTW, absolute urls is not the only case. So far it was perfectly fine to return short paths in the Location header. For example, if Dispatcher only accepts short urls you want to return `/en/one` instead of `/content/mysite/en/one`  and it will not work with the content-prefix enabled.  Again, I'm fine with that as long it is clear to the users.

senn:
@YegorKozlov I guess a check can be added that verifies if a url starts with `https?://` or `//` so the context prefix will be skipped.
In cases where you want to skip the context prefix but still use a short path like `/en/one` the escape char "!" is an option.

I'll see if I can find the time in the next few days to add the following things
* a check on absolute urls
* escape char 
* unit tests for all these cases
* update the omit message to clearly show what the result of using the context-prefix is

Would you add the usage of the escape char to the omit message or is that more something for in the separate documentation?

The omit message could be something like

```
A context prefix "/content/someprefix" has been set for this configuration.
This prefix should be omitted from the redirect rules.  
Paths that start with "!" and absolute urls will ignore the context prefix.
```

YegorKozlov:
@senn 
> Would you add the usage of the escape char to the omit message or is that more something for in the separate documentation?

I would say it should be reflected in both places. 

The PR is looking very good. Thanks for the great contribution!

senn:
@YegorKozlov I just had an idea for a possibly better way of ignoring the context prefix.

Instead of prefacing a url with "!", it would be nicer (imho) to be able to set a boolean flag on a redirect rule to indicate a context prefix should be ignored.

It would look something like this:
 
image
image
image

The rest of the feature stays the same (i.e. absolute urls also ignore the prefix by default).

If you think this is a good approach, I have the code almost ready to be pushed. 
Just a few minor tweaks left + also change the export/import of redirect rules because the new property should be taken into account there as well.

YegorKozlov:
@senn a checkbox in the dialog might be a good idea. Ideally, it should only be shown if content prefix is enabled. If it is hard then it's fine to have it always. 
Also, the new checkbox should be supported in the Excel Export/Import.

senn:
@YegorKozlov I pushed the changes to use a checkbox.  This includes the changes to import/export.

I'll see if I can hide it if no prefix is set.

senn:
@YegorKozlov  that was easy enough... The checkbox is now hidden when no context prefix is set.

Small remark: if rules are made when a context prefix is set, they might have `contextPrefixIgnore=true`. These values remain when the prefix is cleared. So even when no preset is set, some rules potentially have the checkbox in the table in the overview.

I don't think that's an issue.

YegorKozlov (approved):
The PR is looking very good. The only thing I'd like to change is the potential NPE in  Excel Importer.  Other than that it's good to be merged into master.,

senn:
It looks like something went wrong with pulling the changes from remote. It appears to have overwritten the zulu->temurin change.

senn:
Hi @YegorKozlov, any idea when/if this could be merged into master and when a new version of ACS commons would be released?

YegorKozlov:
> Hi @YegorKozlov, any idea when/if this could be merged into master and when a new version of ACS commons would be released?

I'm not sure. I'd say late April / May.

davidjgonzalez:
FYI - this is in v5.3.0 which was just released

Issue #2806, opened 2022-03-29
AEM Environment Indicator rendered two times when opening a Experience Fragment variation


gquintillan:
## Required Information

* [X] AEM Version, including Service Packs, Cumulative Fix Packs, etc: AEM 6.5.12.0
* [X] ACS AEM Commons Version: 5.0.14
* [X] Reproducible on Latest? yes, tested also in 5.2.0

### Expected Behavior

The AEM Environment indicador should be rendered only one time.
![image](https://user-images.githubusercontent.com/25983819/160564816-83cfed69-045c-45d1-af04-396d8cdb2e2e.png)

### Actual Behavior

The AEM Environment indicator is shown twice when there is an iframe, for instance in the Experience Fragments editor:
![image](https://user-images.githubusercontent.com/25983819/160565236-54db6025-587a-47b3-b501-5dfa5c2ec57e.png)

### Steps to Reproduce

- Navigate to a Experience Fragment, for instance: http://localhost:4502/editor.html/content/experience-fragments/core-components-examples/library/simple-experience-fragment/master.html
- Click in an existing variation
- A new AEM Environment Indicator is rendered inside the iframe.

### Links

The related code is here: https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/AemEnvironmentIndicatorFilter.java#L229

gquintillan:
I will try to fix, and if I can't I would let you know.

gquintillan:
PR: https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/2807

davidjgonzalez:
Fixed via https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/2807

davidjgonzalez:
released in 5.3.0

PR #2807, opened 2022-03-29
#2806 added new exception on in the accepts method to avoid rendering…

gquintillan:
- Added new exception on the accepts method to avoid rendering 2 times the AEM Environment indicator when editing an Experience Fragment variation
 - StringUtils.lastIndexOf(contents, "</body>") instead of contents.indexOf("</body>") to avoid possible issues with iframes.


davidjgonzalez (approved):
Thanks!

davidjgonzalez:
released in 5.3.0

Issue #2822, opened 2022-04-11
Resolve OakPal issue #92 False Positive


adamcin:
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: n/a
* [x] ACS AEM Commons Version: 5.2.0
* [x] Reproducible on Latest? yes (when oakpal scan is uncommented in content package pom)

### Expected Behavior

oakpal should not do the thing identified here: [oakpal#92](https://github.com/adamcin/oakpal/issues/92)

### Actual Behavior

it does do that, because it doesn't automatically create `/apps` as an `nt:folder` node.

### Steps to Reproduce

* Check out the commit identified in [oakpal#92](https://github.com/adamcin/oakpal/issues/92)

* Uncomment the identified oakpal scan execution section in content/pom.xml

PR #2823, opened 2022-04-11
issue #2822 resolve oakpal 92 false positive

adamcin:
resolved issue described here: [oakpal#92](https://github.com/adamcin/oakpal/issues/92)

@kwin please review 

I included some other minor fixes after resolving the specific reported issue. 

I replaced usage of the .cnd file included in `oakpal-checks/src/main/resources` with a plugin dependency on the netcentric nodetypes artifact, matching the filevault-package-maven-plugin, as I was eliminating possible causes. 

Because the same commit converted the `acs-aem-commons-content` package into a `container` type with the OSGi configs, I discovered the current oakpal version's composite-store-alignment check assumes that containers with subpackages 1) only embed those subpackages under `/etc/packages` and 2) don't contain other content, even if the content is limited to folders and bundles/osgi configs on immutable mounts, so I excluded the `acs-aem-commons-content` package from the check scope.

When I began regression testing the project after completing the above, I had NPEs being thrown in the jaxb-api by the org.mock-server unit tests, and noticed the rather hacky mockserver-netty dependency arrangement to avoid transitive conflicts with the uber-jar. I reviewed the latest docs on https://www.mock-server.com/where/maven_central.html and figured that it would be easier to use the mockserver-netty-no-dependencies variant and add the junit rules dependency with mockserver exclusions than to exclude the dependency conflicts explicitly across multiple uber-jar versions.


kwin (commented):
LGTM, but can we remove https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/oakpal-checks/src/main/resources/nodetypes-aem640.cnd?

adamcin:
@kwin 

> LGTM, but can we remove https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/oakpal-checks/src/main/resources/nodetypes-aem640.cnd?

Sure. PR updated.

Issue #2821, opened 2022-04-09
ACS Commons 5.1.2 | Composite Multifield and Loadash behaviour


nehmalho:
After installing ACS commons 5.1.2, we are facing 2 issues. AEM Version 6.5 SP10

ISSUE 1: Below console error while authoring any component.

![image](https://user-images.githubusercontent.com/52490490/162575484-35f22c67-03d0-44ef-80ca-be0d69e45fec.png)

 
In above screenshot, the console error points to the below lines of code in multifield-extension-wrapper.js , this file was dependent on lodash.js under modern folder which was available under Libs folder (/libs/clientlibs/granite/lodash) till SP 6.5.9 and is missing in SP 6.5.10.

Now this lodash.js is brough back with ACS commons 5.1.2 under (/apps/acs-commons/clientlibs/vendor/lodash)

Temporary fix: We embed this lodash JS Category to our code base.

Concern: If ACS is providing this file, then why not adding up this on authoring mode by itself.

 

ISSUE 2: We are using an ACS AEM Commons feature called Multifield Extension for building nested multifield widgets in Touch-UI dialogs. We add the following property acs-commons-nested="" in the dialog definition on the multifiled widget node or below it.

The multifield widgets in our application having nested fields starts to break (Issue while retrieving stored multifield values) after installing 5.1.2.

Temporary fix: JS present under Composite-multifield (/apps/acs-commons/touchui-widgets/composite-multifield) folder in acs-commons folder helps to resolve the issue and restore values in multifield dialog. So we embed this Composite-multifield Category to our code base.

Concern: If ACS is providing this file, then why not adding up this on authoring mode by itself .

 

Let us know if anyone has any pointers on same.

ararat:
Would you be able to share the JS changes you implemented?
When you say break, can you elaborate?
Beyond adding the property to the node what else is the process to reproduce?

nehmalho:
hi @ararat 

The Implementation for fix we have done is just added a multifield-extension-wrapper folder as below:
Category = [cq.authoring.dialog, granite.ui.coral.foundation]
Embed = [acs-commons.lodash, acs-commons.authoring.composite-multifield]

 
![image](https://user-images.githubusercontent.com/52490490/164161639-38e86060-f5d8-4059-99c9-6bc8261a4363.png)



When you say break, can you elaborate?
[...]

PR #2828, opened 2022-04-22
replaced underscore with acs-commons.lodash based on 2821

ararat:
As suggested in  ACS Commons 5.1.2 | Composite Multifield and Loadash behaviour #2821 (https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2821) a change of the dependency of multifield has been raised.
This has been through a local build and partially tested on a local 6.5.12 + Asset search HF.
I had no time to add a multifield component to my local thus tests are partial
@wrandall22 would be a good reviewer

### DIFF
--- a/ui.apps/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/composite-multifield/.content.xml
+++ b/ui.apps/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/composite-multifield/.content.xml
@@ -4,7 +4,7 @@
           jsProcessor="[default:none,min:gcc]"
           allowProxy="{Boolean}true"
           categories="[acs-commons.authoring.composite-multifield,acs-commons.granite.ui.coral2.foundation]"
-          dependencies="[underscore]"/>
+          dependencies="[acs-commons.lodash]"/>
 <!--
 
 Wrapper Client Library definition:


ui.apps/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/composite-multifield/.content.xml
```diff
@@ -4,7 +4,7 @@
           jsProcessor="[default:none,min:gcc]"
           allowProxy="{Boolean}true"
           categories="[acs-commons.authoring.composite-multifield,acs-commons.granite.ui.coral2.foundation]"
-          dependencies="[underscore]"/>
+          dependencies="[acs-commons.lodash]"/>
```
kwin, line 7: We need to update lodash, or are you sure that we are not affected by any of those vulnerabilities: https://snyk.io/vuln/npm:lodash?

kwin (changes requested):

wrandall22:
This is more or less what I was thinking of doing. However, in my testing locally against a 6.5 instance it wasn't working as expected. Also, there are 2 other references to `underscore` which should probably be updated at the same time.

ui.apps/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/composite-multifield/.content.xml
wrandall22: Just for what it's worth, the version of lodash here is the one that used to be in AEM (e.g. in 6.4) so it isn't going backwards in security in those lower environments. Perhaps lodash was removed from 6.5 for security reasons? I haven't checked into why.

wrandall22: And, in fact, there is still this version of lodash in AEM at `/libs/clientlibs/granite/lodash/underscore` with the category `lodash.underscore`

wrandall22:
Got my local 6.5 test to work properly with multifield.

wrandall22:
The other two `underscore` references are at `ui.apps/src/main/content/jcr_root/apps/acs-commons/rte-plugins/dialog-plugin/.content.xml` and `ui.apps/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/search-based-path-browser/.content.xml`

ui.apps/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/composite-multifield/.content.xml
kwin: Then we should rather rely on the clientlib shipped with AEM as Adobe takes care of fixing vulnerabilities then. To still be compatible with AEM 66.4 we probably need to depend on both lodash categories.

ararat:
> The other two `underscore` references are at `ui.apps/src/main/content/jcr_root/apps/acs-commons/rte-plugins/dialog-plugin/.content.xml` and `ui.apps/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/search-based-path-browser/.content.xml`

Did you verify that these are affected from the same issue or is this just a search for other appearances?
As i am not able to test thoroughly i would be apprehensive of just search and replace.

ararat:
Changes done. probably worth a test but given Lodash is a superset of Underscore we should be OK.

ui.apps/src/main/content/jcr_root/apps/acs-commons/rte-plugins/dialog-plugin/.content.xml
```diff
@@ -4,7 +4,7 @@
           jsProcessor="[default:none,min:gcc]"
           allowProxy="{Boolean}true"
           categories="acs-commons.authoring.rte-plugins.dialog-plugin.coralui2"
-          dependencies="[underscore]"/>
+          dependencies="[lodash.underscore]"/>
```
kwin, line 7: Have you tried on AEM 6.4? Is that category available in that version?

wrandall22: I haven't tested this branch on 6.4, but I do see the category on 6.4 at `/libs/clientlibs/granite/lodash/underscore`

ui.apps/src/main/content/jcr_root/apps/acs-commons/touchui-widgets/composite-multifield/.content.xml
ararat: In the commit it is lodash.underscore for all 3 components so i expect that to suffice.
Verified the existence of /libs/clientlibs/granite/lodash/undescore in 6.3 as well so we are backward compatible in the support window (6.2 is no longer under support)

ui.apps/src/main/content/jcr_root/apps/acs-commons/rte-plugins/dialog-plugin/.content.xml
ararat: Same in 6.3

kwin (changes requested):
LGTM now, We should probably remove the embedded underscore version from acs aem commons now, as it may be vulnerable and is no longer in use.

kwin (approved):

PR #2835, opened 2022-05-02
Fix/2830 - Dynamic Deck Dynamo rendering fix

davidjgonzalez:
Allows a generic list to have no `jcr:content/list` without throwing NPE's as this is the initial state. A GL with no `jcr:content/list` is considered empty.


kwin (approved):
LGTM

Issue #2837, opened 2022-05-03
MCP report is blank on ACS commons version 5.2.0 on ACS 2022.4.7138


rkondaveeti:
## Required Information

* [ ] ACS AEM Commons Version: 2022.4.7138, same works on feature works on ACS version 2022.1.6198 with ACS commons version 5.2.0/5.3.0
[MCP_report_error_log.txt](https://github.com/Adobe-Consulting-Services/acs-aem-commons/files/8610406/MCP_report_error_log.txt)

* [ ] Reproducible on Latest? Yes

### Expected Behavior

After running the MCP program, when generating the report in HTML/excel it should generate/print report in/to excel/page.

### Actual Behavior

After running the MCP program, when generating the report in HTML we see a blank page and when generating the report in excel, we get a 500 error.

### Steps to Reproduce

This issue can be seen in "Asset Report" feature as well.

1. Click on "Start Process" button
2. Select "Asset Report" process
3. Accept the defaults and click Start
4. once finished, navigate to the list of processes screen "/apps/acs-commons/content/manage-controlled-processes.html"
5. Click on the "Asset Report" instance and select "view" or "Downlod (Excel)"

Installed ACS AEM SDK version "2022.4.7138" and installed ACS commons version 5.2.0. Run any utility which generates reports based on "/apps/acs-commons/components/utilities/process-instance/process-generic-report" component. We can either view the report in HTML or download the report in excel. Currently, the HTML is blank and throws 500 error while downloading the excel. Attached the error log "MCP_report_error_log.txt" for reference.

The same features are working fine in ACS AEM version "2022.1.6198".

[MCP_report_error_log.txt](https://github.com/Adobe-Consulting-Services/acs-aem-commons/files/8610396/MCP_report_error_log.txt)

### Links

_/apps/acs-commons/content/manage-controlled-processes.html_
_/apps/acs-commons/components/utilities/process-instance/process-generic-report_

YegorKozlov:
I can't reproduce it on my local "2022.4.7138" + ACS Commons v5.2.0. I also tried with 5.3.1-SNAPSHOT which is also working fine.
Are you running the report for the entire /content/dam or for a sub-folder? Is there anything special in the data ? 

Going by the exception, the report has no columns. Can you please confirm it? The exception message contains the path of report node,e.g. ` /var/acs-commons/mcp/instances/6A64C1698EE3B81B/_jcr_content/report `

[...]

PR #2846, opened 2022-05-16
#2837: use  injector-specific annotations to prevent the wrong injector from kicking in

YegorKozlov:
Fix https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2837

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/mcp/model/GenericBlobReport.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/mcp/model/GenericBlobReport.java
@@ -43,6 +43,7 @@ import org.apache.sling.api.resource.ResourceUtil;
 import org.apache.sling.api.wrappers.ValueMapDecorator;
 import org.apache.sling.models.annotations.Model;
 import org.apache.sling.models.annotations.injectorspecific.ChildResource;
+import org.apache.sling.models.annotations.injectorspecific.ValueMapValue;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -68,7 +69,7 @@ public class GenericBlobReport extends AbstractReport {
     @Inject
     String name;
 
-    @Inject
+    @ValueMapValue
     List<String> columns;
 
     private static final Logger LOG = LoggerFactory.getLogger(GenericBlobReport.class);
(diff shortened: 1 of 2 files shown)


kwin (approved):
LGTM, although I would use injector specific annotations for all fields of one model (not only multi-value ones)

bundle/src/main/java/com/adobe/acs/commons/mcp/model/GenericBlobReport.java
```diff
@@ -68,7 +69,7 @@ public class GenericBlobReport extends AbstractReport {
     @Inject
```
davidjgonzalez, line 69:
```suggestion
    @ValueMapValue
```

YegorKozlov:
> @YegorKozlov / @kwin those generic @Inject's should all be pulling from @ValueMapValue right? (I made a commit to the branch w/ this change - there were 2 other fields across the 2 fields)

I agree that the dualism smells. Unfortunately Sling Mocks only support generic @Injects and TestGenericReportExcelServlet is failing now. I will see if I can refactor it.

bundle/src/test/java/com/adobe/acs/commons/mcp/impl/TestGenericReportExcelServlet.java
```diff
@@ -39,7 +43,15 @@
 public class TestGenericReportExcelServlet {
     @Rule
     public final SlingContext slingContext = new SlingContext(ResourceResolverType.RESOURCERESOLVER_MOCK);
-    
+
+    @Before
+    public void setUp(){
+        slingContext.registerService(Injector.class, new ValueMapInjector());
```
kwin, line 49: This should be registered by default: https://github.com/apache/sling-org-apache-sling-testing-sling-mock/blob/a223de695f7ce3b55e51767ac58874a0fac08adf/core/src/main/java/org/apache/sling/testing/mock/sling/context/SlingContextImpl.java#L169-L183, same for ChildResourceInjector.

YegorKozlov: A good catch. I thought that the test was failing because these were missing, but it turned out to be ValueMapInjector 
 instead of ChildResourceInjector .

Issue #2712, opened 2021-10-06
MCP Content Fragment Import Not Working


AmoghDaryapurkar:
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: AEM SDK
* [x] ACS AEM Commons Version: 5.1.12
* [x] Reproducible on Latest? yes

### Expected Behavior

_No exception should occur during the upload process of the excel file and the content fragment should get created successfully_

### Actual Behavior

_The process is not able to create any content fragment and only creating folders_

### Steps to Reproduce

_Run the content fragment import process with required excel_

Note: Currently there is no explanation and sample present for this utility.

YegorKozlov:
Please attach the input spreadsheet

AmoghDaryapurkar:
@YegorKozlov Did you get a chance to check on this?

AmoghDaryapurkar:
@YegorKozlov - Did you get a chance to check my use case above? I am just trying to avoid writing new importer on my own.

YegorKozlov:
@AmoghDaryapurkar 

I was able to create your CFs with the following template and input spreadsheet:

[cf-events.xlsx](https://github.com/Adobe-Consulting-Services/acs-aem-commons/files/7460954/cf-events.xlsx)
[cf-template-events.zip](https://github.com/Adobe-Consulting-Services/acs-aem-commons/files/7460955/cf-template-events.zip)

http://localhost:4504/content/dam/vuse/events/bank-of-new-hampshire-pavilion-9.tidy.-1.json
```json
{
  "jcr:primaryType": "dam:Asset",
  "jcr:mixinTypes": [
    "mix:referenceable"
  ],
  "jcr:createdBy": "admin",
  "jcr:created": "Tue Nov 02 2021 16:17:08 GMT+0300",
  "jcr:uuid": "c10aec3d-0031-40b0-9b89-8acf13aa1024",
  "jcr:content": {
    "jcr:primaryType": "dam:AssetContent",
    "jcr:title": "Arizona - Alternative Rock",
    "jcr:lastModifiedBy": "admin",
    "contentFragment": true,
    "jcr:lastModified": "Tue Nov 02 2021 16:17:08 GMT+0300",
    "renditions": {
      "jcr:primaryType": "nt:folder",
      "jcr:createdBy": "admin",
      "jcr:created": "Tue Nov 02 2021 16:17:08 GMT+0300"
    },
    "related": {
[...]

PR #2870, opened 2022-06-22
Fix/2869 - Error page handler support 500s on AEM CS

davidjgonzalez:


kwin (changes requested):
@davidjgonzalez It looks like there is too many commits on this PR, some seem to refer to #2712

davidjgonzalez:
@kwin - yeh, just realized i didnt reset upstream from yegor before doing this, one sec - let me fix

davidjgonzalez:
OK - should be better now

kwin (approved):
LGTM

PR #2930, opened 2022-08-09
Providing a way to store SCP in experience fragments

Derthenier:
Enabling shared component nodes in experience fragment folder structure


davidjgonzalez:
@Derthenier looks good - are there possibilities of conflicting w/ people that are already overlaying this space? Could we collide/overwrite with some else's customization?

Derthenier:
It is possible that it will create conflicts for people extending the experience fragments

davidjgonzalez:
I wonder if we just package this up and link to it from the docs page?

PR #2934, opened 2022-08-17
Fix/2927 - Cloud Manager reporting issued with location of clientlib resources

davidjgonzalez:


content/pom.xml
```diff
@@ -24,31 +24,32 @@
...
+                        <cloudManagerTarget>all</cloudManagerTarget>
+                    </properties>
                     <embeddeds>
                         <embedded>
                             <groupId>${project.groupId}</groupId>
                             <artifactId>acs-aem-commons-bundle</artifactId>
-                            <target>/apps/acs-commons/install</target>
-                            <filter>true</filter>
-                            <isAllVersionsFilter>true</isAllVersionsFilter>
+                            <type>zip</type>
```
kwin, line 37: I don‘t understand this. The bundle does not have type zip. Also it is no application package but an OSGi bundle

davidjgonzalez: This is closed - I accidentally added the other fix to an unfinished branch I was working on. Yes, that is incorrect. :)

davidjgonzalez: @kwin i was trying to recall if there was a reason we were still using subpackages tho - I could've sworn we had moved to embeddeds?

kwin: subpackage installation works way better in AEM 6.5 (as then the asynchronous OSGi installer is not involved).
For AEM Cloud it doesn't make a difference as both are anyhow converted (in the same fashion) through https://github.com/apache/sling-org-apache-sling-feature-cpconverter.

davidjgonzalez: Gotcha - it was messing up the build when i deployed ACS Commons source direct to CM to run the Quality Pipeline on it ... i figured it had something to do w/ 6.5 and package manager providing a more robust mechanism than jcr installer

PR #2937, opened 2022-08-23
Feature: Providing a way to clean up package installs

jfmitchell:
When you use Adobe Cloud Manager to deploy new code to your environments on Adobe Managed Service instances, it will generate a unique version number for your package, based on the configuration in the pom.xml file.
The downside of this approach, is that over time, multiple versions of the packages are building up in the /etc/packages part of the repository. Eventually, this leads to the exhaustion of the storage

This is a configurable scheduled job that will only run on classic installations that can be used to clear down specific packages.


bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
```diff
@@ -0,0 +1,126 @@
...
+                session.logout();
+            }
+        }
+        LOG.info("Package Garbage Collector job finished");
+        return JobResult.OK;
+    }
+
+    private boolean isPackageOldEnough(JcrPackage jcrPackage, Integer maxAgeInDays) throws RepositoryException, IOException {
+        DateTimeFormatter formatter = DateTimeFormatter.ofLocalizedDate(FormatStyle.MEDIUM);
+        Period maxAge = Period.ofDays(maxAgeInDays);
```
kwin, line 104: IMHO only package versions which are not currently installed should ever be eligible for GC

jfmitchell: Fair to a degree - I can see your point for the "latest version" of installed packages that are older than the defined configuration. I.e. - there a deployment cycle that for some reason happens to be longer than the the configured max age. I can see this happening perhaps during change freezes where a normal two week deployment. turns into 2 months. To a degree, this could be handled by careful configuration of the max age parameter....but I do agree, it might be good to never remove the latest one...

Or is this not what you were referring to?

kwin: The latest one is not necessarily the one which is installed. Please put a guard in place that never removes the currently installed one (no matter whether it is the latest or some other version).  Here is some code example which should give you an idea how to implement such a check

```
/**
     * Verifies if a package is installed. A package is considered installed if it has a lastUnpacked property
     * and no other version of that package was installed at a more recent date.
     *
     * @param package
     * @return {@code true} if the package is installed
     */
    public boolean isInstalled(Collection<ContentPackage> installedPackages, Logger logger) {

        Optional<ContentPackage> lastInstalledPackageOptional = installedPackages.stream()
                .filter(this::isSameGroupAndName)
                .sorted(Collections.reverseOrder(Comparator.comparing(ContentPackage::getLastUnpacked)))
                .findFirst();
        if (!lastInstalledPackageOptional.isPresent()) {
            logger.info("Package " + getName() + " not found in any version");
            return false;
        }
        ContentPackage lastInstalledPackage = lastInstalledPackageOptional.get();
        
        if (lastInstalledPackage.getLastUnpacked() == null) {
            logger.info("Package " + getName() + " is present but has not been installed yet");
            return false;
        }
        
        boolean isInstalled = hasSamePid(lastInstalledPackage);
        if (!isInstalled) {
            logger.info("Package " + getName() + " was last installed with another version "
                    + lastInstalledPackage.getVersion() + " at " + DateTimeFormatter.ISO_DATE_TIME.format(lastInstalledPackage.getLastUnpacked()));
        }
        return isInstalled;
    }
```

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionConfig.java
```diff
@@ -0,0 +1,38 @@
...
+
+@ObjectClassDefinition(name = "ACS Commons - Package Garbage Collection Configuration", description = "Used to config a package garbage collection job")
+public @interface PackageGarbageCollectionConfig {
+    @AttributeDefinition(name = "Schedule", description = "Cron expression detailing when the garbage collection is run. Default runs at 02:30 every day")
+    String scheduler() default "0 30 2 ? * * *";
+
+    @AttributeDefinition(name = "Package Group Name", description = "The group name of the packages to remove")
+    String groupName() default "";
+
+    @AttributeDefinition(name = "Max age of package", description = "Packages older than this will be removed. Default is 60")
```
joerghoh, line 33:
```suggestion
    @AttributeDefinition(name = "Max age of package", description = "Packages older than this (in days) will be removed. Default is 60 days")
```

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
```diff
@@ -0,0 +1,126 @@
...
+
+        try (ResourceResolver resourceResolver = resourceResolverFactory.getServiceResourceResolver(
+            Collections.singletonMap(ResourceResolverFactory.SUBSERVICE, SERVICE_USER))) {
+            session = resourceResolver.adaptTo(Session.class);
+            JcrPackageManager packageManager = packaging.getPackageManager(session);
+            List<JcrPackage> packages = packageManager.listPackages(groupName, false);
+
+            for (JcrPackage jcrPackage : packages) {
+                String packageDescription = getPackageDescription(jcrPackage);
+                if (isPackageOldEnough(jcrPackage, maxAgeInDays)) {
```
joerghoh, line 83: if you have not deployed for 60 days, you will remove the latest package as well. Probably this will work, too, but I would always keep the latest version in the system, even if it's older than 60 days.

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
```diff
@@ -0,0 +1,126 @@
...
+                    packageManager.remove(jcrPackage);
+                    LOG.info("Deleted package {}", packageDescription);
```
joerghoh, line 85: I am afraid that this just lists "customer package" (not containing any version number or other useful information). Can you add the jcr path and the version?

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
```diff
@@ -0,0 +1,126 @@
...
+                } else {
+                    LOG.info("Not removing package because it's not old enough {}", packageDescription);
```
joerghoh, line 87: I would downgrade this to DEBUG. Don't log no-actions.

```suggestion
                    LOG.debug("Not removing package because it's not old enough {}", packageDescription);
```

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
```diff
@@ -0,0 +1,126 @@
...
+                }
+            }
+        } catch (LoginException | RepositoryException | IOException e) {
+            LOG.error("Unable to clear packages due to {}", e.toString());
```
joerghoh, line 91: Is this exception message good enough to pinpoint the reason just by error message? Can you include the package where the exception happened to the message? Also I think that a stacktrace can contain useful information, because the amount of work covered by this try-catch block is non-trivial, so additional information where exactly the code failed is useful

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
```diff
@@ -0,0 +1,126 @@
...
+            return JobResult.FAILED;
+        } finally {
+            if (session != null) {
```
joerghoh, line 94: not need to close the session explicitly. As you are adapting this session from the ResourceResolver (which is implicitly closed by this try-with-resource approach) this session is closed as well. You can remove this code.

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
```diff
@@ -0,0 +1,126 @@
...
+                session.logout();
+            }
+        }
+        LOG.info("Package Garbage Collector job finished");
```
joerghoh, line 98: Do you think it's useful if you include the number of packages you have removed?

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
```diff
@@ -0,0 +1,126 @@
...
+        return JobResult.OK;
+    }
+
+    private boolean isPackageOldEnough(JcrPackage jcrPackage, Integer maxAgeInDays) throws RepositoryException, IOException {
+        DateTimeFormatter formatter = DateTimeFormatter.ofLocalizedDate(FormatStyle.MEDIUM);
```
joerghoh, line 103: please make this a static field to this class. No need to create it on every invocation.

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
```diff
@@ -0,0 +1,126 @@
...
+        Period maxAge = Period.ofDays(maxAgeInDays);
+        LocalDate oldestAge = LocalDate.now().minus(maxAge);
+        Calendar packageCreatedAtCalendar = jcrPackage.getPackage().getCreated();
+        LocalDate packageCreatedAt = LocalDateTime.ofInstant(
+            packageCreatedAtCalendar.toInstant(),
+            packageCreatedAtCalendar.getTimeZone().toZoneId()).toLocalDate();
+        String packageDescription = getPackageDescription(jcrPackage);
+
+        if (LOG.isDebugEnabled()) {
```
joerghoh, line 112: just log, no need to check.

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
```diff
@@ -0,0 +1,126 @@
...
+            LOG.debug("Checking if package is old enough: Name: {}, Created At: {}, Oldest Age: {}",
+                packageDescription, packageCreatedAt.format(formatter), oldestAge.format(formatter));
+        }
+        return packageCreatedAt.isBefore(oldestAge) || packageCreatedAt.isEqual(oldestAge);
```
joerghoh, line 116: Not sure if this is easier to understand...
```suggestion
        return !packageCreatedAt.isAfter(oldestAge);
```

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    private String getPackageDescription(JcrPackage jcrPackage) throws RepositoryException {
+        Node packageNode = jcrPackage.getNode();
+        if (packageNode != null) {
+            return packageNode.getPath();
+        }
+        return "Unknown package";
```
joerghoh, line 124: I don't think that the code receiving this message is capable to handle this return string... Can you return ```null``` and let the invoker check the return value?

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionScheduler.java
```diff
@@ -0,0 +1,85 @@
...
+    @Reference(target="(distribution=classic)")
+    RequireAem requireAem;
+
+    ScheduledJobInfo job;
+
+    @Activate
+    protected void activate(PackageGarbageCollectionConfig config) {
+        job = scheduleJob(config);
+        if (LOG.isInfoEnabled()) {
+            LOG.info("Next scheduled run at {}", job.getNextScheduledExecution());
```
joerghoh, line 60: Please include some job details (e.g. the name of the package to check). This is a factory, so you might see this message multiple times.

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionScheduler.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @Reference(target="(distribution=classic)")
+    RequireAem requireAem;
+
+    ScheduledJobInfo job;
+
+    @Activate
+    protected void activate(PackageGarbageCollectionConfig config) {
+        job = scheduleJob(config);
+        if (LOG.isInfoEnabled()) {
```
joerghoh, line 59: No need to check the log level. 
https://stackoverflow.com/questions/8444266/even-with-slf4j-should-you-guard-your-logging

bundle/src/test/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJobTest.java
```diff
@@ -0,0 +1,131 @@
...
+    void mockPackaging(Integer daysAgo, boolean fail) {
+        try {
+            JcrPackageManager packageManager = mock(JcrPackageManager.class);
+            when(packaging.getPackageManager(any())).thenReturn(packageManager);
+            JcrPackage mockPackage = mockPackage(daysAgo);
+            when(packageManager.listPackages(anyString(), anyBoolean())).thenReturn(Collections.singletonList(mockPackage));
+            if (fail) {
+                doThrow(new RepositoryException()).when(packageManager).remove(any());
+            }
+        } catch (RepositoryException | IOException e) {
```
joerghoh, line 105: no need to do that, just throw the exception; any non-caught exception makes the test fail

joerghoh (changes requested):

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionScheduler.java
davidjgonzalez: @joerghoh I swear CM code quality reports on you if you don't - do you recall if this is the case?

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
jfmitchell: This is just used by the log messages - and will only happen if there is an issue accessing the meta data from the repository.

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionScheduler.java
jfmitchell: I included cos the base sonar rules complain if you log without checking - so I think CM also does complain.

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
jfmitchell: Added logic to cater for this guard.

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
jfmitchell: Added guard for this

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
jfmitchell: Have replaced with the exception stacktrace

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionScheduler.java
jfmitchell: Done

bundle/src/test/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJobTest.java
jfmitchell: Removed

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
```diff
@@ -0,0 +1,216 @@
...
+                        LOG.info("Deleted package {}", packageDescription);
+                    } else {
+                        LOG.info("Not removing package because it's the current installed one {}", packageDescription);
+                    }
+                } else {
+                    LOG.debug("Not removing package because it's not old enough {}", packageDescription);
+                }
+            }
+        } catch (LoginException | RepositoryException | IOException e) {
+            LOG.error("Unable to clear packages", e);
```
joerghoh, line 104: You might have cleared packages already when you reach this point. It could make sense to log this information as well, when you fail the job here.
Also, is it possible to log the information at which package the process failed?

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionScheduler.java
```diff
@@ -0,0 +1,85 @@
...
+    JobManager jobManager;
+
+    @Reference(target="(distribution=classic)")
+    RequireAem requireAem;
+
+    ScheduledJobInfo job;
+
+    @Activate
+    protected void activate(PackageGarbageCollectionConfig config) {
+        job = scheduleJob(config);
```
joerghoh, line 58: What happens if you have the system running and it does terminate without proper shutdown? I think you will end up with the same job running twice after the next restart.
So I would check on startup if such a job is already present and unregister all of them. And then register a new one. (That's probably easier to implement than checking existing jobs if they have the correct schedule).
This also improves the handling in case you are running on a cluster.

I recently updated the Sling documentation covering this aspect: https://sling.apache.org/documentation/bundles/apache-sling-eventing-and-job-handling.html#scheduled-jobs-1

joerghoh (changes requested):

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
jfmitchell: Will log an info level message to show when a package is started to be checked, and if any packages have been removed - I'll log this in this catch block.

joerghoh (approved):
LGTM

Issue #2939, opened 2022-08-30
[Redirect Manager] Properties for existing redirect rules are overwritten when XLS file is imported


sozores:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 2022.7.8085.20220725T140323Z-220600
* [ ] ACS AEM Commons Version: 5.3.2
* [ ] Reproducible on Latest? yes

### Expected Behavior

When importing a XLS file via the Redirect Manager Importer, those rules that are not modified in the XLS are not modified nor recreated.

### Actual Behavior

When importing a XLS file via the Redirect Manager Importer, all existent rules stored in CRX are deleted and recreated even if no update is required.

### Steps to Reproduce

1. Create a new Redirect Rule via the UI
2. Observe in the CRX the node name of this property and its properties
3. Import a new Redirect Rule via the Importer UI (different from the one created in step 1) 
4. Observe in the CRX the redirect rule has been renamed and initial properties `jcr:created` and `jcr:createdBy` are missing.

### Links

[1] Feature doc page: https://adobe-consulting-services.github.io/acs-aem-commons/features/redirect-manager/index.html
[2] Redirect Rules Import Servlet: https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/06d2c1c5e82545feabba8251e6f3a0ee503964e4/bundle/src/main/java/com/adobe/acs/commons/redirects/servlets/ImportRedirectMapServlet.java#L108

YegorKozlov:
@sozores Thanks for this bug  report. The PR with the fix is https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/2943

Issue #2938, opened 2022-08-29
[Redirect Manager] Redirect rules imported from excel file do not store `jcr:created` and `jcr:createdBy` properties


sozores:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 2022.7.8085.20220725T140323Z-220600
* [ ] ACS AEM Commons Version: 5.3.2
* [ ] Reproducible on Latest? yes

### Expected Behavior

When importing redirect rules from an excel file:
- the rules imported from the excel file contain the creation date and creator user information in CRX properties `jcr:created` and `jcr:createdBy` respectively
- the rules created via UI, keep the values of these properties at the moment of creation.

### Actual Behavior

When redirect rules are imported from an excel file:
- existent redirect rules are overwritten and do not longer contain the properties `jcr:created` and `jcr:createdBy`
- redirect rules in CRX do not contain the `jcr:created` and `jcr:createdBy`, although included in the code (link [2])

### Steps to Reproduce

1. Navigate to the redirect manager tool
2. Create a new redirect rule
3. Observe in CRX properties `jcr:created` and `jcr:createdBy` are set
4. Import an excel file with new redirect rules
3. Observe in CRX the previous rules added from the UI no longer contain `jcr:created` and `jcr:createdBy` properties and either the new rules imported via excel file contain them

### Links

[1] Feature doc page: https://adobe-consulting-services.github.io/acs-aem-commons/features/redirect-manager/index.html
[2] Redirect Rules Import Servlet: https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/06d2c1c5e82545feabba8251e6f3a0ee503964e4/bundle/src/main/java/com/adobe/acs/commons/redirects/servlets/ImportRedirectMapServlet.java#L129
[3] Rules added from UI
<img width="887" alt="rules_created_from_UI" src="https://user-images.githubusercontent.com/10310287/187178136-cad235dc-38cc-4024-977a-dcb06d19dbb1.png">
[4] Rules imported from Excel
<img width="875" alt="rules_from_imported_file" src="https://user-images.githubusercontent.com/10310287/187178248-622e2536-102f-4a69-9f4a-dcd1a6f4302f.png">

YegorKozlov:
@sozores Thanks for this bug  report. The PR with the fix is https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/2943

Issue #2936, opened 2022-08-23
Expose the "Redirect Creator" property in the Redirect Manager tool


sozores:
Expose the "Redirect Creator" property in the Redirect Manager tool, will help identify easily who created the redirect.
This property is already available among the redirect properties, the aim of this ticket is to expose it in the UI and also include a new column in the exported excel file.

### Expected Behavior
'jcr:createdBy' property is exposed:
- in the UI in a new column "Created by"
<img width="1388" alt="RDM-createBy-UIproposal" src="https://user-images.githubusercontent.com/10310287/186146234-c792813b-09b6-42d0-8e74-8981ed5f0c28.png">

- in the exported excel file in a new column "Created by (read only)"

### Links
Feature doc page: https://adobe-consulting-services.github.io/acs-aem-commons/features/redirect-manager/index.html

YegorKozlov:
This sounds like a good improvement, we can support the Created By property in next release. 

Please note that in the current implementation the jcr:createdBy property is only set when editing redirects from Touch UI. Rules imported from an Excel spreadsheet won't have it. We can also fix it and set jcr:createdBy  from the current session.

YegorKozlov:
@sozores Thanks for this feature request. The PR with the fix is https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/2943

PR #2943, opened 2022-09-05
Redirect Manager: support jcr:created and jcr:createdBy in Touch UI and Excel export

YegorKozlov:
The PR addresses https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2939, https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2938 and https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2936

Summary of changes: 

1. added a 'Created By' column in Touch UI and Excel Exports
2. Refactored Import From Excel to preserve existing nodes and custom properties. The old impl deleted and then re-created redirects and it could result in a loss of properties.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectRule.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectRule.java
@@ -19,13 +19,17 @@
  */
 package com.adobe.acs.commons.redirects.models;
 
+import com.adobe.granite.security.user.util.AuthorizableUtil;
 import org.apache.sling.api.resource.Resource;
 import org.apache.sling.api.resource.ValueMap;
 import org.apache.sling.models.annotations.Model;
+import org.apache.sling.models.annotations.Optional;
+import org.apache.sling.models.annotations.injectorspecific.Self;
+import org.apache.sling.models.annotations.injectorspecific.ValueMapValue;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.inject.Inject;
+import javax.annotation.PostConstruct;
 import java.lang.invoke.MethodHandles;
 import java.time.ZonedDateTime;
 import java.util.Calendar;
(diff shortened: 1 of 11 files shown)


bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectRule.java
```diff
@@ -129,13 +170,13 @@ public Pattern getRegex() {
     }
 
     public ZonedDateTime getUntilDate() {
-        return untilDate;
+        return zonedUntilDate;
     }
 
     @Override
     public String toString() {
         return String.format("RedirectRule{source='%s', target='%s', statusCode=%s, untilDate=%s, note=%s, contextPrefixIgnored=%s}",
```
joerghoh, line 178: Because all properties were included here before, you should do the same with the ```createdBy``` property and include it in the ``´toString()```  implementation.

bundle/src/main/java/com/adobe/acs/commons/redirects/servlets/ImportRedirectMapServlet.java
```diff
@@ -126,9 +133,23 @@ void update(Resource root, Collection<RedirectRule> rules) throws PersistenceExc
...
+
+            Resource redirect = jcrRedirects.get(rule.getSource());
+            if(redirect == null){
+                // add mix:created, AEM will initialize jcr:created and jcr:createdBy from the current session
+                props.put(JCR_MIXINTYPES, MIX_CREATED);
+                String nodeName = ResourceUtil.createUniqueChildName(root, "redirect-rule-");
+                resolver.create(root, nodeName, props);
+            } else {
+                // add mix:lastModified so that AEM updates jcr:lastModified and jcr:lastModifiedBy
+                ValueMap valueMap = redirect.adaptTo(ModifiableValueMap.class);
```
joerghoh, line 145: if you don't have write permissions in this node, this will return ``` null``` and the next line will throw a NPE. You should add a check here and abort the whole process if the user does not have this permission.

ui.apps/src/main/content/jcr_root/apps/acs-commons/components/utilities/manage-redirects/redirect-row/redirect-row.html
```diff
@@ -9,13 +9,17 @@
       ${properties.target}
    </td>
    <td is="coral-table-cell" class="statusCode" data-value="${properties.statusCode}" style="width:50px;">${properties.statusCode}</td>
+   <td is="coral-table-cell" class="createdBy"
```
joerghoh, line 12: there is an inconsistency in the use of the model class for the existing fields and the new ```createdBy``` field. Can you harmonize them?

joerghoh (changes requested):

bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectRule.java
YegorKozlov: I removed the zonedUntilDate property to avoid confusion, the code can do without it.

bundle/src/main/java/com/adobe/acs/commons/redirects/servlets/ImportRedirectMapServlet.java
YegorKozlov: the code now throws PersistenceException, same exception as would happen when creating redirects via resourceResolver.create

ui.apps/src/main/content/jcr_root/apps/acs-commons/components/utilities/manage-redirects/redirect-row/redirect-row.html
YegorKozlov: I refactored the code to use the model class across the component.

davidjgonzalez (approved):

Issue #2960, opened 2022-10-05
SharedComponentPropertiesBindingsValuesProvider should support LazyBindings


joerghoh:
## Required Information

* AEM Version, including Service Packs, Cumulative Fix Packs, etc: AEM 6.5, AEM CS
* ACS AEM Commons Version: all
*  Reproducible on Latest? yes

### Expected Behavior

The SharedComponentPropertiesBindingsValuesProvider always resolves all values before they are added to the bindings; this happens irrespective if these values are required or not. 
When using Lazybindings (https://sling.apache.org/apidocs/sling12/org/apache/sling/api/scripting/LazyBindings.html) this could be avoided, and values are only resolved when they are needed.

### Actual Behavior

the values are always computed, leading into performance degradation.

adamcin:
@joerghoh TL;DR: To take advantage of LazyBindings, we have to implement the `Supplier` type it introduces. Unless we use reflection to access the LazyBindings type, this will require an upgrade of our `org.apache.sling.api` dependency version to at least `2.22.0`, which will require dropping support for AEM versions prior to 6.5 SP7.

`org.apache.sling.api.scripting.LazyBindings` was introduced in `org.apache.sling.api:2.22.0`. Between this and its previous release artifact, the `org.apache.sling.api.scripting` package exported version was bumped from `2.3.2` to `2.5.0`, which will introduce a breaking change to the Import-Package range for deployment to previous AEM releases (the current version of `org.apache.sling.api` we depend on is `2.16.4`, which exports the `.scripting` package at version `2.3.1`) . The latest version of this package in the [apache/sling-org-apache-sling-api master branch](https://github.com/apache/sling-org-apache-sling-api/blob/92b15740e433e44c815dd9739b62ecdfaa713db8/src/main/java/org/apache/sling/api/scripting/package-info.java#L20) is still `2.5.0`.

[...]

PR #2962, opened 2022-10-08
issue #2960 shared-component-properties add support for lazy bindings

adamcin:
This is a hack approach to resolving #2960 in a way that doesn't force a significant change in the supported AEM version.

NOTE: Please see #2964 for non-hacky approach (against the 6.0.0 branch) that bumps the minimum AEM version to 6.5.7 and uses `LazyBindings` types properly.

Yet to be tested in AEM 6.4, 6.5 < SP7, 6.5 >= SP7,  and Cloud Service.

The benefits of this approach:
1. Should significantly improve page rendering performance in more recent versions of AEM 6.5 and Cloud Service where SCP is enabled.
2. Avoids a rather drastic increase of the minimum supported AEM version for SCP from 6.4.0 to 6.5.7.
3. Uses of reflection should be easy to prune without major refactoring when the minimum supported version of the project is raised past 6.5.7.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 <!-- Keep this up to date! After a release, change the tag name to the latest release -->
 [unreleased changes details]: https://github.com/Adobe-Consulting-Services/acs-aem-commons/compare/acs-aem-commons-5.0.14...HEAD
 
+### Added
+
 - #2941 - Add Query Builder support in Report Builder
+
+### Fixed
+
+- #2960 - SharedComponentPropertiesBindingsValuesProvider should support LazyBindings
 
 ## 5.3.4 - 2022-08-22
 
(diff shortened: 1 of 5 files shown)


adamcin:
Ready for review. @davidjgonzalez @HitmanInWis @joerghoh

joerghoh:
This is a real hack[TM]. 

@adamcin I appreciate the effort you put into this PR to make it somehow work, but I am not sure if this is worth the effort. Maybe we should just wait until ACS AEM Commons drops support for older versions of AEM than AEM 6.5 SP7?

@davidjgonzalez Any plans to cut the support older SPs?

kwin:
I would suggest requiring 6.5.x for ACS AEM Commons 6

davidjgonzalez:
@joerghoh @kwin im all on board with 6.5.x+ for AEM 6.0.0 ... i wouldnt even be adverse to specifying a higher level SP if it provides us more flexibility. One that that comes to mind is if AEM 6.5.x started supporting OSGi R7, which IIRC would let us move fully off Felix SCR without breaking OSGi properties that have `-` in them (which would be a big back-breaking change for alot of those older services)

PR #2964, opened 2022-10-09
issue #2960 shared-component-properties add support for lazy bindings, require AEM 6.5.7+

adamcin:
This is a proper approach to resolving https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2960 in a way that requires an increase in the minimum supported AEM version to 6.5 SP7. 

I rebased [the master PR branch for pre-6.5.7 support](https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/2962) and then added a commit to upgrade the platform dependencies to 6.5.7 and replace the use of reflection with proper LazyBinding types.

@joerghoh @kwin @HitmanInWis @davidjgonzalez


adamcin:
So... I think something related to these two links is causing some unit tests to break in java17 that pass in java11, AFTER upgrading to `org.apache.sling.api` 2.22.0:

[SLING-8116 ValueMap - implement default methods using OSGI Converter](https://issues.apache.org/jira/browse/SLING-8116)

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/doc-files/ValueBased.html

PR #2972, opened 2022-10-19
Update MS Office Asset Selector to work in an AEM as a Cloud Service environment - bundle change

LucaNerlich:
Website PR: https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io/pull/236

- Add `event.stopImmediatePropagation();` to stop the event from triggering twice and adding the image twice to the PowerPoint slide.
- Hide "more Details" button to prevent the author from breaking out of the asset picker
- Add isAuthorMode Check to the asset-selector.html template
  - the office.js library crashes when loaded in the author env, which in return prohibits authors from opening the page and the components dialog itself
  - load the asset-selector page with `?wcmmode=disabled` -> set in add-in manifest
  - e.g `<bt:Url id="Taskpane.Url" DefaultValue="http://localhost:4502/content/asset-selectors/assetpicker.html?wcmmode=disabled"/>`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 <!-- Keep this up to date! After a release, change the tag name to the latest release -->
 [unreleased changes details]: https://github.com/Adobe-Consulting-Services/acs-aem-commons/compare/acs-aem-commons-5.0.14...HEAD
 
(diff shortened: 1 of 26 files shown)


LucaNerlich:
Question: do "we" really not support let and const? Check / Build is complaining. It works fine in AEM itself.

davidjgonzalez (approved):

PR #2979, opened 2022-11-03
Add crx credentials to content-package-maven-plugin

senn:
I noticed that when deploying locally the bundle is installed correctly but ui.apps, content, ... are not installed if the default password or user was changed in the pom.

Adding these credential properties to the configuration of the content-package-maven-plugin in the base pom solves this.


kwin (approved):
Thanks, LGTM

davidjgonzalez (approved):

PR #3000, opened 2022-12-02
Remove macOS bug due to bug in GA plugin

davidjgonzalez:
macOS builds always fail due to bug in GA plugin.


kwin:
Let us rather switch to https://github.com/marketplace/actions/publish-test-results

PR #3015, opened 2023-01-06
#2532 - Re-added permissions for on-deploy service user

### LINKED ISSUE #2532: Add additional permissions for acs-commons-on-deploy-scripts-service service user
Hi, 

**What we ran into:**
While implementing some on-deploy scripts I ran into the issue that the standard service user ACS Commons uses (`acs-commons-on-deploy-scripts-service`) does not have any write permissions to the `/conf` folder.

**Our use case:**
We created new Context-Aware configurations  and wanted to populate some properties using an on-deploy script, but this was not possible without giving the `acs-commons-on-deploy-scripts-service` service user more rights via a Sling repoinit configuration. I'd rather have this service user have additional permissions.

Could a standard write permission for the `/conf` folder be added to this service user? Most appreciated!

davidjgonzalez:
@Ixbitz and @AntonValadzko i think this should fix #2532 - look right to you?

### DIFF
--- a/content/src/main/content/jcr_root/apps/acs-commons/config/org.apache.sling.jcr.repoinit.RepositoryInitializer-root-acls.config
+++ b/content/src/main/content/jcr_root/apps/acs-commons/config/org.apache.sling.jcr.repoinit.RepositoryInitializer-root-acls.config
@@ -12,7 +12,7 @@ end
 
 create service user acs-commons-on-deploy-scripts-service with path system/acs-commons
 set ACL for acs-commons-on-deploy-scripts-service
-    allow jcr:read on /
+    allow jcr:versionManagement,jcr:read,rep:write,jcr:lockManagement,crx:replicate on /
 end
 "
 ]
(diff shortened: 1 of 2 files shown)


content/src/main/content/jcr_root/apps/acs-commons/config/org.apache.sling.jcr.repoinit.RepositoryInitializer-root-acls.config
```diff
@@ -12,7 +12,7 @@ end
 
 create service user acs-commons-on-deploy-scripts-service with path system/acs-commons
 set ACL for acs-commons-on-deploy-scripts-service
-    allow jcr:read on /
+    allow jcr:versionManagement,jcr:read,rep:write,jcr:lockManagement,crx:replicate on /
```
kwin, line 15: This is a lot, if there is some vulnerability with that service user one can do a lot of damage. I would rather restrict it more by default. Those permissions can be extended on a per use case basis if really necessary

PR #3019, opened 2023-01-11
Feature/crawler delay

Polleke007:


royteeuwen:
@justinedelson  @davidjgonzalez  why does this ScrMetadataIT test exist? We added an extra OSGi config property, this should not be the reason for a build to fail I'd think? As long as you set defaults?

kwin:
@royteeuwen This test is for ensuring backwards-compatibility with regards to configuration properties (compare with https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/37e385cb1ffadb6996af981ebbb6f1daa609438b/bundle/src/test/java/com/adobe/acs/commons/it/build/ScrMetadataIT.java#L67). In the build it failed with a Gateway timeout (error while downloading the old ACS AEM Commons bundle from Central Repo). Just try to trigger the build again.

PR #3072, opened 2023-03-20
Removed RDE logger config

davidjgonzalez:


all/src/main/content/META-INF/vault/filter.xml
```diff
@@ -19,7 +19,6 @@
 
 <workspaceFilter version="1.0">
     <filter root="/apps/acs-commons/config"/>
-    <filter root="/apps/acs-commons/config.rde"/>
```
kwin, line 22: I would rather leave that entry with `type="cleanup"` to make sure this node is removed again on 6.5.x/AEM Cloud SDK servers having previously installed 6.0.4.

kwin (approved):

Issue #3096, opened 2023-04-25
Documentation Update: NamedTransformImage Servlet


joerghoh:
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: ALL
* [x] ACS AEM Commons Version: ALL
* [x] Reproducible on Latest? yes

The documentation of the NamedImageTransformer should be extended. It should cover the fact, that the creation of renditions on-the-fly can be problematic, especially when the underlying image is large. In that case it can cause memory/garbage collection issues and even Out-Of-Memory situations, leading to overall slowdown of the instance up to a "crash" of the JVM.

Instead all required renditions should be pre-created using Processing profiles ([AEM CS](https://experienceleague.adobe.com/docs/experience-manager-cloud-service/content/assets/manage/asset-microservices-configure-and-use.html?lang=en#custom-config); [AEM 6.x](https://experienceleague.adobe.com/docs/experience-manager-65/assets/administer/processing-profiles.html?lang=en
)); in this case the rendition will only be created once, and for that case more optimized solutions are available (for example the use of imagemagick on AEM 6.x, and assetCompute on AEM CS), which are much more suited for that case.

davidjgonzalez:
@joerghoh This can be marked as not-compatible with AEM CS, especially now that we have web-optimized renditions (which is the proper way - better than using static renditions).

IIUC the on the fly renditioning is supported by AEM 6.5 since this is how the AEM Adaptive Servlet works.

joerghoh:
Yeah, I think we should add the same warning also to the Core Components in case Renditions (of whatever size) are resized on the fly.

davidjgonzalez:
@joerghoh so you want to write up the docs PR?

joerghoh:
https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io/pull/248

joerghoh:
I added a warning whenever this servlet is used. And even if this warning is turned off, we will see that it is turned off :-)

https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/3097

PR #3097, opened 2023-04-26
#3096 warn when the NamedTransformImageServlet is used

joerghoh:
This servlet can be quite resource demanding, and in most cases you should create the required renditions upfront using processing profiles. See https://adobe-consulting-services.github.io/acs-aem-commons/features/named-image-transform/index.html

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/images/impl/NamedTransformImageServlet.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/images/impl/NamedTransformImageServlet.java
@@ -126,7 +126,7 @@ public class NamedTransformImageServlet extends SlingSafeMethodsServlet implemen
     private static final Logger log = LoggerFactory.getLogger(NamedTransformImageServlet.class);
     
     private static final Logger AVOID_USAGE_LOGGER = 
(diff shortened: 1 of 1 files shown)


bundle/src/main/java/com/adobe/acs/commons/images/impl/NamedTransformImageServlet.java
```diff
@@ -204,6 +208,14 @@ public final boolean accepts(final SlingHttpServletRequest request) {
     @Override
     protected final void doGet(final SlingHttpServletRequest request, final SlingHttpServletResponse response) throws
             ServletException, IOException {
+    	
+    	
+    	// Warn when this servlet is used
+    	AVOID_USAGE_LOGGER.warn("An image is transformed on-the-fly, which can be a very resource intensive operation. "
+    			+ "If done frequently, you should consider to create such a rendition upfront using processing profiles. "
```
davidjgonzalez, line 215:
```suggestion
    			+ "If done frequently, you should consider switching to dynamic AEM web-optimized images or creating such a rendition upfront using processing profiles. "
```

davidjgonzalez (approved):
Added not about using web-optimized renditions which are the most preferred route (and most similar to this approach, reducing the amount of cost to a user switching off this solution)

PR #3109, opened 2023-05-17
Require Java 11 for building

### LINKED ISSUE #3108: Require Java 11 for building
Multiple test libraries require Java 11 these days:

- AEM Mocks since version 5.2.0
- Mockito since version 5.0

We should require Java 11 for building acs-aem-commons. That also allows to get rid of animalsniffer-maven-plugin and leverage `maven.compiler.release` instead of `maven.compiler.source` and `maven.compiler.target`. The minimum Java version should be enforced.

For now we can still build for Java 8 (even JDK 17 can still emit bytecode compatible with Java 8)

kwin:
This closes #3108

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        jdk: [8, 11, 17]
+        jdk: [11, 17, 21]
         include:
           # lengthy build steps should only be performed on linux with Java 11 (CodeCov analysis, deployment)
           - os: ubuntu-latest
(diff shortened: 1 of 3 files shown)


kwin:
In order to still execute UTs/ITs with Java 8, Maven Toolchains should be used together with m-surefire-p/m-failsafe-p.

davidjgonzalez:
@kwin  is this ready for review/merge?

kwin:
Currently it will drop testing (i.e. run ITs) with Java 8 but I don't think this is really required...
Testing with Java 8 requires more tweaking on the GitHub Actions.

kwin:
@davidjgonzalez This should be the last missing piece for also bringing the failing JDK8 back to green (as that was using the implicitly activated `java8` profile which lead to the `cloud` profile no longer being active implicitly).

davidjgonzalez (approved):

Issue #3112, opened 2023-05-18
AngularJS v1.2 and v1.3 have security vulnerabilities


davidjgonzalez:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: _______
* [ ] ACS AEM Commons Version: _____
* [ ] Reproducible on Latest? yes/no

### Expected Behavior

_Please describe briefly the expected behavior, i.e. when I do X, Y should happen._

### Actual Behavior

_Please describe briefly the actual behavior, i.e. when I do X, Z happens._

### Steps to Reproduce

_Starting from a vanilla instance of AEM at the above version, describe in explicit steps how to reproduce. Link to
any supporting assets, e.g. content packages, below_

### Links

_Links to related assets, e.g. content packages containing test components_

davidjgonzalez:
Updated AngularJS to latest (1.8.2)

han-dsa:
hi, any plan to upgrade Angular JS to latest 1.8.3?

davidjgonzalez:
@han-dsa  can you link to where 1.8.2 is available? I dont see it on the AngularJS site anywhere (1.8.2 seems to be latest).

https://angularjs.org/ and https://code.angularjs.org/

han-dsa:
> @han-dsa can you link to where 1.8.3 is available? I dont see it on the AngularJS site anywhere (1.8.2 seems to be latest).
> 
> https://angularjs.org/ and https://code.angularjs.org/

hi David,
https://www.reddit.com/r/angularjs/comments/v6na5f/angularjs_183_download/

PR #3113, opened 2023-05-18
#3112 - Update AngularJS to latest v1.8.2

davidjgonzalez:


ui.apps/src/main/content/jcr_root/apps/acs-commons/clientlibs/vendor/angularjs/v1.8.2/all/.content.xml
```diff
@@ -18,7 +18,7 @@
   -->
 
 <jcr:root xmlns:cq="http://www.day.com/jcr/cq/1.0" xmlns:jcr="http://www.jcp.org/jcr/1.0"
-    jcr:primaryType="cq:ClientLibraryFolder"
+          jcr:primaryType="cq:ClientLibraryFolder"
           allowProxy="{Boolean}true"
           jsProcessor="[default:none,min:gcc]"
-          categories="[acs-commons.angularjs,acs-commons.angularjs.v1.2.all]"/>
+          categories="[acs-commons.angularjs,acs-commons.angularjs.v1.8.2.all]"/>
```
kwin, line 24: I would make the version specific category only end with 1.8. The micro version is usually not worth pinning. That way micro updates can be done without adjusting the category name.

Issue #3128, opened 2023-07-06
Redirect Manager - Import from excel not working due to "Apache Poi" dependencies upgrade


lollegp:
## Required Information

* [x] AEM Version: AEM Cloud SDK 2023.6.12255.20230608T053118Z
* [x] ACS AEM Commons Version: 6.0.10
* [x] Reproducible on Latest? yes

### Expected Behavior

In redirect manager I should be able to import a set of redirect from an excel file

### Actual Behavior

In redirect manager when I try to import a set of redirect from an excel file I'm getting 500 with an exception and can't proceed with the import

### Steps to Reproduce

Tools --> ACS AEM Commons --> Manage Redirects --> Choose one conf (not relevant which one) --> Import --> Upload xlsx file (even if you upload an empty one is not working) --> Import Redirect Map --> FAIL

### Links

<img width="1670" alt="Screenshot 2023-07-06 at 18 57 51" src="https://github.com/Adobe-Consulting-Services/acs-aem-commons/assets/56459343/1f21034e-fbf8-4b2b-a400-62a7156e7444">

lollegp:
After a quick analysis it seems that with the new AEM Cloud SDK the "Apache Poi" bundle has been updated and in the new version all deprecated methods have been removed and should be replaced with some new available methods.

YegorKozlov:
I confirmed the issue with aem-sdk-2023.6.12441.20230623T133322Z and AEM 6.5.17

ACS commons does not control the version of Apache POI - Apache POI is a transitive dependency of Apache Tika which is used internally by Jackrabbit to index binary data. 

It appears the POI version has been upgraded from 4.0.1 to 5.2.3 which is quite a leap.  And it impacted every place where we read data from spreadsheets: Redirect Manager, MCP jobs, reports, etc . 

The minimal version of AEM supported by  ACS Commons is 6.5.10,  which means we need to support both versions: v4.0.1 and  v5.2.3.

davidjgonzalez:
Fixed in 6.0.12

PR #3129, opened 2023-07-06
#3128 - Redirect Manager import from xlsx not working

lollegp:
(#3128)


YegorKozlov:
Thanks for the PR. This class was impacted by the POI version upgrade, but it's not used by redirect manager.

The classes  to fix are com/adobe/acs/commons/redirects/servlets/ExportRedirectMapServlet.java and com.adobe.acs.commons.redirects.servlets.ImportRedirectMapServlet

lollegp:
Thanks, you are right, I add the fix in the right files and also in another one spotted in the GenericReportServlet.

YegorKozlov (approved):
LGTM

Issue #3130, opened 2023-07-07
NPE in remote asset sync on empty multi-value properties


Roman-Skripka:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.5.17
* [ ] ACS AEM Commons Version: 6.0.10
* [ ] Reproducible on Latest? yes

### Expected Behavior

No error. Sync is successful.

### Actual Behavior

NPE in https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/e3cf05e1bd43f164583138976fc84c904d1b83e1/bundle/src/main/java/com/adobe/acs/commons/remoteassets/impl/RemoteAssetsNodeSyncImpl.java#L351 on 0 length String array property in metadata.

### Steps to Reproduce

Add e.g. dc:title multivalue property to metadata node with no elements. Run remote sync.

PR #3133, opened 2023-07-10
Fixed  Error in remote asset sync on empty multi-value properties #3130

davidjgonzalez:
fixed https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/3130

@Roman-Skripka how's this look? It should skip making the property if the json array is empty.


Roman-Skripka:
looks good, thanks @davidjgonzalez . I will verify this at some point. For now I removed these properties from source instance to get it sync'ed.

davidjgonzalez:
This fix is in 6.0.14 (released)

Issue #3148, opened 2023-08-04
Get DispatcherFlushRules working again on AEMaaCS


royteeuwen:
The current implementation depends on the `Preprocessor` interface, which is no longer used by Sling Content Distribution on AEMaaCS.

I have an open PR at Sling's side to get it fixed (in a nice manner) in AEMaaCS to be able to provide custom ImportPostProcessors:
https://issues.apache.org/jira/browse/SLING-11991

Afterwards we can implement an ImportPostProcessor that would trigger the existing DispatcherFlushRules for AEMaaCS

royteeuwen:
See also https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/3127

royteeuwen:
@davidjgonzalez  Konrad approved the PR, could you also have a look and see how we will be doing the releases in the future?

kwin:
Fixed in 6.0.14

PR #3149, opened 2023-08-04
solves #3148: Add cloud bundle + cloud all package containing fixes for dispatcher flush rules

royteeuwen:


royteeuwen:
@davidjgonzalez  @kwin  can you already have an initial look please ?

all/pom.xml
```diff
@@ -46,6 +46,7 @@
                 <artifactId>filevault-package-maven-plugin</artifactId>
                 <configuration>
                     <packageType>container</packageType>
+                    <name>acs-aem-commons-all</name>
```
kwin, line 49:
```suggestion
                    <name>acs-aem-commons-all-classic</name>
```
I would make it more explicit that this is not suitable for cloud.

all/pom.xml
```diff
@@ -213,5 +214,108 @@
...
+            <id>cloud</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.jackrabbit</groupId>
+                        <artifactId>filevault-package-maven-plugin</artifactId>
+                        <inherited>false</inherited>
```
kwin, line 228: why inherited false?

all/pom.xml
```diff
@@ -213,5 +214,108 @@
...
+                        <executions>
+                            <execution>
+                                <id>generate-metadata-cloud</id>
+                                <goals>
+                                    <goal>generate-metadata</goal>
+                                </goals>
+                                <configuration>
+                                    <classifier>cloud</classifier>
+                                    <name>acs-aem-commons-all-cloud</name>
+                                    <packageType>container</packageType>
```
kwin, line 238: This is redundant, I would only set configuration which differs from the default set outside the profile.

bundle-cloud/pom.xml
```diff
@@ -0,0 +1,141 @@
...
+Bundle-Description: ACS AEM Commons Core OSGi bundle fragment containing AEM server-side logic requiring AEMaaCS
+Fragment-Host: com.adobe.acs.acs-aem-commons-bundle
+                                ]]></bnd>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>biz.aQute.bnd</groupId>
+                <artifactId>bnd-baseline-maven-plugin</artifactId>
+            </plugin>
+            <plugin>
```
kwin, line 58: let us extract all common configuration for bundles into the parent pom.xml into the pluginMgmt.

bundle-cloud/pom.xml
```diff
@@ -0,0 +1,141 @@
...
+                <artifactId>sling-maven-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>biz.aQute.bnd</groupId>
+                <artifactId>bnd-maven-plugin</artifactId>
+                <configuration>
+                    <bnd><![CDATA[
+Bundle-Name: ACS AEM Commons - Bundle Fragment (for AEMaaCS)
+Bundle-Description: ACS AEM Commons Core OSGi bundle fragment containing AEM server-side logic requiring AEMaaCS
+Fragment-Host: com.adobe.acs.acs-aem-commons-bundle
```
kwin, line 50: The host bundle should already be adjusted so that declarative services from this fragment are started as well (reference: https://github.com/adobe/asset-share-commons/blob/9bceb024dc0121c5ed64047e128437174dd3e72a/core/pom.xml#L116).

pom.xml
```diff
@@ -83,6 +83,11 @@
         <sonar.issue.ignore.multicriteria.e1.resourceKey>oakpal-checks/pom.xml
         </sonar.issue.ignore.multicriteria.e1.resourceKey>
 
+        <!-- the minimum AEM 6.5 version being supported -->
+        <aem.classic.api.version>6.5.10.0002</aem.classic.api.version>
+        <!-- the minimum AEMaaCS version being supported -->
+        <aem.sdk.api.version>2023.6.12255.20230608T053118Z-230400</aem.sdk.api.version>
```
kwin, line 89: not every customer is that up2date unfortunately. Do we really require such a new cloud version? In any case we should highlight this in the https://adobe-consulting-services.github.io/acs-aem-commons/pages/compatibility.html.

all/pom.xml
```diff
@@ -213,5 +214,108 @@
...
+                                            <isAllVersionsFilter>true</isAllVersionsFilter>
+                                        </subPackage>
+                                        <subPackage>
+                                            <groupId>${project.groupId}</groupId>
+                                            <artifactId>acs-aem-commons-ui.content</artifactId>
+                                            <filter>true</filter>
+                                            <isAllVersionsFilter>true</isAllVersionsFilter>
+                                        </subPackage>
+                                    </subPackages>
+                                </configuration>
```
kwin, line 270: please add package dependencies to require the cloud version and similar the SP level for AEM classic.

kwin (changes requested):

all/pom.xml
royteeuwen: I would suspect that this would be a breaking change, on-premise customers who would use this new version would have two acs-aem-commons in their CRX Package Manager list, both on status installed. Not sure if this is wanted?

all/pom.xml
royteeuwen: do you have an example of which dependency to use for the cloud version? 

I see the core components content package uses day/cq60/product:cq-platform-content

pom.xml
royteeuwen: Updated to the first one of this year (2023.1.10675.20230113T110236Z-220900)

all/pom.xml
royteeuwen: Copied from assets share commons, removed

all/pom.xml
kwin: For Cloud it doesn't seem applicable as the cp2fm tool doesn't evaluate it. Just for enforcing the SP10 with Classic it would be nice (and to prevent the classic version from being installed in Cloud).

royteeuwen: @kwin  I have set a package dependency for both (can't only set it for classic, else it will be taken in the cloud profile as well, unless I make a profile for classic as well and not build anything by default)

The strange thing is, I have to disable the jackrabbit-packagetype validation, it seems that a container is not supposed to have dependencies? Not sure what the reasoning is there

all/pom.xml
kwin: Right, let's leave it at the default.

all/pom.xml
kwin: I would use `day/cq60/product:cq-content:6.6` to enforce AEM Cloud, the reason why this is only allowed on non-container packages is that container packages are dissolved for AEM Cloud (with https://github.com/apache/sling-org-apache-sling-feature-cpconverter). Therefore please create dedicated apps packages with specific package dependencies.

kwin: @davidjgonzalez Do you know if there is any means to depend on a specific AEM Cloud version (as minimum requirement) via package dependencies? This would be crucial as sometimes the import-packages versions from bundles are not restrictive enough (e.g. when certain referenced resource types have only been introduced in a certain AEM Cloud version).

royteeuwen: @kwin  seeing as the package itself will be named acs-aem-commons-cloud, I'd expect this to be restrictive enough. We will also explicitly mention this in the readme. 

If we do want to go for the dedicated ui.apps.cloud route, what would this entail? What would be in that ui.apps package? 

- use the maven resources plugin to copy everything over from the default ui.apps to an ui.apps.cloud? 
- leave it empty for now and just put the package dependency? This would mean that if you install the acs-aem-commons-cloud on on-prem installation, it would still actually install the "all" package and the bundle, just not the ui.apps.cloud + the bundle.cloud fragment.

kwin: Using `day/cq60/product:cq-content:6.6` would resolve only in AEMaaCS, but on every version of it. But if this the best we can enforce, we should go for it.

davidjgonzalez:
@royteeuwen @kwin is this good to go?

kwin:
This still requires moving the package dependencies for either AEM Classic or Cloud to the app package and building the application package with classifiers. I can have a look next week if @royteeuwen won't beat me to it.

royteeuwen:
@kwin  sorry I was on holiday, I will pick this up tomorrow!

royteeuwen:
@kwin @davidjgonzalez  you can have another look.

I also changed the setup of the dispatcher flush rules by using the recommended approach mentioned in the cloud docs, namely using the INVALIDATE distribution type from Author. This will require an update for the Dispatcher Flush Rules docs

bundle/src/main/java/com/adobe/acs/commons/replication/dispatcher/impl/DispatcherFlusherImpl.java
```diff
@@ -44,12 +45,8 @@
...
  * Service used to issue flush requests to enabled Dispatcher Flush Agents.
  */
-@Component(
-        property = {
-                "service.ranking=-10000"
-        },
-        service = {DispatcherFlusher.class}
-)
+@Component
+@ServiceRanking(-10000)
```
royteeuwen, line 49: @kwin @davidjgonzalez 
I fixed an error where the service ranking was wrongly saved as a String instead of an Integer, making the actual value for OSGi zero

This does make the following test fail though:
com.adobe.acs.commons.it.build.ScrMetadataIT.test
Property service.ranking on component com.adobe.acs.commons.replication.dispatcher.impl.DispatcherFlusherImpl has different types (was: {String}, is: {Integer})

WDYT? can we add an ignore in it

kwin: please add it to ignored properties temporarily

royteeuwen: done

all/pom.xml
```diff
@@ -213,5 +225,121 @@
...
+                                            <artifactId>acs-aem-commons-ui.content</artifactId>
+                                            <target>/apps/acs-commons/install</target>
+                                            <filter>true</filter>
+                                            <isAllVersionsFilter>true</isAllVersionsFilter>
+                                        </embedded>
+                                    </embeddeds>
+                                </configuration>
+                            </execution>
+                            <execution>
+                                <id>on-prem-package</id>
```
kwin, line 285: This doesn't look right to me. The default execution id is `default-package` or `default-metadata` (https://maven.apache.org/guides/mini/guide-default-execution-ids.html#default-executionids-for-implied-executions). You can either disable those or just reconfigure those ids.

ui.apps/pom.xml
```diff
@@ -47,15 +48,42 @@
...
-                            <group>day/cq60/product</group>
-                            <name>cq-content</name>
-                            <version>[6.5.6,)</version><!-- version shipped with AEM 6.5.0 GA -->
-                        </dependency>
-                        <!-- cannot depend on SP 10 as never satisfied on AEMCloud -->
-                    </dependencies>
                 </configuration>
+                <executions>
+                    <execution>
+                        <id>default-package</id>
```
kwin, line 54: the id for metadata is `default-metadata`. Please separate the configuration per goal/execution id.

kwin (changes requested):
Just two minor remarks, other than that looks good to me. Thanks.

all/pom.xml
```diff
@@ -238,7 +238,7 @@
                         <artifactId>filevault-package-maven-plugin</artifactId>
                         <executions>
                             <execution>
-                                <id>generate-metadata-cloud</id>
+                                <id>default-metadata</id>
```
kwin, line 241: you need to have two executions for goal `metadata`: one for classic called `default-metadata` (without classifier) and one for Cloud (with classifier) with a different execution id.

kwin: Maybe the logic is not clear to me... Using profile "cloud" will only build the cloud package? So you either build one or the other but never both in one reactor build? I would prefer a solution where I can build both in one run.

kwin: I would like an approach like in https://github.com/adobe/aem-core-wcm-components/blob/main/all/pom.xml, where profile cloud just builds both artifacts.

royteeuwen: cool, I didn't know about the exclude selector (~cloud) and that was the reason I didn't get it working to have both compiled at the same time. PR is updated to use this.

I saw that the validate-package-cloud doesn't pick up the correct package to validate if only the classifier is mentioned as configuration, I also had to explicitly mention the packageFile. Not sure if this is a bug?

royteeuwen:
@kwin any time for another review ;)?

kwin (approved):
LGTM, thanks.

royteeuwen:
@davidjgonzalez  any insights in when you would be able to provide a new build where you release both the "classic" and "cloud" build?

Best to also update some docs / readme's somewhere?

kwin:
@royteeuwen Can you prepare a documentation PR for https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io/blob/master/acs-aem-commons/pages/maven.md?

kwin:
This looks like a bug in FileVault as it checks all attached artifacts with a given classifier (irrespective of its type/extension). For now just skip validation during the release with `-Dvault.skipValidation=true` (https://jackrabbit.apache.org/filevault-package-maven-plugin/validate-package-mojo.html#skipValidation

davidjgonzalez:
Thanks @kwin  - that works. 
I just realized that the new cloud fragment doesnt have any public classes/interfaces - so no javadocs are generated for it. But OSS nexus projects require (AFIAK) javadoc for all jars deployed to it. Any idea how i can force javadoc to somehow generate an empty javadoc jar for this? Or do should I  add an empty public interface to the project to trigger javadoc generation?

davidjgonzalez:
@kwin  - just FYI, i created a empty public interface in the cloud bundle and that worked. Not sure if there's a more elegant way.

@royteeuwen - this is out at 6.3.0 -- can you test it out and make sure everything got up to maven central? If you can let me know, then we can update the version on the docs site. Thanks!

kwin:
@davidjgonzalez The approach of generating an empty javadoc from https://stackoverflow.com/a/53707024 doesn’t require to add useless interfaces/classes.

royteeuwen:
> @royteeuwen Can you prepare a documentation PR for https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io/blob/master/acs-aem-commons/pages/maven.md?

Ah that's where the documentation is hosted :D good to know, I will provide a PR!

royteeuwen:
@davidjgonzalez  I had the opportunity to test the 6.3.0 release today on an AEMaaCS instance. The Dispatcher Flush Rules work correctly again, using the cloud package :)! I also updated the docs PR

PR #3151, opened 2023-08-05
Feature: Content Sync

YegorKozlov:
## Purpose
Incrementally synchronize content between AEM Author instances.

ACS Content Sync is yet another solution aiming to overcome the shortcomings of using CRX Package Manager / VLT RCP to copy content between AEM environments.

### How does Content Sync Work?
Content Sync uses JSON Export/Import between Sling GET/POST servlets. The tool fetches a catalog of resources from the source instance, e.g.
```
/content/path1  lastModified1
/content/path2  lastModified2
/content/path3  lastModified3
```
then it compares it with the local state, computes the delta and synchronizes it via JSON Export/Import.

### How many HTTP calls does Content Sync make?
The ballpark estimation is O(N) where N is the number of content resources (cq:Page, dam:Asset, etc.) being sync-ed. 
For example, it takes an HTTP call to sync a _cq:Page_ resource, so to sync a tree of N _cq:Page_ nodes the tool will make minimum N HTTP calls. To sync a _dam:Asset_ node the tool will make an HTTP call to grab the jcr:content and sub-nodes, and an HTTP call to grab binary data for each rendition.

### What can I sync?
In general - any content resources, like

- Pages
- Assets
- Experience Fragments
- Content Fragments
- Tags

## How to Use
In order to use the Content Sync tool, you need the latest version of ACS Commons installed on the source and target AEM instances.

### Configure Sling Get Servlet
Create an OSGi configuration for PID _org.apache.sling.servlets.get.DefaultGetServlet_ and make sure that:

- JSON Renderer is enabled
- JSON Max Results parameter is set to at least 1000
- Legacy ECMA date format is unchecked

image

### Configure the host(s) to synchronize from
Navigate to the Content Sync tool (http://localhost:4502/apps/acs-commons/content/contentsync.html) and click the Configure button in top right corner. Then click _Add Host_

image
image
The _contentsync-user_ user needs read permissions on the content you are going to sync.

### Run the Content Sync tool
Navigate to the Content Sync tool http://localhost:4502/apps/acs-commons/content/contentsync.html and select a path to synchronize. 
The default mode is Incremental, which means only new or modified resources will be downloaded. If the Incremental option is unchecked then the tool will do full sync and download the entire content tree.

The 'On-complete' workflow option allows you to define a custom post-update action. For example, you can define a workflow to clear properties like _cq:replicationStatus_, or automatically publish a resource if it is published in the source environment.

image

## Pluggable Update Strategy
The strategy to select incremental delta is pluggable. 
The default algorithm is based on the _cq:lastModified/jcr:lastModified_ timestamp. 
As an example, Content Sync ships with another strategy which synchronizes _dam:Asset_ nodes based on the _dam:sha1_ checksum in the asset metadata. Users can plug-in their custom strategies by implementing `com.adobe.acs.commons.contentsync.UpdateStrategy` interface;


davidjgonzalez:
@YegorKozlov is this read for review?

YegorKozlov:
> @YegorKozlov is this read for review?

yes @davidjgonzalez . it's ready for review

AntonValadzko:
Hi @YegorKozlov, Is it possible to sync only root page without child pages below it? Like only this path /content/site/en/home/jcr:content

YegorKozlov:
> Hi @YegorKozlov, Is it possible to sync only root page without child pages below it? Like only this path /content/site/en/home/jcr:content

In the current implementation it's not possible. I'm thinking of adding a new Query strategy, so that instead of a path user would pass a jcr query. This way you will have a finer control what to sync.

PR #3179, opened 2023-09-06
Issue 3177 - Automate creation of vanity URLs for each country & lang…

yelouardi:
…uage

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/redirectmaps/impl/AddEntryServlet.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/redirectmaps/impl/AddEntryServlet.java
@@ -18,15 +18,19 @@
 package com.adobe.acs.commons.redirectmaps.impl;
 
 import java.io.IOException;
-import java.util.List;
+import java.util.*;
 
 import javax.servlet.ServletException;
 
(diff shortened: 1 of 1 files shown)


yelouardi:
Hi @davidjgonzalez  can you check please my dev ?
Thank you

yelouardi:
Hi @rombert  @davidjgonzalez 
Can you please review my PR ?
Thank you

davidjgonzalez:
@yelouardi - can you describe what this does exactly, and what usecase your tackling?

It seems like it 

Uses a generic list whose items are in the format:

   ```
    + item-1
        + jcr:title: lang code (ex. en)
        + value: country code (ex. us)

Turns this into a map (so i guess this list can have only 1 item for each country?)

Then assuming there are countries, adds the redirect entries for each country/lang?

Or does this ONLY support countries and not langs? 

Its hard to tell since the tests dont have representative real-life params.

yelouardi:
Hi @davidjgonzalez 
Thank you 

Turns this into a map (so i guess this list can have only 1 item for each country?) :  **YES**

Then assuming there are countries, adds the redirect entries for each country/lang? **Each pays has one country language** by default  

Or does this ONLY support countries and not langs? **YES**

PR #3199, opened 2023-10-25
Update maven.yml

davidjgonzalez:


kwin:
@davidjgonzalez Does this fix #3201 ?

.github/workflows/maven.yml
```diff
@@ -89,7 +89,7 @@ jobs:
         run: mvn -e -B -V ${{ env.MVN_GOAL }} ${{ env.MVN_ADDITIONAL_OPTS }}
 
       - name: ${{ matrix.namePrefix }} Build cloud all package with Maven ${{ env.STEP_NAME_SUFFIX }}
```
kwin, line 91: why do we need two Maven runs at all? That is IMHO not working as the module `all` relies on dependencies only available in a reactor build (or if previously installed in the local repo).

royteeuwen: no longer required, leftover from when it was still required to run twice but after review remarks from you we were able to have one build using the `cloud` profile

royteeuwen: @davidjgonzalez  could you remove this entire run (line 91 and 92) and add -P cloud to the previous run (line 89). This will release both artifacts

kwin:
Superseded by #3208

PR #3211, opened 2023-11-15
Always enable cloud profile explicitly if other profiles are enabled via

kwin:
CLI


kwin:
@davidjgonzalez I forgot to enable the cloud profile for the main build. This should finally fix all builds in master.

PR #3224, opened 2023-11-30
Do no longer register DialogProviderAnnotationProcessor implicitly

### LINKED ISSUE #3223: Project with class extending WCMUsePojo leads to build error: cannot access aQute.bnd.annotation.ConsumerType
## Required Information

* [X] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.5.17.0
* [ ] ACS AEM Commons Version: 5.7.0, 6.3.2
* [X] Reproducible on Latest? yes (reproducible on every version since 4.7.2. Last working version: 4.7.0.

### Expected Behavior

I uploaded a simple maven testproject to

https://github.com/HENgames/acscommonsprob

The only class extends com.adobe.cq.sightly.WCMUsePojo, the only dependencies are uber-jar 6.5.17 and acs-aem-commons-bundle 6.3.2. The maven build should finish without any problem.

### Actual Behavior

I get the following errors on mvn clean install:

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project acscommonsprob-reproducer: Compilation failure
[ERROR] cannot access aQute.bnd.annotation.ConsumerType
[ERROR]   class file for aQute.bnd.annotation.ConsumerType not found

### Steps to Reproduce

Run

mvn clean install

on the mencioned project. You should get the messages stated above. When downgrading acs-aem-commons-bundle to 4.7.0 or lower, the maven build finishes without any error.

My current configuration is:

mvn -v
Apache Maven 3.8.6 (84538c9988a25aec085021c365c560670ad80f63)
Maven home: [..]
Java version: 1.8.0_392, vendor: Azul Systems, Inc., runtime: [..]
Default locale: de_DE, platform encoding: UTF-8
OS name: "mac os x", version: "13.6.2", arch: "aarch64", family: "mac"

We have several different build systems, all with the same problem.

### Links

I have seen this problem mencioned in other places, like

https://sourcedcode.com/blog/aem/aem-project-wcmusepojo-class-error-of-class-file-for-aqute-bnd-annotation-consumertype-not-found

https://github.com/projectlombok/lombok/issues/2145

The common advice is to include dependency

<dependency>
    <groupId>biz.aQute.bnd</groupId>
    <artifactId>bndlib</artifactId>
    <version>2.4.0</version>
    <scope>provided</scope>
</dependency>

This is no solution in our case, as the dependency is from 2014 and contains older versions of the official OSGi annotations (<= 5.0 it seems). I have found out, that including the following dependency solves the problem:

 <dependency>
     <groupId>biz.aQute.bnd</groupId>
     <artifactId>biz.aQute.bnd.annotation</artifactId>
      <version>6.4.1</version>
 </dependency>

I have seen the newer version 7.0.0, but for now we are bound to 6.4.1 as the newest java 8 compatible version.

Can you recommend this solution / workaround?

kwin:
Clarify in javadoc how to use the annotation processor explicitly Using the DialogProviderAnnotationProcessor requires that all annotation classes are available on the classpath.

This closes #3223


kwin:
Compare also with https://bugs.openjdk.org/browse/JDK-8306819

henrykuijpers:
+1 for this change - Would be good to instruct consumers of ACS AEM Commons to enable this so that the annotation processor is not going to be used when it is not needed. :)

kwin:
> Would be good to instruct consumers of ACS AEM Commons to enable this so that the annotation processor is not going to be used when it is not needed. :)

Definitely, unfortunately the whole feature is completely undocumented right now, I created https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io/issues/256 to track that.

kwin:
@davidjgonzalez Any concerns with merging this since it causes issues for customers (see #3223). Unfortunately I haven't heard from @badvision about this.

henrykuijpers:
Thanks for merging this! Saves us quite the hassle!

PR #3227, opened 2023-12-04
Use JcrPackageDefinition for retrieving metadata

### LINKED ISSUE #3225: PackageGarbageCollector leaves temp files behind
## Required Information

* [X] AEM Version, including Service Packs, Cumulative Fix Packs, etc: AEM CS, release 14227
* [X] ACS AEM Commons Version: 6.2.0
* [X] Reproducible on Latest? yes/no: yes

### Expected Behavior

The PackageGarbageCollector must not leave temporary files behind.

### Actual Behavior

When the PackageCollectorJob runs across a large amount of packages, which are not deleted, all these packages remain in the java.io.tmpdir; depending on the size of the allocated scratch space and the package size of these non-removed packages, this can lead to stability problems.

### Steps to Reproduce

* have a spare test instance which you can throw away after this test (local SDK is ok)
* configure the logger com.adobe.acs.commons.packagegarbagecollector to DEBUG
* configure the PackageGarbageCollector in a way, that it does not remove any package
 ** configure the "packageName" property to ``day/cq60/product``
 ** set the cron expression to ``0 17 12 ? * * *`` to execute it at 12:17 PM (local time), so it's executed immediately and you don't need to wait
* wait until the job has been executed 
* now check  the files in your "java.io.tmpdir" directory, and you will find (at least some of) these packages left there.

(you can find the actual path for ``java.io.tmpdir`` in the OSGI webconsole: "Main" -> "JMX" -> "Runtime")

(the root cause seems to be that the jcrPackage objects are not closed.)

kwin:
This is less error prone and faster than acting on the compressed underlying ZIP binary
Always close JcrPackageManager (although it should now no longer be necessary)
Some refactoring

This closes #3225


bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
```diff
@@ -118,120 +126,78 @@ public JobResult process(Job job) {
...
 
-        public boolean isSameNameAndGroup(JcrPackage otherPackage) {
-            Optional<PackageId> otherPackageId = getPid(otherPackage);
-            Optional<PackageId> thisPackageId = getPid(jcrPackage);
-            if (otherPackageId.isPresent() && thisPackageId.isPresent()) {
-                return otherPackageId.get().getGroup().equals(thisPackageId.get().getGroup())
-                        && otherPackageId.get().getName().equals(thisPackageId.get().getName());
+    private boolean isLatestInstalled(JcrPackageDefinition referencePkgDefinition, Stream<JcrPackage> installedPackages) throws RepositoryException {
+        try {
+            Optional<@Nullable JcrPackageDefinition> lastInstalledPckDefinitionOptional = installedPackages
```
joerghoh, line 154: that type definition seems a bit odd to me. Isn't ```Optional<>``` just covering the nullable case?

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
```diff
@@ -118,120 +126,78 @@ public JobResult process(Job job) {
...
+               .map(p -> {
+                    try {
+                        return p.getDefinition();
+                    } catch (RepositoryException e) {
+                        throw new UncheckedRepositoryException(e);
```
joerghoh, line 159: Does this exception contain enough metadata to identify the package where the exception happened? Or should we add an additional message to it?

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
kwin: Right, that was a leftover from the refactoring. Fixed now.

bundle/src/main/java/com/adobe/acs/commons/packagegarbagecollector/PackageGarbageCollectionJob.java
kwin: I now enrich the relevant exception with the package path

joerghoh (approved):
lgtm

davidjgonzalez (approved):

Issue #3264, opened 2024-02-22
NullPointerException when trying to open dialog generated via @DialogProvider


csaboka:
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: AEM Cloud 2023.12.14697.20231215T125030Z-231200
* [x] ACS AEM Commons Version: 6.3.8
* [x] Reproducible on Latest? yes

### Expected Behavior

Composite widgets should work on dialogs generated by the @DialogProvider annotation.

### Actual Behavior

A NullPointerException is thrown when you try to open a dialog built using @DialogProvider and @FormField, but only if a field of type MultifieldComponent is used and a complex type is used as the component type.

### Steps to Reproduce

1. Start from a new project created from the AEM archetype using [these instructions](https://github.com/adobe/aem-project-archetype) . Use aemVersion=cloud
2. [Add ACS AEM Commons to the project](https://adobe-consulting-services.github.io/acs-aem-commons/pages/maven.html)
3. Delete ui.apps/src/main/content/jcr_root/apps/mysite/components/helloworld/_cq_dialog . We will replace this dialog via annotations.
4. Add the necessary annotation processor to core/pom.xml:
```
            <plugin>
                  <artifactId>maven-compiler-plugin</artifactId>
                  <configuration>
                    <annotationProcessors>
                      <annotationProcessor>com.adobe.acs.commons.mcp.form.DialogProviderAnnotationProcessor</annotationProcessor>
                    </annotationProcessors>
                  </configuration>
            </plugin>
```
5. Delete the HelloWorldModelTest class so it won't interfere with the changes.
6. Replace the HelloWorldModel class with this code:
```
package com.mysite.core.models;

import java.util.List;

import com.adobe.acs.commons.mcp.form.DialogProvider;
import com.adobe.acs.commons.mcp.form.FormField;
import com.adobe.acs.commons.mcp.form.MultifieldComponent;
import org.apache.sling.api.resource.Resource;
import org.apache.sling.models.annotations.Model;

@DialogProvider(title = "Example Component")
@Model(adaptables = Resource.class, resourceType = "mysite/components/helloworld")
public class HelloWorldModel {

    public static class Nested {
        @FormField(name = "Dummy Text")
        private String dummy;
    }

    @FormField(name = "Nested data", component = MultifieldComponent.class)
    private List<Nested> nesteds;

    public String getMessage() {
        return "Hello, world!";
    }

}
```
[...]

Issue #2500, opened 2020-12-22
CodeClimate only accepts headers with Adobe as copyright owner


kwin:
According to https://www.apache.org/licenses/LICENSE-2.0#apply the first line consists out of

> Copyright [yyyy] [name of copyright owner]

The CodeClimate rule only accepts lines matching the following regex ` ?* Copyright (C) \d\d\d\d(( -)|( - \d\d\d\d))? Adobe`
That excludes all files having a different copyright owner than "Adobe".

That rule should be relaxed, as not for all source code files Adobe is necessarily the copyright owner (compare with https://softwareengineering.stackexchange.com/questions/150798/on-what-name-should-i-claim-copyright-in-open-source-software).

One such false positive is reported at https://codeclimate.com/github/Adobe-Consulting-Services/acs-aem-commons/pull/2499#issues.

badvision:
I understand your concern.  One counterpoint is disputed ownership can wreak havoc for the project as a whole.  Case in point: Consider what happened to the Bukkit project a few years ago when one of the authors issued a takedown for all the code that they contributed.  It basically killed the project.  https://blog.jwf.io/2020/04/open-source-minecraft-bukkit-gpl/

Granted that was more to do with the GPL license than Apache but it is still a concern.  Happy to hear others' perspectives on the topic, I don't really have a preference here unless it's the decision that helps ensure the long-term health of this community project.

justinedelson:
The COPYRIGHT file at the root of the repository clearly states that "Adobe holds the copyright for all the files found in this repository."

kwin:
The COPYRIGHT file is IMHO more targeted at consumers, not at contributors. If you have to give Adobe the full copyright, this should be clearly mentioned in https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/CONTRIBUTING.md. In contrast to e.g. https://github.com/adobe/aem-core-wcm-components/blob/master/CONTRIBUTING.md signing the Adobe CLA (which implies giving Adobe the copyright) is not necessary for this project... The main concern with giving Adobe the full copyright is that this would even allow Adobe to do anything with that source code (even selling it under a separate license which is incompatible with the ASF license)

kwin:
[...]

PR #3265, opened 2024-02-22
Defect/mcp forms/npe on multifield

csaboka:
#3264 Make sure no null values are added to ValueMaps.

I had to add some new dependencies because JUnit 4 tests are not currently picked up by JUnit 5, but the JUnit 5 versions of some testing utilities were missing. You probably want to include the "junit-vintage" artifact to have the JUnit 5 engine run older tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 
 ## Unreleased ([details][unreleased changes details])
 
+### Fix
+
+- #3264 - NullPointerException while displaying MCP forms
+
 ## 6.3.8 - 2024-02-02
 
 ### Fix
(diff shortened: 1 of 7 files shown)


csaboka:
Well, codeclimate is complaining about the license header, even though I used `mvn license:update-file-header` as described in CONTRIBUTING.md. Please let me know if I should change the copyright holder to Adobe, although I'm pretty sure the current value is the correct one.

kwin:
Look at https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2500.

csaboka:
Thank you. I'll keep things simple and change the copyright line to say Adobe. I agree it would be better if there was some clear indication about it, though.

davidjgonzalez (approved):

PR #3266, opened 2024-02-23
Allow to configure Component/BundleDisabler via Configuration Factories

### LINKED ISSUE #3262: Allow usage of BundleDisabler and ComponentDisabler via factory configuration
Currently both 

- https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/ae96011c9104b4c4ffb3958ac3031688881f22c6/bundle/src/main/java/com/adobe/acs/commons/util/impl/BundleDisabler.java#L51 and
- https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/ae96011c9104b4c4ffb3958ac3031688881f22c6/bundle/src/main/java/com/adobe/acs/commons/util/impl/ComponentDisabler.java

are singletons with each having only a single configuration component.
One should allow to have several instances via the `Configuration Factory` pattern (https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#service.component-deployment) to allow independent software parts to disable certain components independent of each other.

kwin:
This allows them to be no longer singletons.
Migrate to OSGi DS/Metatype annotations.
Cleanup of Javadoc.

This closes #3262

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 ### Changed
 
 - #3267 - Remove JSR305 dependency
+- #3262 - Allow to configure Component/BundleDisabler via Configuration Factories
 
 ### Fixed
 
--- a/bundle/pom.xml
+++ b/bundle/pom.xml
@@ -350,6 +350,11 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
+            <groupId>org.osgi</groupId>
+            <artifactId>org.osgi.service.event</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
             <groupId>org.slf4j</groupId>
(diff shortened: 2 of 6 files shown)


kwin:
I am still in clarification if a component should be usable both through factory configurations and regular configurations (for backwards-compatibility reasons): https://github.com/bndtools/bnd/issues/6042.

stbischof:
But why do you need a factory.pid and a pid?

Just use a normal pid and make the conponent required. Will work same.
You can use every pid in configadmin to make a factory pid fron it.

kwin:
> You can use every pid in configadmin to make a factory pid fron it.

This is about the metatype being created. If it doesn't have both delegate for factory and non factory PID it won't be exposed correctly in Felix Web Console UI.

stbischof:
But isnt that a Missing Feature of webconsole.
There should also be normal pids be configurable .

PR #3233, opened 2023-12-15
Fix java 11 warning for DialogProviderAnnotationProcessor - Fixes #3187

henrykuijpers:

PR #3273, opened 2024-02-25
Re-enable JUnit 4 and JUnit 3 unit tests

### LINKED ISSUE #3270: Only JUnit5 tests executed by m-surefire-p
The commit https://github.com/Adobe-Consulting-Services/acs-aem-commons/commit/66497685d3df8de35a4b7fc5e575cb4270dda4c6 broke execution of JUnit4 test cases.
That leads to the fact that only 7 from the 1800 are being executed.

We need to explicitly add the `junit-vintage-engine` according to <https://maven.apache.org/surefire/maven-surefire-plugin/examples/junit-platform.html#how-to-run-multiple-apis-or-engines>

csaboka:
Looks like this was broken in #3233. With the addition of JUnit 5, the existing unit tests got ignored (note the sharp drop in overall coverage in the Codecov report). This PR should restore those old tests to full functionality.

Fixes #3270

### DIFF
--- a/bundle/pom.xml
+++ b/bundle/pom.xml
@@ -504,6 +504,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <scope>test</scope>
--- a/bundle/pom.xml
+++ b/bundle/pom.xml
@@ -500,7 +500,12 @@
         </dependency>
         <dependency>
             <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter</artifactId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/bundle/pom.xml
+++ b/bundle/pom.xml
@@ -504,11 +504,6 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.junit.vintage</groupId>
-            <artifactId>junit-vintage-engine</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <scope>test</scope>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -610,6 +610,12 @@ Service-Component: OSGI-INF/*.xml
                 <version>5.10.0</version>
                 <scope>test</scope>
             </dependency>
+            <dependency>
```
kwin, line 613: Let us rather manage all versions together with the junit-bom (like outlined in https://github.com/junit-team/junit5-samples/blob/efac0e2c2f59eff4b147f67145010d890e062f55/junit5-jupiter-starter-maven/pom.xml#L18).

bundle/pom.xml
```diff
@@ -503,6 +503,11 @@
             <artifactId>junit-jupiter</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
kwin, line 506: The engine should rather be declared only as plugin dependency like outlined in https://maven.apache.org/surefire/maven-surefire-plugin/examples/junit-platform.html#select-engine-and-use-multiple-apis.

kwin (changes requested):
Looks good in general, just some minor remarks regarding dependency scopes/management.

pom.xml
csaboka: Makes sense. This should be fixed now.

bundle/pom.xml
csaboka: I have made this change, but since the dependencyManagement section doesn't seem to apply to plugin dependencies, I had to repeat the desired JUnit version twice after all. It feels this is in the opposite direction than your other request.

We could always extract the desired JUnit version to a property, so it can be reused in different parts of the POM even without relying on the dependency management feature.

kwin: Indeed, but I prefer not bloating the test classpath with the engines for no reason. So instead of referencing `junit-jupiter` I would recommend to only depend on its relevant individual transitive dependencies, i.e. `junit-juper-api` and `junit-jupiter-params` (https://github.com/junit-team/junit5/blob/main/junit-jupiter/junit-jupiter.gradle.kts) with scope `test`. Although the jupiter engine is just having scope `runtime` in `junit-jupiter`, this will be transitively converted to scope `test` (https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope).

Then both engines are only extending the surefire classpath. Introducing a common property name for easily updating all junit5 dependencies would be good though.

csaboka: I have pushed the requested changes. Does it look good now?

kwin: Looks great, thanks a lot.

kwin (approved):

csaboka:
I have no idea why the Mac test failed or why the codeclimate check is stuck. Is there a way to redo the checks? Should I push a dummy commit?

kwin:
> I have no idea why the Mac test failed or why the codeclimate check is stuck

The first seems to be a race condition happening from time to time. Only happens on Mac CI, not related to your commit. No clue about the second one. Nothing to worry about though, this doesn't prevent the merge. I will wait till the end of the week before merging to give @davidjgonzalez to have a look as well.

Issue #3295, opened 2024-03-19
ACS Commons Reports not generating on AEMaaCS throws error


ankit-gubrani:
## Required Information

* [X] AEM Version, including Service Packs, Cumulative Fix Packs, etc: AEMaaCS 6.5 (2023.11.14227.20231108T162349Z)
* [X] ACS AEM Commons Version: 6.0.8
* [X] Reproducible on Latest? yes

### Expected Behavior

Sample reports should generate without any error on AEMaaCS

### Actual Behavior

On generating a report on AEMaaCS throws error: "Unable to run report due to error!"

### Steps to Reproduce

Go to ACS AEM Commons from Tools on AEM author welcome screen. Next click on Reports tile and click on any sample or custom report & fill the required fields, on executing or downloading the report it throws the error : "Unable to run report due to error!".

<img width="1714" alt="Screenshot 2024-03-19 at 9 21 09 AM" src="https://github.com/Adobe-Consulting-Services/acs-aem-commons/assets/8350518/13a14992-ca8c-47f2-b964-2fa2bd4d9816">

### Error from log files

```
19.03.2024 16:21:01.574 *ERROR* [76.103.45.134 [1710865261567] GET /var/acs-commons/reports/sample-page-report/_jcr_content.results.html HTTP/1.1] org.apache.sling.engine.impl.SlingRequestProcessorImpl service: Uncaught Throwable
javax.el.ELException: Error reading 'results' on type com.adobe.acs.commons.reports.models.QueryReportExecutor
	at javax.el.BeanELResolver.getValue(BeanELResolver.java:87) [org.apache.sling.scripting.el-api:1.0.4]
	at javax.el.CompositeELResolver.getValue(CompositeELResolver.java:54) [org.apache.sling.scripting.el-api:1.0.4]
	at org.apache.el.parser.AstValue.getValue(AstValue.java:123)
	at org.apache.el.ValueExpressionImpl.getValue(ValueExpressionImpl.java:186)
	at org.apache.sling.scripting.jsp.jasper.runtime.PageContextImpl.proprietaryEvaluate(PageContextImpl.java:975) [org.apache.sling.scripting.jsp:2.6.0]
	at apps.acs__002d__commons.components.utilities.report__002d__builder.report__002d__page.results__002e__jsp._jspx_meth_c_set_1(results__002e__jsp.java:417) [customer-precompiled-scripts:1.0.0.SNAPSHOT]
[...]

PR #3296, opened 2024-03-19
Updated the annotations in QueryReportConfig fixing the query manager issue due to empty query language

ankit-gubrani:
Updated the annotations in QueryReportConfig fixing the query manager issue due to empty query language : Issue #3295

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/reports/models/package-info.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/reports/models/package-info.java
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-@org.osgi.annotation.versioning.Version("1.2.0")
+@org.osgi.annotation.versioning.Version("1.2.1")
 package com.adobe.acs.commons.reports.models;
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 
 - #3270 - Re-enable accidentally disabled JUnit3/4 tests
 - #3200 - Remove useless public interface in Cloud Bundle to get javadocs to be built
+- #3295 - Updated the annotations in QueryReportConfig fixing the query manager issue due to empty query language
 
 ## 6.4.0 - 2024-02-22
 
--- a/bundle/src/main/java/com/adobe/acs/commons/reports/models/DateReportCellCSVExporter.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/reports/models/DateReportCellCSVExporter.java
@@ -20,8 +20,6 @@ package com.adobe.acs.commons.reports.models;
 import java.text.SimpleDateFormat;
 import java.util.Calendar;
 
-import javax.inject.Inject;
-
 import com.adobe.acs.commons.reports.internal.ExporterUtil;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.sling.api.resource.Resource;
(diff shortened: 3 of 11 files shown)


masmohammed:
+1

masmohammed (approved):

bundle/src/main/java/com/adobe/acs/commons/reports/models/package-info.java
```diff
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-@org.osgi.annotation.versioning.Version("1.2.0")
+@org.osgi.annotation.versioning.Version("1.2.1")
```
davidjgonzalez, line 18:
```suggestion
@org.osgi.annotation.versioning.Version("1.2.0")
```

Why does this require a version bump to 1.21? Im not seeing where the public API is changed in this package. Am i missing something?

davidjgonzalez (changes requested):
Looks good other than the package-info bump... just want to confirm that bnd is asking for this -- im not seeing why it would.

bundle/src/main/java/com/adobe/acs/commons/reports/models/package-info.java
kwin: Micro version adjustments are required for almost all changes. You should never modify bytecode in a package without adjusting at least micro version.

kwin:
I don’t understand what this fixes, by just switching to injector specific annotations? Which value was not correctly injected with `@Inject`?

davidjgonzalez:
IIRC Sling Bindings is the only injector that precedes ValueMap .. so assuming this fixes it, `queryLanguage` (https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/3296/files#diff-59c585809b6d7e2373fbf6a355cd03425c103bd3aebd478b5cfa16820238d3bbR36) would be set in sling bindings?

I do like the change - as i always find @Inject unclear wrt to where the value is coming from -- but agreed, it's odd this is the fix.

@ankit-gubrani you did very this resolves the issue, correct?

kwin:
@davidjgonzalez You can check via `/system/console/scriptingvariables` if a scripting variable with name `queryLanguage` does exist for a particular resource. I would be interested which Scripting Variables Provider would add such a variable (and with which value that is added which is not supported by the QueryManager....)

ankit-gubrani:
> IIRC Sling Bindings is the only injector that precedes ValueMap .. so assuming this fixes it, `queryLanguage` (https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/3296/files#diff-59c585809b6d7e2373fbf6a355cd03425c103bd3aebd478b5cfa16820238d3bbR36) would be set in sling bindings?
> 
> I do like the change - as i always find @Inject unclear wrt to where the value is coming from -- but agreed, it's odd this is the fix.
> 
> @ankit-gubrani you did very this resolves the issue, correct?

Yes @davidjgonzalez , I did verify it fixes the issue. I initially updated the annotations in QueryReportConfig.java but then reports were showing data on executing the report but download was empty. So in order to fix that I had to update the annotations all the other places.

ankit-gubrani:
> @davidjgonzalez You can check via `/system/console/scriptingvariables` if a scripting variable with name `queryLanguage` does exist for a particular resource. I would be interested which Scripting Variables Provider would add such a variable (and with which value that is added which is not supported by the QueryManager....). In a standard AEM cloud SDK server no such variable exists.

@kwin I didn't find a script variable named 'queryLanguage'. 
I have noticed with last few releases/versions of AEMaaCS @Inject annotation specially is not able to read resource properties, either we had to add @Named or diectly use @ValueMapValue, although I am not sure if others are also facing the similar issue but we ran into very similar issue few weeks back with a vendor provided translation connector as well.

kwin:
If Inject annotations do no longer work, then this is a bug in AEMaaCS. Please raise a ticket with Adobe support to get that fixed.

ankit-gubrani:
> I don’t understand what this fixes, by just switching to injector specific annotations? Which value was not correctly injected with `@Inject`?

So @Inject is not able to read the properties from JCR nodes in the Sling Models and thats why I initially thought it was just the queryLanguage which was empty but everything was empty so I had to update all the sling models later in my 2nd commit.

Issue #3308, opened 2024-04-01
Redirect Manager should support fulltext search for locating redirect rules in long lists


davidjgonzalez:
It would be nice if Redirect Manager's redirect rule listing UI would support fulltext search so its easier to find and locate redirect rules in very long lists.

PR #3309, opened 2024-04-01
#3308 - Add fulltext search to redirect manager

davidjgonzalez:
This performs a fulltext search over:

+ note
+ source
+ target
+ cq:tags. ... tho this is on tag IDs so isnt that useful
+ cacheControlHeader
+ createdBy
+ lastModifiedBy

2024-04-01 at 10 00 AM

2024-04-01 at 10 00 AM

For this to work the checkbox needs to be toggled on; and you need a supporting Oak index deployed (example below). If this is not checked, then it falls back to orginal search. When search is enabled, you cannot re-arrange rows.

2024-04-03 at 10 23 AM

I also made some cosmetic changes to the size of the table, and change the sling resource include for each row, to use the HTL template w/ RequestAttribute in SlingModel which i think speeds up the page rendering. 

## Oak index required

```
{
    "compatVersion": 2,
    "async": "async",
    "queryPaths": [
        "/conf"
    ],
    "includedPaths": [
        "/conf"
    ],
    "jcr:primaryType": "oak:QueryIndexDefinition",
    "evaluatePathRestrictions": true,
    "type": "lucene",
    "indexRules": {
        "jcr:primaryType": "nt:unstructured",
        "nt:unstructured": {
            "jcr:primaryType": "nt:unstructured",
            "properties": {
                "jcr:primaryType": "nt:unstructured",
                "cacheControlHeader": {
                    "name": "cacheControlHeader",
                    "analyzed": true,
                    "jcr:primaryType": "nt:unstructured"
                },
                "note": {
                    "name": "note",
                    "analyzed": true,
                    "jcr:primaryType": "nt:unstructured"
                },
                "createdBy": {
                    "name": "jcr:createdBy",
                    "analyzed": true,
                    "jcr:primaryType": "nt:unstructured"
                },
               "statusCode": {
                    "name": "statusCode",
                    "analyzed": true,
                    "jcr:primaryType": "nt:unstructured"
                }
                "tags": {
                    "name": "cq:tags",
                    "analyzed": true,
                    "jcr:primaryType": "nt:unstructured"
                },
                "source": {
                    "name": "source",
                    "analyzed": true,
                    "jcr:primaryType": "nt:unstructured"
                },
                "resourceType": {
                    "name": "sling:resourceType",
                    "propertyIndex": true,
                    "jcr:primaryType": "nt:unstructured"
                },
                "target": {
                    "name": "target",
                    "analyzed": true,
                    "jcr:primaryType": "nt:unstructured"
                },
                "statusCode": {
                    "name": "statusCode",
                    "analyzed": true,
                    "jcr:primaryType": "nt:unstructured"
                }
            }
        }
    }
}
```


YegorKozlov:
@davidjgonzalez  Thanks! It looks very cool. 

The original limit of 1000 is an overkill, we can surely drop it down to 100.  

I think the last touch would be to warn users if the Oak index is missing. I think we can catch the exception and show an alert. Can't we?

davidjgonzalez:
@YegorKozlov I just made the page sizes (browse and search) configurable under Options; just incase someone likes the big pages.

Wrt to the missing - no exception is thrown -- i believe a WARN is just logged (atleast if there are not MAX_TRAVERSE_RESULTS are not there) ... I think the only ways to check this is:

1. Run an Explain on the query before it runs, and determine if its traversal form that.
2. Use a service user can check if a well-named index exists. 

I updated the Options field for the checkbox, it display inline text stating an index is required. 

Open to other ideas .. just not sure there are any great ones.

YegorKozlov (approved):

bundle/src/main/java/com/adobe/acs/commons/redirects/models/Redirects.java
```diff
@@ -103,4 +129,76 @@ public String getContextPrefix() {
...
+                    // Get a reference to QB's leaking ResourceResolver
+                    leakingResourceResolver = hit.getResource().getResourceResolver();
+                }
+
+                Resource resource = request.getResourceResolver().getResource(hit.getPath());
+                resources.add(resource);
+            }
+
+        } catch (RepositoryException e) {
+            //log.error("Error collecting search results", e);
```
davidjgonzalez, line 194:
```suggestion

```

PR #3311, opened 2024-04-02
#3112 - upgrade AngularJS to v1.8.3

davidjgonzalez:


davidjgonzalez:
@kwin - Ah - some folks indicated it fixed a security issues but that doesnt appear to be the case per the readme.

PR #3317, opened 2024-04-09
Use compile scope to transitively inherit acs-aem-commons-bundle

kwin:
This eases introducing Cloud-only API as the consumer should only depend on "acs-aem-commons-bundle-cloud" instead of "acs-aem-commons-bundle" (when running in AEMaaCS)


kwin:
It is hard to exclude a dependency including its transitive dependencies in https://maven.apache.org/enforcer/enforcer-rules/bannedDependencies.html. Also I stumbled upon https://issues.apache.org/jira/browse/MENFORCER-502 which leads to different behaviour when building the module standalone (success) vs in the multimodule build (failure)

PR #3332, opened 2024-05-01
Removed some Flex SCR annotations and replaced with OSGi annotations

davidjgonzalez:


bundle/src/main/java/com/adobe/acs/commons/http/injectors/AbstractHtmlRequestInjector.java
```diff
@@ -35,10 +35,10 @@
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang3.ArrayUtils;
-import org.apache.felix.scr.annotations.Deactivate;
 import org.apache.sling.api.SlingHttpServletRequest;
 import org.osgi.framework.ServiceRegistration;
 import org.osgi.service.component.ComponentContext;
+import org.osgi.service.component.annotations.Deactivate;
```
adamcin, line 41: Because this is a public abstract class, I recommend leaving the SCR annotation in place here, in addition to the DS annotation, so that consumers using Felix SCR tooling will still find the old annotation in the class file.

bundle/src/main/java/com/adobe/acs/commons/util/ResourceServiceManager.java
```diff
@@ -38,6 +38,8 @@
 import org.osgi.framework.ServiceReference;
 import org.osgi.framework.ServiceRegistration;
 import org.osgi.service.component.ComponentContext;
+import org.osgi.service.component.annotations.Activate;
```
adamcin, line 41: Because this is a public abstract class, I recommend leaving the SCR annotation in place here, in addition to the DS annotation, so that consumers using Felix SCR tooling will still find the old annotation in the class file.

adamcin (changes requested):
A couple public abstract classes should probably retain the SCR annotation variants in addition to the DS annotations. Because they don't have `@Component` annotations it shouldn't result in duplicate or conflicting SCR descriptors for them.

bundle/src/main/java/com/adobe/acs/commons/util/ResourceServiceManager.java
davidjgonzalez: @adamcin Are you saying it should have both/2 @Activate annotations? One from felix.scr and other from osgi.annotations?

Curious - what Felix SCR tooling might someone still be using today?

adamcin: Yes, exactly. As in:

```
@Activate
@org.apache.felix.scr.annotations.Activate
protected void activate() {
}
```

Any customers with components that extend from these classes without overriding the annotated method would still have to be using SCR annotations. (Using DS component inheritance is discouraged, but it is supported by default with maven-bundle-plugin when using SCR annotations).

kwin: I wouldn’t consider such edge cases. Too much maintenance overhead and easy to fix on consumer side.

bundle/src/main/java/com/adobe/acs/commons/packaging/impl/ACLPackagerServletImpl.java
```diff
@@ -46,17 +45,26 @@
...
-@SlingServlet(
-        methods = { HttpConstants.METHOD_POST },
-        resourceTypes = { "acs-commons/components/utilities/packager/acl-packager" },
-        selectors = { "package" },
-        extensions = { "json" }
+
+@Component(
+        service = {Servlet.class},
+        property = {
+                SLING_SERVLET_RESOURCE_TYPES + "=acs-commons/components/utilities/packager/acl-packager",
```
kwin, line 62: Let us use Sling Servlet Property Type Annotations: https://github.com/apache/sling-org-apache-sling-servlets-annotations

PR #3380, opened 2024-07-11
Remove forced theme from new system notifications

scotty6435:
The recently amended system notifications dialog has been re-styles using CoralUI 3 elements but for some reason a hardcoded red theme was applied to the content area.

In local testing, all alert themes set appropriate colours when this CSS was removed. I can't find a reference to why this was added, maybe it was required only for an earlier version of the code before the themes were implemented?


scotty6435:
Before:
before
After:
after

PR #3389, opened 2024-07-25
Replace usage of commons collections 3 with commons collections 4

cziegeler:
Commons collections 3 is outdated and not maintained anymore, this change updates all usage to commons collections 4 - which is already used in some places

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 
 ## Unreleased ([details][unreleased changes details])
 
+### Changed
+
+- #3390 - Remove usage of commons collections 3
+
 ## 6.6.2 - 2024-06-25
 
 ### Fixed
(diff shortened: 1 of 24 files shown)


kwin:
As long as it is provided by the uber-jar it will be tricky to prevent those imports from sneaking in again. Any reason why even https://repo1.maven.org/maven2/com/adobe/aem/uber-jar/6.5.21/uber-jar-6.5.21.jar still contains Collections 3? I would expect that to be only provided in https://repo1.maven.org/maven2/com/adobe/aem/uber-jar/6.5.21/uber-jar-6.5.21-apis-with-deprecations.jar....

kwin:
Is there maybe some configuration on https://github.com/adobe/aemanalyser-maven-plugin/tree/main/aemanalyser-maven-plugin#configuration which could be leveraged to make the build fail in case of using deprecated API (is there even an analyser task for it)?

cziegeler:
@kwin I agree that it would be nice to prevent reintroduction of usage of such APIs. The best place is a PR review - not fool proof of course.
The aemanalyser maven plugin checks such API usage and by default WARNs. It can be changed to fail.
(Site note: the reporting in the latest versions of the plugin is currently broken)

cziegeler:
Is anything missing to apply this PR?

kwin (approved):

cziegeler:
Can someone please merge this?

PR #3402, opened 2024-08-05
Creations or updates of EnsureOakIndex configurations get picked up

jasper-s:
The EnsureOakIndexManagerImpl was not consistently picking up changes to EnsureOakIndex configurations, as well as ignoring additional configurations until it was restarted.
Setting the referencePolicyOption to Greedy guarantees additional services are picked up when they become active, while the immediate flag makes them active immediately.


bundle/src/main/java/com/adobe/acs/commons/oak/impl/EnsureOakIndex.java
```diff
@@ -48,7 +48,8 @@
         description = "Component Factory to manage Oak indexes.",
         configurationFactory = true,
         policy = ConfigurationPolicy.REQUIRE,
-        metatype = true
+        metatype = true,
+        immediate = true
```
kwin, line 52: No need for this, they will be started ASAP if they are referenced in greedy fashion.

jasper-s: It seems like my AEM instance disagrees with that statement. I also expected that, but without `immediate=true` all EnsureOakIndex components remain in the "satisfied" state

jasper-s: I have looked around in OSGi documentation and could not find any indication that the greedy reference makes immediate obsolete.

kwin: According to https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#service.component-delayed.component every registered component is becoming active as soon as it is referenced. As greedy references always look for new services in the service registry (https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#d0e37698) and bind them it shouldn't make a difference if the service is immediately activated or just activated once bound.

jasper-s: @kwin, any suggestions on how to move forward?

kwin (approved):

bundle/src/main/java/com/adobe/acs/commons/oak/impl/EnsureOakIndex.java
jasper-s: I did some more digging around. While the EnsureOakIndex references do get bound to the EnsureOakIndexManagerImpl component, the EnsureOakIndex components remain satifsfied.

It seems like the binding of the references is happening as described in the documentation, but that the components are not actived as expected.

jasper-s: The cause seems to be the circular reference where EnsureOakIndex components reference the EnsureOakIndexManager. When removing that reference, activation of all components works as expected without the immediate property.
The reference is required of course, so the addition of immediate seems to be the most realistic option.

jasper-s: I updated the circular reference to make that optional & greedy. Now immediate is no longer required.
@kwin, let me know which you prefer.

Issue #3401, opened 2024-08-05
SyslogAppender not usable for AEMaaCS


cziegeler:
The SyslogAppender is not usable on Cloud Service, it also imports logback classes which are deprecated on cloud service and will be removed in the future.
I think there are two options:
1. Create an onprem bundle and move this component to that bundle
2. Remove the functionality entirely - not sure how much this is used

cziegeler:
@davidjgonzalez  WDYT?

PR #3407, opened 2024-08-07
Move SyslogAppender into separate onprem bundle

cziegeler:
This implements #3401


bundle-onprem/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+                </exclusion>
+                <exclusion>
+                    <groupId>org.apache.jackrabbit</groupId>
+                    <artifactId>jackrabbit-api</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <!-- End Test dependencies -->
+        <!-- the aem-sdk-api should be the last dependency, as other dependencies should take precedence in case of package clashes -->
+        <dependency>
```
kwin, line 114: This is the wrong dependency for AEM 6.5

bundle-onprem/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+            </plugin>
+        </plugins>
+    </build>
+
+    <dependencies>
+        <dependency>
```
kwin, line 67: For a new bundle we should IMHO start with the most recent dependencies (i.e. JUnit5)

bundle-onprem/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+            <artifactId>org.apache.sling.testing.sling-mock.junit4</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.wcm</groupId>
+            <artifactId>io.wcm.testing.aem-mock.junit4</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.sling</groupId>
```
kwin, line 93: Hopefully we don't need this legacy dep.

bundle-onprem/pom.xml
cziegeler: good point, fixed

bundle-onprem/pom.xml
cziegeler: Right now we do, I just ported the test as is

bundle-onprem/pom.xml
```diff
@@ -0,0 +1,195 @@
...
+    <!-- ====================================================================== -->
+
+    <artifactId>acs-aem-commons-bundle-onprem</artifactId>
+    <name>ACS AEM Commons Bundle - On Prem</name>
+    <description>Core ACS AEM Commons OSGi Bundle for AEM on prem only</description>
+   <build>
+        <plugins>
+            <plugin>
+                <groupId>biz.aQute.bnd</groupId>
+                <artifactId>bnd-maven-plugin</artifactId>
```
kwin, line 42: I am wondering whether we should make this a fragment of `com.adobe.acs.acs-aem-commons-bundle` similar to the Cloud bundle in order to consume non-exported classes from it.

cziegeler: I thought about this, too - not sure how likely that use case actually is and therefore decided to go with a plain old bundle.

bundle-onprem/pom.xml
cziegeler: Updated the test to use Mockito for everything

bundle-onprem/pom.xml
cziegeler: Updated to JUnit5

bundle-onprem/pom.xml
kwin: ok, we can do it later on if needed.

cziegeler: @kwin  Thanks, is there anything else to be done?

cziegeler:
@kwin Could you please merge this PR - or is there anything outstanding?

kwin (approved):

PR #3409, opened 2024-08-09
Reduce usage of Apache Commons Lang 2

### LINKED ISSUE #3408: Reduce usage of Apache Commons Lang 2
Apache commons lang 2 is not maintained anymore since years. Usage of it should be replaced with Apache Commons Lang 3. Both libraries are part of AEM as a Cloud Service and AEM 6.x, only a change of the imports is required.

There is one exception, the usage of AEMs MailTemplate API which currently does not provide an alternative. Once that is available we can change the last remaining usage of Apache commons lang 2.

cziegeler:
Replace usage of Apache Commons Lang 2 with Apache Commons Lang 3
As noted in #3408 there are two places remaining (MailTemplate) which can be replaced as soon as there is an alternative.
This fixes #3408


cziegeler:
@kwin You were right, we should convert the on-prem bundle into a fragment; I did this as part of this PR as we will most likely need this once an alternative for the MailTemplate is available.

cziegeler:
@kwin Do you think we can already apply this? The remaining usage will be replaced once an alternative API is available for Cloud Service

bundle-onprem/pom.xml
```diff
@@ -33,13 +33,18 @@
     <!-- ====================================================================== -->
 
     <artifactId>acs-aem-commons-bundle-onprem</artifactId>
-    <name>ACS AEM Commons Bundle - On Prem</name>
-    <description>Core ACS AEM Commons OSGi Bundle for AEM on prem only</description>
+    <name>ACS AEM Commons - Bundle Fragment (for On Prem</name>
```
kwin, line 36: This is missing a closing parenthesis

kwin (approved):
Looks good except for one small glitch in the Maven name.

bundle-onprem/pom.xml
cziegeler: Thanks for spotting, fixed now

PR #3429, opened 2024-09-09
fix: add uniq function to embedded lodash library

jfmitchell:
The composite multifield widget relies on the uniq function - so this adds it. Still based on v4.17.21

### DIFF
--- a/ui.apps/src/main/content/jcr_root/apps/acs-commons/clientlibs/vendor/lodash/lodash-v4.17.21.js
+++ b/ui.apps/src/main/content/jcr_root/apps/acs-commons/clientlibs/vendor/lodash/lodash-v4.17.21.js
@@ -1,7 +1,7 @@
 /**
  * @license
  * Lodash (Custom Build) <https://lodash.com/>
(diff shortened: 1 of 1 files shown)


YegorKozlov (approved):

kwin:
@jfmitchell Can you add an issue which describes describe where and how this error manifests itself? The changelog is supposed to reference issues (not PRs) to give some more written context.

cxc38:
@kwin the Lodash lib was upgraded from 4.17.15 to 4.17.21 during the 6.3.4 release. During the upgrade, the Lodash lib was rebuilt for 4.17.21. Lodash lib can have core build, full build, and custom build. 4.17.15 embedded in acs-aem-common 6.3.2 was a full build. However, the embedded 4.17.21 lib is just a core build, which has minimum functions in it. It missed some functions which are required by multifields widget, typically the uniq() function in Array category. This caused the form using multifields widget failed to populate the content when we upgraded the acs-aem-common to 6.6.2 from a older version. I believe it affects all acs-aem-common releases since 6.3.4.

Issue #3433, opened 2024-09-25
Extract OSGI configuration from acs-aem-commons-all to a new package acs-aem-commons-ui.config


SzymonSzafraniec:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: n/a
* [ ] ACS AEM Commons Version: 6.6.4
* [ ] Reproducible on Latest? yes

### Expected Behavior

Package acs-aem-commons-all should not contain config files. OSGI Configuration files should be located in a separate package called acs-aem-commons-ui.config embedded in acs-aem-commons-all.

See: https://experienceleague.adobe.com/en/docs/experience-manager-cloud-service/content/implementing/developing/aem-project-content-package-structure#container-packages

### Actual Behavior

OSGI Configuration files are located directly in acs-aem-commons-all (src/main/content/jcr_root/apps/acs-commons/config.*). This package embeds other packages and bundles as well as individual configuration files.

### Steps to Reproduce

n/a

### Links

https://experienceleague.adobe.com/en/docs/experience-manager-cloud-service/content/implementing/developing/aem-project-content-package-structure

davidjgonzalez:
Release in 6.7.0

PR #3434, opened 2024-09-26
Extract OSGI configuration to new package ui.config

SzymonSzafraniec:
Corresponding issue: https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/3433

Currently acs-aem-commons-all package mixes subpackages with configuration files. This change bundles the configurations in a separate package which is also included in the all package.

Another benefit is that it allows other projects to include individual ACS Commons packages while not missing the OSGI configurations (e.g. include the acs-aem-commons-bundle and acs-aem-commons-ui.config only). 

Also, by including individual packages separately it's possible to control the packages installation order. This is currently impossible since including the acs-aem-commons-all is the only way to install OSGI Configurations.


davidjgonzalez:
@kwin  any problems with this?

ui.config/pom.xml
```diff
@@ -0,0 +1,170 @@
...
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <!-- ====================================================================== -->
+    <!-- P A R E N T P R O J E C T D E S C R I P T I O N -->
+    <!-- ====================================================================== -->
+    <parent>
+        <groupId>com.adobe.acs</groupId>
+        <artifactId>acs-aem-commons</artifactId>
+        <version>6.6.5-SNAPSHOT</version>
```
kwin, line 28: This is the wrong version and therefore does build.

ui.config/pom.xml
```diff
@@ -0,0 +1,170 @@
...
+        </plugins>
+    </build>
+    <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>acs-aem-commons-oakpal-checks</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
```
kwin, line 133: I don't think this requires logically any dependencies

kwin (changes requested):
Looks good in general, just requires some update due to the latest merge.

ui.config/pom.xml
davidjgonzalez, line 28:
```suggestion
        <version>6.7.0-SNAPSHOT</version>
```

ui.config/pom.xml
```diff
@@ -0,0 +1,170 @@
...
+            <artifactId>acs-aem-commons-bundle</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <!-- put UberJar last so that more specific artifacts take precedence -->
+        <dependency>
+            <groupId>com.adobe.aem</groupId>
+            <artifactId>uber-jar</artifactId>
+            <scope>provided</scope>
+        </dependency>
+    </dependencies>
```
davidjgonzalez, line 144:
```suggestion
```

ui.config/pom.xml
```diff
@@ -0,0 +1,151 @@
...
+        <plugins>
+            <plugin>
+                <groupId>org.apache.jackrabbit</groupId>
+                <artifactId>filevault-package-maven-plugin</artifactId>
+                <configuration>
+                    <packageType>container</packageType>
+                </configuration>
+            </plugin>
+
+            <plugin>
```
kwin, line 51: I think filevault validation is enough for this package.

ui.config/pom.xml
```diff
@@ -0,0 +1,151 @@
...
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>license-maven-plugin</artifactId>
+                <configuration>
+                    <excludes>
+                        <exclude>**/*.xml</exclude>
+                    </excludes>
+                </configuration>
+            </plugin>
```
davidjgonzalez, line 123:
```suggestion           
```

PR #3450, opened 2024-10-17
Adjusted exclusions to use regex, and force Cloud Manager-required pa…

davidjgonzalez:
…ths into exclusion list


pat-lego (commented):
Works perfectly

pat-lego (approved):

PR #3452, opened 2024-10-19
#3443 Content Sync: don't drill down into content tree if recursion is off

### LINKED ISSUE #3443: Content Sync without Recurse slow as though it's still analyzing children
## Required Information

* [x] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 2024.9.17964.20240924T170544Z
* [x] ACS AEM Commons Version: 6.7.0

### Expected Behavior

_Please describe briefly the expected behavior, i.e. when I do X, Y should happen._
When content syncing a sites page with a lot of children pages but unchecking the recurse button, the content sync should quickly synchronize the one site page.

### Actual Behavior

_Please describe briefly the actual behavior, i.e. when I do X, Z happens._
I get a 503 as though it is still processing all children just in the same way it does when the Recurse button is still checked.

### Steps to Reproduce

_Starting from a vanilla instance of AEM at the above version, describe in explicit steps how to reproduce. Link to
any supporting assets, e.g. content packages, below_
Parent page with LOTS of content underneath it.

### Links

_Links to related assets, e.g. content packages containing test components_

YegorKozlov:
fixes #3443

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/contentsync/impl/LastModifiedStrategy.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/contentsync/impl/LastModifiedStrategy.java
@@ -54,33 +54,50 @@ public class LastModifiedStrategy implements UpdateStrategy {
     @Reference
     private ServletResolver servletResolver;
 
+    /**
+     * The ContentCatalog class provides methods to fetch and process content catalogs
+     * from a remote instance.
+     */
     @Override
     public List<CatalogItem> getItems(SlingHttpServletRequest request) {
         String rootPath = request.getParameter("root");
         if (rootPath == null) {
             throw new IllegalArgumentException("root request parameter is required");
         }
-        boolean recursive = "true".equals(request.getParameter("recursive"));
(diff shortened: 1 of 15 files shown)


bundle/src/main/java/com/adobe/acs/commons/contentsync/impl/LastModifiedStrategy.java
```diff
@@ -54,33 +54,50 @@ public class LastModifiedStrategy implements UpdateStrategy {
...
+     * from a remote instance.
+     */
     @Override
     public List<CatalogItem> getItems(SlingHttpServletRequest request) {
         String rootPath = request.getParameter("root");
         if (rootPath == null) {
             throw new IllegalArgumentException("root request parameter is required");
         }
-        boolean recursive = "true".equals(request.getParameter("recursive"));
+        boolean nonRecursive = "false".equals(request.getParameter("recursive"));
```
kwin, line 67: `Boolean.valueOf(...)` to be case insensitive maybe?

kwin (approved):

PR #3457, opened 2024-10-22
Granite include top level fixes

niekraaijmakers:
Currently, properties on the top level of the snippet are ignored and not included.
This forces you to put them on the include level.

With this PR, the snippet top level values count, unless overridden by specifying it on the include level.
Also, the default resourceType is now set to the container resource type.


davidjgonzalez (approved):

ui.apps/src/main/content/jcr_root/apps/acs-commons/granite/ui/components/include/include.jsp
```diff
@@ -22,5 +22,5 @@
     NamespacedTransformedResourceProvider transformedResourceProvider = sling.getService(NamespacedTransformedResourceProvider.class);
     Resource resourceWrapper = transformedResourceProvider.transformResourceWithNameSpacing(slingRequest, targetResource);
 
-    cmp.include(resourceWrapper, cfg.get("resourceType", String.class), cmp.getOptions());
-%>
\ No newline at end of file
+    cmp.include(resourceWrapper, cfg.get("resourceType", "granite/ui/components/coral/foundation/container"), cmp.getOptions());
```
rbotha78, line 25: This was a breaking change. All snippets included now need to also include the resourceType property matching the resourceType of the snippet or it just renders blank where previously it would use the resourceType as contained in the including snippet.

PR #3466, opened 2024-11-06
fix: syntax error in default errorpagehandler

jdavenport-vervio:
Syntax error in default errorpagehandler causes a lot of AEM functionality to break. This was due to a syntax error where the `&&` in the JSP file was duplicated.


niekraaijmakers (approved):
good catch.

jdavenport-vervio:
@davidjgonzalez can you please review this?

davidjgonzalez (approved):

PR #3399, opened 2024-08-01
Create "redirect" resource as type "sling:Folder"

kwin:
This is allowed even below "nt:folder" resources and does not impose any restrictions.

This closes #3398

PR #3479, opened 2024-11-24
Bugfix: Fixed Configuration Model for Redirect Manager after change in "redirect" resource as "sling:Folder"

vabs95:
### **Cause**:  
The resource type for redirects has been changed to `"sling:Folder"`. However, the JCR query used to display the list of configurations has not been updated and currently only retrieves nodes of type `"nt:unstructured"`.

### **Fix**:  
Refactor Configurations Model to read the configs directly without query. 

Regression via https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/3399

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com)
 
 ## Unreleased ([details][unreleased changes details])
 
+### Fixed
+- #3479 - Fixed Configurations Model for Redirect Manager after change in "redirect" resource as "sling:Folder"
+
 ## 6.9.6 - 2024-11-20
 
 ### Fixed
--- a/bundle/src/main/java/com/adobe/acs/commons/redirects/models/Configurations.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/redirects/models/Configurations.java
@@ -43,7 +43,6 @@ import java.util.List;
  */
 @Model(adaptables = SlingHttpServletRequest.class)
 public class Configurations {
-    private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
     @SlingObject
     private SlingHttpServletRequest request;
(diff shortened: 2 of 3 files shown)


vabs95:
FYI @kwin and @YegorKozlov

YegorKozlov (changes requested):
@vabs95  a good catch!

I'd say we should rewrite the code further and get rid of the jcr query at all. We can simply iterate over `/conf/*`, something like:

```
    public Collection<RedirectConfiguration> getConfigurations() {
        List<RedirectConfiguration> lst = new ArrayList<>();
        String bucketName = redirectFilter == null ? RedirectFilter.DEFAULT_CONFIG_BUCKET : redirectFilter.getBucket();
        String configName = redirectFilter == null ? RedirectFilter.DEFAULT_CONFIG_NAME : redirectFilter.getConfigName();
        Resource conf = request.getResourceResolver().getResource("/conf");
        for (Resource confRoot : conf.getChildren()) {
            String storageSuffix = bucketName + "/" + configName;
            Resource res = confRoot.getChild(storageSuffix);
            if (res != null && res.isResourceType(REDIRECTS_RESOURCE_TYPE)) {
                RedirectConfiguration cfg = new RedirectConfiguration(res, storageSuffix);
                lst.add(cfg);
            }
        }
        lst.sort(Comparator.comparing(RedirectConfiguration::getName));
        return lst;
    }
```

It would be also great to update the tests. There are tests for CreateRedirectConfigurationServlet, e.g.
https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/213279b88757a3798893397f09ec74e39f16a1b3/bundle/src/test/java/com/adobe/acs/commons/redirects/servlets/CreateRedirectConfigurationServletTest.java#L60C1-L61C1

but the way it reads the created resources is not aligned with the code.
instead of reading the resource directly:

```
        Resource cfg = context.resourceResolver().getResource("/conf/global/settings/redirects");
```
the test should read via the model:
```
        Configurations confModel = context.request().adaptTo(Configurations.class);
        Collection<RedirectConfiguration> configurations = confModel.getConfigurations();
        RedirectConfiguration cfg = configurations.iterator().next();
        assertEquals("/conf/global/settings/redirects", cfg.getPath());
```

this way the test will cover *both* the create and read parts.

vabs95:
@YegorKozlov  Changes done as requested.

YegorKozlov (approved):

davidjgonzalez:
Neither here nor there, but if you do write a query, and Oak determines that the faster way to collect the results is to just traverse the tree, it will traverse the tree. The overhead of the query is negligible.

PR #1345, opened 2018-04-25
add license checking to codeclimate/checkstyle configuration. Fixes #…

justinedelson:
…1343

Also fixing all the current license issues

PR #3484, opened 2024-11-30
A servlet to export redirects to a TXT file to use with pipeline-free redirects

YegorKozlov:
@krystiannowak @kwin @joerghoh 
the PR adds a servlet to export redirects into a text file, e.g. http://localhost:4502/conf/my-site/settings/redirects.txt 

```
# Redirect Map File
/abc2 /content/applications
/xyz /legacy-1
/content/applications1 /legacy-2
```

This way Redirect Manager  can serve redirects for [Pipeline-free URL Redirects](https://experienceleague.adobe.com/en/docs/experience-manager-cloud-service/content/implementing/content-delivery/pipeline-free-url-redirects) 

`src/opt-in/managed-rewrite-maps.yaml` in the dispatcher configuration would look like
```yaml
maps:
- name: my-site.map
  path: /conf/my-site/settings/redirects.txt 
- name: another-site.map
  path: /conf/another-site/settings/redirects.txt 
```

and vhosts can reference redirects like
```
# RewriteMap from managed rewrite maps
RewriteMap my-site.foo dbm=sdbm:/tmp/rewrites/my-site.map
RewriteCond ${my-site.foo:$1} !=""
RewriteRule ^(.*)$ ${my-site.foo:$1|/} [L,R=301]
```

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/redirects/servlets/RewriteMapServlet.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/redirects/servlets/RewriteMapServlet.java
@@ -1,8 +1,9 @@
-/*
+/*-
+ * #%L
  * ACS AEM Commons Bundle
- *
+ * %%
  * Copyright (C) 2013 - 2024 Adobe
- *
+ * %%
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
@@ -14,6 +15,7 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ * #L%
  */
 package com.adobe.acs.commons.redirects.servlets;
 
@@ -51,6 +53,7 @@ public class RewriteMapServlet extends SlingSafeMethodsServlet {
     private static final long serialVersionUID = -3564475196678277711L;
 
     @Override
+    @SuppressWarnings("java:S3457")
     protected void doGet(SlingHttpServletRequest request, SlingHttpServletResponse response)
             throws ServletException, IOException {
         response.setContentType(ContentType.TEXT_PLAIN.getMimeType());
--- a/bundle/src/main/java/com/adobe/acs/commons/redirects/servlets/RewriteMapServlet.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/redirects/servlets/RewriteMapServlet.java
@@ -1,5 +1,5 @@
 /*
- * ACS AEM Commons
+ * ACS AEM Commons Bundle
  *
  * Copyright (C) 2013 - 2024 Adobe
  *
@@ -62,16 +62,16 @@ public class RewriteMapServlet extends SlingSafeMethodsServlet {
         }
         Collection<RedirectRule> rules = RedirectFilter.getRules(request.getResource());
         PrintWriter out = response.getWriter();
-        out.printf("# %s Redirects\n", statusCode == 0 ? "All" : "" + statusCode);
+        out.printf("# %s Redirects%n", statusCode == 0 ? "All" : "" + statusCode);
         for (RedirectRule rule : rules) {
             if(statusCode != 0 && rule.getStatusCode() != statusCode) {
                 continue;
             }
             String note = rule.getNote();
             if(note != null && !note.isEmpty()) {
-                out.printf("# %s\n", note);
+                out.printf("# %s%n", note);
             }
-            out.printf("%s %s\n", rule.getSource(), rule.getTarget());
+            out.printf("%s %s%n", rule.getSource(), rule.getTarget());
         }
     }
 }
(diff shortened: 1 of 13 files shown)


bundle/src/test/java/com/adobe/acs/commons/redirects/servlets/RewriteMapServletTest.java
```diff
@@ -0,0 +1,111 @@
+/*
+ * ACS AEM Commons
+ *
+ * Copyright (C) 2013 - 2023 Adobe
```
krystiannowak, line 4: @YegorKozlov you might want to update the year range to include 2024 as current year, too ;)

bundle/src/main/java/com/adobe/acs/commons/redirects/servlets/RewriteMapServlet.java
```diff
@@ -0,0 +1,70 @@
+/*
+ * ACS AEM Commons
+ *
+ * Copyright (C) 2013 - 2023 Adobe
```
krystiannowak, line 4: @YegorKozlov you might want to update the year range to include 2024 as current year, too ;)

krystiannowak (changes requested):
@YegorKozlov just small note about copyrights year range, otherwise ok - good to see another source for Pipeline-free URL Redirects map data!

cc: @joerghoh @kwin

krystiannowak:
@YegorKozlov you might also add an option (e.g. selector or param) to separately generate list of entries for 301 redirection, and separately for 302 ones, so that on Apache HTTPD level they can be treated separately

krystiannowak (approved):
@YegorKozlov super, thanks!

krystiannowak:
@YegorKozlov looks like Windows-based test runs cry now

bundle/src/main/java/com/adobe/acs/commons/redirects/servlets/RewriteMapServlet.java
```diff
@@ -62,16 +62,16 @@ protected void doGet(SlingHttpServletRequest request, SlingHttpServletResponse r
         }
         Collection<RedirectRule> rules = RedirectFilter.getRules(request.getResource());
         PrintWriter out = response.getWriter();
-        out.printf("# %s Redirects\n", statusCode == 0 ? "All" : "" + statusCode);
+        out.printf("# %s Redirects%n", statusCode == 0 ? "All" : "" + statusCode);
```
krystiannowak, line 65: @YegorKozlov after starting to use %n as line separator, splitting only by \n in tests by
```
String[] lines = response.getOutputAsString().split("\n");
```
in https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/3484/commits/858fda63ef67e26ec8d0fa5f6424a2e94557fc19#diff-36adc0c81c9585f4cadf9e45117faabfdda5ef6d7e197ade7d39e0a896c6853eR87
might not be precise enough anymore on Windows

bundle/src/main/java/com/adobe/acs/commons/redirects/servlets/RewriteMapServlet.java
```diff
@@ -1,8 +1,9 @@
-/*
+/*-
+ * #%L
  * ACS AEM Commons Bundle
- *
+ * %%
  * Copyright (C) 2013 - 2024 Adobe
- *
+ * %%
```
krystiannowak, line 6: @YegorKozlov I'm not sure if that makes sense - or should rather https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/.codeclimate/header.txt be tuned instead? Also the other Java files do not seem to adhere to this pattern.

@kwin @joerghoh @davidjgonzalez WDYT?

YegorKozlov: @krystiannowak  this `%%` thing does not make much sense to me, but I believe we've always had it, since the header check was introduced in https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/1345 back in 2018

> Also the other Java files do not seem to adhere to this pattern.

Yep. We have quite a bit of the header variations across the project.

kwin:
Look at https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2500 for a discussion on license headers.

krystiannowak:
@YegorKozlov Probably you need to "touch" it to rerun tests, as I can see there are "status code: 504, reason phrase: Gateway Time-out" errors

krystiannowak (approved):

krystiannowak:
@YegorKozlov looks like checks passed now, but my ack is not strong enough - maybe it needs an ack from other parties like @kwin @joerghoh @davidjgonzalez etc?

krystiannowak:
@YegorKozlov Don't you want to update https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io/blob/master/_acs-aem-commons/features/redirect-manager/index.md as well to reflect those changes?

cc: @davidjgonzalez

YegorKozlov:
> @YegorKozlov Don't you want to update https://github.com/Adobe-Consulting-Services/adobe-consulting-services.github.io/blob/master/_acs-aem-commons/features/redirect-manager/index.md as well to reflect those changes?
> 
> cc: @davidjgonzalez

Sure thing. I was going to give it shot first, but didn't yet get around.

Issue #3471, opened 2024-11-13
EmailService not working due to unsatisfied reference to MailTemplateManager in AEM on prem


bbruns1926:
## Required Information

* [ ] AEM Version, including Service Packs, Cumulative Fix Packs, etc: 6.5.21
* [ ] ACS AEM Commons Version: 6.9.0+
* [ ] Reproducible on Latest? yes

### Expected Behavior

ACS Commons EmailService works in AEM on prem.

### Actual Behavior

With https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/3422 (ACS Commons 6.9.0) MailTemplateManager was introduced and added as a reference to EmailService and the implementation for the MailTemplateManager was placed in bundle-cloud and bundle-onprem. Unfortunately the bundle-onprem is missing in acs-aem-commons-all-6.9.*.zip files. Looks like the dependency entry for bundle-onprem is missing in all/pom.xml.
I did a quick try and after adding the dependency to all/pom.xml bundle-onprem was part of the resulting zip file but the build failed with:
`[ERROR] [artifact-rules] com.adobe.acs:acs-aem-commons-bundle-onprem:6.9.5-SNAPSHOT: The Adobe ACS Commons onprem bundle is not supported (adobe/consulting:acs-aem-commons-all:6.9.5-SNAPSHOT)`

In addition to the missing bundle in the zip file it looks like something with the implementation is not working. I manually installed a locally build acs-aem-commons-bundle-onprem:6.9.4 but then the following error occurred:
`java.lang.ClassNotFoundException: com.adobe.acs.commons.email.impl.MailTemplateManager not found by acs-aem-commons-bundle-onprem`

### Steps to Reproduce

For the first part of the issue AEM is not needed to reproduce this. Just open acs-aem-commons-all-6.9.*.zip and go to jcr_root\apps\acs-commons\install. There only acs-aem-commons-bundle is present and acs-aem-commons-bundle-onprem is missing. 
With AEM navigate to http://localhost:4502/system/console/components and search for com.adobe.acs.commons.email.impl.EmailServiceImpl. The status will be "unsatisfied (reference)"

For the second part navigate to http://localhost:4502/system/console/components and search for com.adobe.acs.commons.email.impl.MailTemplateManagerImpl. The status will be "satisfied" and in the error.log of AEM there will be messages like:
[...]

PR #3488, opened 2024-12-05
Fix 3471

davidjgonzalez:
2024-12-04 at 7 35 PM


cziegeler:
I am not sure that this is the right fix, see discussion in #3471

PR #3517, opened 2025-01-30
fix(bugfix/3516): 'Redirect Manager' - set 'jcr:primaryType' of impor…

sabdouni:
…ted 'redirect row' as 'nt:unstructured'


YegorKozlov:
duplicates a change in https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/3504

Issue #3547, opened 2025-03-26
Allow to turn off the ContentPolicyValueInjector


joerghoh:
Whenever the bundle is active, the ContentPolicyValueInjector is active; and under circumstances it is possible, that this injector is tried, even if it's never planned to be active.
And if the injection is tried, the creation of the ContentPolicy can be an expensive operation; which is is a waste of time, if the injection of a ContentPolicy was never intended or planned.

For that reason it must be possible to turn of the injector. Would have been better if this would have been the default since the beginning, but here we are :-|

kwin:
How is that injector different from all others in https://github.com/Adobe-Consulting-Services/acs-aem-commons/tree/master/bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl. Currently all of them by active by default...

joerghoh:
This injector is not different in any way. It's just that it can be very expensive to try it out for no result.

PR #3549, opened 2025-03-27
#3457 allow disabling the ContentPolicyValueInjector

joerghoh:
by default it should be enabled (not to break b/w compatibility), fixing #3547

### DIFF
--- a/bundle/pom.xml
+++ b/bundle/pom.xml
@@ -562,6 +562,12 @@
         </dependency>
         <dependency>
             <groupId>org.mockito</groupId>
+            <artifactId>mockito-inline</artifactId>
+            <scope>test</scope>
+            <version>5.2.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
             <artifactId>mockito-junit-jupiter</artifactId>
             <scope>test</scope>
         </dependency>
--- a/bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/ContentPolicyValueInjector.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/ContentPolicyValueInjector.java
@@ -62,7 +62,7 @@ public class ContentPolicyValueInjector  implements Injector {
     Configuration config;
 
     @Activate
-    public void activate (Configuration c) {
+    public void activate(Configuration c) {
         this.config = c;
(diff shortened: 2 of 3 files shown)


niekraaijmakers (approved):
LGTM. why not?

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/ContentPolicyValueInjector.java
```diff
@@ -61,15 +77,25 @@ public String getName() {
...
-
         return null;
     }
 
 
+    @ObjectClassDefinition(name="ACS AEM Commons ContentPolicyValueInjector")
+    public @interface Configuration {
+
+        @AttributeDefinition(name="enabled")
+        public boolean enabled() default true;
```
kwin, line 97: Even with the risk of breaking existing consumers I would switch to false (i.e. opt in) like for all other ACS AEM commons features and just mention this in the release notes

kwin:
What about all other injectors?

bundle/pom.xml
```diff
@@ -560,6 +560,12 @@
             <artifactId>mockito-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-inline</artifactId>
```
kwin, line 565: https://stackoverflow.com/a/76095346

bundle/src/main/java/com/adobe/acs/commons/models/injectors/impl/ContentPolicyValueInjector.java
joerghoh: @kwin +1

@niekraaijmakers as you have added this injector a year ago, do you have any insights into its usage?

joerghoh:
> What about all other injectors?

I have not seen them yet as problematic wrt to performance; but when we are at it, we should not only disable one Injector by default, but rather all of them.

niekraaijmakers:
I don't think many customers use this yet. I'd say it's fine to disable it. Regarding other injectors, I think that might be a bad idea.

kwin:
Maybe that was not clear enough: https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/3549#discussion_r2018856978.

Build fails on Java 21 with this. @joerghoh Can you try to come up with an alternative approach for testing which works with Java 21?

Issue #3551, opened 2025-03-31
Redirect Manager shows incorrect Publication status for all redirects stored with Shard stratey


drj101687:
https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectRule.java#L336

This line that checks the publication status of a `redirects` node does not account for the Shard strategy used when many redirects have been authored for a configuration.
eg:

- redirects
  - shard-0
    - redirect-rule-
    - redirect-rule-0
    - redirect-rule-1
    - ...

It needs to be updated to check the publication status of the correct node for both storage strategies.

The publication status still currently works if the configuration does not have a sharded storage tree.
eg: 

- redirects
  - redirect-rule-
  - redirect-rule-0
  - redirect-rule-1
  - ...

YegorKozlov:
fixed in v6.12

PR #3552, opened 2025-03-31
#3551 - added logic to check the correct parent node if the redirect …

drj101687:
…belongs to a sharded collection.

fixes the issue reported in #3551


drj101687:
I don't understand why the JDK 21 ubuntu check failed. GitHub doesn't contain the full output for that check.

Issue #3555, opened 2025-04-07
ACS Commons - Issue Failing for copy content sync with Large Volume


jatankumar-git:
l. We have acs commons version upgraded to latest 6.10.0 and implemented the required setup to make it work.

We are trying to copy content from source to target environment, It works fine for smaller sub set path or pages 1-2k but if we provide a parent level path like 4k 5k pages it throws an service outage error. We were not able to see any errors and tried multiple times without any active deployment still the issue persists.

We could not find any specific details in documentations/blogs to triage the issue or if any limitation for number of pages that can be copied as such. Please look into the issue to identify what is causing this issue.

Adding screenshots for more details.

![Image](https://github.com/user-attachments/assets/ec8ece5a-bd35-4329-8388-21c11ffb252e)

Steps to reproduce- 
Select flags as recursive, incremental (Dry run enabled/disabled) and submit the job, after 3-4 mins we get an service outage error.

Similarly we tried smaller paths it worked and other locale/parent path it fails. We raised a adobe support ticket and they routed us to get a issue raised here as its related to ACS-aem-commons. 

Thanks,
Jatan

YegorKozlov:
What are the source and target AEM instances? Are you sync-ing an AEM Cloud from another AEM cloud instance?

I seems to be a kind of a Timeout error. The target AEM instances makes a call to the source to fetch the list of page to sync. This can take time on large content trees, but I've never seen the error you are getting.

YegorKozlov:
I think I know what's the problem is.

AEM Cloud has the default read timeout of 60 seconds. This means a request will be aborted if it didn't read at least a byte from the socket within 60 seconds. It applies to any servlet in AEM Cloud not responding for longer than 60 sec.  

I theory you can bump it up in the Jetty OSGi config, but I don't recommend it and Adobe wouldn't welcome this change either. 

My plan is to change the mechanics of getting the delta to sync from a request/response to an async job and then poll for completion. 

So instead of sending the http request to the remote AEM instance and waiting for response (which fails after 60 sec) I'm going to change Content Sync to 

 - send http request to the remote instance to create a background job to create the delta
 - poll every N seconds and print status. The socket will read these messages and keep the connection

It's a good catch and it's going to be a good improvement.

jatankumar-git:
Yes, we are syncing from one AEM cloud to another in same program between our non prod environments. 

[...]

PR #3558, opened 2025-04-12
#3555 Content Sync: prevent timeout errors when sync-ing AEM cloud instances with large volumes of data

YegorKozlov:
addresses #3555

AEM Cloud has the default read timeout of 60 seconds and ContentSync fails when collecting resources on the remote instance  for longer than that. 

With this PR the resources are collected in a Sling Job and the tool polls for completion every 3 seconds. 

image


simontunnat:
Sounds like a great solution, would really like for this to be reviewed and merged. :)

davidjgonzalez (approved):

PR #3608, opened 2025-07-18
Completely move everything to the executor, else the circular depende…

royteeuwen:
…ncy remains


royteeuwen:
@kwin  @davidjgonzalez  sadly enough the rework to use OSGi DS was not complete, I had to really move the interface to the Executor.

YegorKozlov (approved):

Issue #3613, opened 2025-08-08
ACS Commons Redirect Manager - Replace Mode Issue


sabdouni:
Technical Env:

AEM Cloud SDK : 2025.7.21772.20250730T182007Z
ACS Commons : 6.14.0

Issue:
We noticed an issue with the Redirect Manager when our marketing team uses 'Replace Mode' on import.
They are getting a 500 error.

Without 'Replace Mode', everything works fine and the imports go smoothly.

After the investigation, we discovered the following

- Under /conf/<site>/settings/redirects, we have _rep:policy_ node
- The servlet com.adobe.acs.commons.redirects.servlets.ImportRedirectMapServlet, try to delete all the nodes under  /conf/<site>/settings/redirects, including _rep:policy_ causing an error

Possible solution:
- Ignore _rep:policy_  when deleting the redirections

I can provide a PR if my analysis is correct.

Thanks

YegorKozlov:
@sabdouni good catch. 

> I can provide a PR if my analysis is correct.

PRs are very much welcome. See
https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/7c88f0de005f80851dac38e5214d328f77a9e16c/bundle/src/main/java/com/adobe/acs/commons/redirects/servlets/ImportRedirectMapServlet.java#L112

it should only delete the `nt:unstructured` nodes which coves redirects and shards. 

```java
if(ch.isResourceType(NT_UNSTRUCTURED)) {
    ch.getResourceResolver().delete(ch);
}
```

YegorKozlov:
@sabdouni the PR is approved and will be included in the next release

PR #3626, opened 2025-09-11
Redirect Manager: Fix Replace Mode issue, delete only 'redirect-row' …

sabdouni:
…nodes (#3613)


YegorKozlov (approved):
looks good to me

PR #3639, opened 2025-10-03
Redirect Manager: fix search in tags property. Fixes #3638

### LINKED ISSUE #3638: Redirect Manager: fulltext search does not work for tags property
**Steps to reproduce:**
- Go to redirect manager options and enable fulltext search
Image

- Create any redirect rule and add tag
- Search by tag

**Expected result:**
Redirect rule with the entered tag is found

**Actual result:**
nothing is found
Image

mpankratiev:


kwin (approved):
Thanks, good catch

PR #3642, opened 2025-10-08
Use Njord for publishing to Central Portal

### LINKED ISSUE #3630: Switch from OSSRH publishing to Central Portal
As publishing via OSSRH is [no longer possible](https://central.sonatype.org/news/20250326_ossrh_sunset/) we need to switch to Central Portal API for publishing releases to Maven Central. I would recommend using [njord](https://github.com/maveniverse/njord) over the closed-source [Sonatype plugin](https://central.sonatype.org/publish/publish-portal-maven/) as the latter is quite invasive (i.e. replaces maven-deploy-plugin via extension), is closed-source and in the past the Sonatype OSSRH plugin was poorly maintained.

kwin:
Get rid of of all mentions of the no longer existing OSSRH service Adjust release documentation accordingly
This closes #3630


davidjgonzalez:
@kwin Ok - i added the token username and password as secrets to the Actions secrets. 

Is there anything else you want to test/try out now that these are in place?

davidjgonzalez:
@kwin  - Sorry; i added the username/token from the old oss.sonatype.org -- updated to a new one from sonatype.com.

I also added GPG_* secrets that are use to deploy if you want to use those to automate releases; do we want to just cut releases when on commit to master?

PR #3645, opened 2025-10-14
Build with Java 25

kwin:
Update several test dependencies.
Remove oakpal as not compatible with Java23+
(https://github.com/adamcin/oakpal/issues/106)


DidierLoiseau:
Is there a release planned for Java 25 compatibility?

kwin:
The library is already fully compatible with Java 25 at runtime. This PR is only about running the build with Java 25!

PR #3647, opened 2025-10-21
Update Sonatype repository URLs in pom.xml

davidjgonzalez:
@kwin  - i think we need to update these, else we were getting a 403 when the new njord action ran. Does this look right?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -882,14 +882,14 @@ Bundle-DocURL: https://adobe-consulting-services.github.io/acs-aem-commons/
     </profiles>
 
     <distributionManagement>
-        <repository>
-            <id>sonatype-central-portal</id>
-            <url>https://repo.maven.apache.org/maven2</url>
-        </repository>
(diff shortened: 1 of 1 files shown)


kwin:
I see now this is failing with GHA: https://github.com/Adobe-Consulting-Services/acs-aem-commons/actions/runs/18684145217/job/53272175613.
Please make sure you enable the namespace for SNAPSHOT deployments (https://central.sonatype.org/publish/publish-portal-snapshots/#enabling-snapshot-releases-for-your-namespace)

kwin:
Sorry, my bad, should be fixed with https://github.com/Adobe-Consulting-Services/acs-aem-commons/commit/6d40bba1ab2ef1aa99f452214bb308459e408ab9. Otherwise the SNAPSHOT deployment is done directly via maven-deploy-plugin which takes the endpoint from the POM's distributionMgmt (which is only suitable for download nor for upload).

PR #3707, opened 2026-02-21
Make SharedComponentPropertiesBindingsValuesProvider opt-in (disabled…

srma4tech:
… by default)

## Summary

This change makes `SharedComponentPropertiesBindingsValuesProvider` opt-in via an OSGi configuration property.

By default, the provider is now disabled (`enabled=false`) and must be explicitly enabled via OSGi configuration.

## Motivation

Currently, the provider is always active once the bundle is installed, which may introduce shared/global property bindings even when not explicitly desired.

Making it opt-in provides safer default behavior and more explicit control over script bindings.

## Changes

- Added SCR boolean property `enabled` (default = false).
- Guarded shared/global binding logic behind the `enabled` flag.
- Ensured `setDefaultBindings()` is always executed to preserve backward compatibility.
- Updated unit tests to validate both enabled and disabled scenarios.

## Backward Compatibility

- Default bindings remain available when disabled.
- When `enabled=true`, behavior matches the previous implementation.
- No public APIs were modified.
- No regression observed in HTL rendering during local validation.

## Validation

Validated locally using AEM SDK:

- Bundle active and SCR metadata generated.
- OSGi config visible in `/system/console/configMgr`.
- Disabled state preserves default bindings.
- Enabled state does not introduce errors.
- No errors observed in `error.log`.


kwin:
@srma4tech Thanks for the contribution. Would it be possible to convert from SCR to OSGi DS/Metatype annotations? Also for the next minor version we should probably leave it enabled by default. With the next major version upgrade we can toggle the default to become `false`.

srma4tech:
Hi @kwin 

Thank you for the feedback.

Yes, I can convert the component from SCR annotations to OSGi DS/Metatype annotations.

Regarding the default behavior — that makes sense. I will keep `enabled=true` for the next minor version to preserve backward compatibility and document that it can be toggled. We can then change the default to `false` in the next major release.

I’ll update the PR accordingly.

srma4tech:
Hi @kwin, Thank you for the feedback.

I have now converted SharedComponentPropertiesBindingsValuesProvider from Felix SCR annotations to OSGi DS (R6) annotations using @Component, @ObjectClassDefinition, and @Designate.

Key updates:

Replaced legacy SCR annotations with OSGi DS annotations

Introduced a typed Config interface for metatype configuration

Retained enabled=true as the default to preserve backward compatibility for this minor release

Updated unit tests to support DS activation and private reference injection

Verified local build (mvn clean install) and successful deployment

Please let me know if any further adjustments are required.

kwin:
@srma4tech Build fails with

```
Error:    SharedComponentPropertiesBindingsValuesProviderTest » UnnecessaryStubbing 
Unnecessary stubbings detected in test class: SharedComponentPropertiesBindingsValuesProviderTest
Clean & maintainable test code requires zero unnecessary code.
Following stubbings are unnecessary (click to navigate to relevant line of code):
  1. -> at com.adobe.acs.commons.wcm.properties.shared.impl.SharedComponentPropertiesBindingsValuesProviderTest.setUp(SharedComponentPropertiesBindingsValuesProviderTest.java:71)
  2. -> at com.adobe.acs.commons.wcm.properties.shared.impl.SharedComponentPropertiesBindingsValuesProviderTest.setUp(SharedComponentPropertiesBindingsValuesProviderTest.java:72)
  3. -> at com.adobe.acs.commons.wcm.properties.shared.impl.SharedComponentPropertiesBindingsValuesProviderTest.setUp(SharedComponentPropertiesBindingsValuesProviderTest.java:73)
  4. -> at com.adobe.acs.commons.wcm.properties.shared.impl.SharedComponentPropertiesBindingsValuesProviderTest.setUp(SharedComponentPropertiesBindingsValuesProviderTest.java:79)
  5. -> at com.adobe.acs.commons.wcm.properties.shared.impl.SharedComponentPropertiesBindingsValuesProviderTest.setUp(SharedComponentPropertiesBindingsValuesProviderTest.java:80)
Please remove unnecessary stubbings or use 'lenient' strictness. More info: javadoc for UnnecessaryStubbingException class.
```

E.g. at https://github.com/Adobe-Consulting-Services/acs-aem-commons/actions/runs/27087844607/job/79945681128.

kwin (changes requested):
Build is failing

srma4tech:
@kwin, Thanks for the review!

I addressed the test issues:

Removed the unnecessary Mockito stubbings by moving setup into the individual tests where needed.
Added coverage for the merged properties binding.
Removed unnecessary throws Exception declarations.
Verified the tests pass locally (mvn test) without UnnecessaryStubbingException.

The updated changes have been pushed.

kwin (approved):

srma4tech:
@kwin Thanks for reviewing and approving the changes! I appreciate your feedback throughout the process.

Issue #3701, opened 2026-01-30
AEM as Cloud - Error Publishing Redirect Configuration in acs-aem-commons 6.15.0


rahulbpatel:
Version : 6.15.0
AEM : AEM as Cloud
Publishing redirects –javax.jcr.AccessDeniedException detected in the author logs.

This used to work but all of a sudden has stopped work for our users.  Not sure what permissions are required to allow this.  

User has `jcr:read, crx:replicate, rep:write` permissions on `conf/gsi/settings/redirects` 

```
30.01.2026 22:23:21.096 [...] *ERROR* [... [1769811801091] POST /conf/gsi/settings/redirects.replicate.json HTTP/1.1] com.adobe.acs.commons.redirects.servlets.ReplicateRedirectMapServlet failed to replicate redirects
javax.jcr.AccessDeniedException: Access denied.
	at org.apache.jackrabbit.oak.jcr.security.AccessManager.checkPermissions(AccessManager.java:71) [org.apache.jackrabbit.oak-jcr:1.88.0.T20251029163355-5d5086d]
	at org.apache.jackrabbit.oak.jcr.session.NodeImpl$5.perform(NodeImpl.java:318) [org.apache.jackrabbit.oak-jcr:1.88.0.T20251029163355-5d5086d]
	at org.apache.jackrabbit.oak.jcr.session.NodeImpl$5.perform(NodeImpl.java:287) [org.apache.jackrabbit.oak-jcr:1.88.0.T20251029163355-5d5086d]
	at org.apache.jackrabbit.oak.jcr.delegate.SessionDelegate.perform(SessionDelegate.java:236) [org.apache.jackrabbit.oak-jcr:1.88.0.T20251029163355-5d5086d]
	at org.apache.jackrabbit.oak.jcr.session.ItemImpl.perform(ItemImpl.java:113) [org.apache.jackrabbit.oak-jcr:1.88.0.T20251029163355-5d5086d]
	at org.apache.jackrabbit.oak.jcr.session.NodeImpl.addNode(NodeImpl.java:287) [org.apache.jackrabbit.oak-jcr:1.88.0.T20251029163355-5d5086d]
	at org.apache.jackrabbit.commons.JcrUtils.getOrCreateByPath(JcrUtils.java:1617) [org.apache.jackrabbit.jackrabbit-jcr-commons:2.22.2]
[...]

PR #3711, opened 2026-02-25
#3710 Use a dedicated service user/subservice for Redirects Package creation and replication

### LINKED ISSUE #3710: Use a dedicated service user/subservice for Redirects Package creation and replication
### Type 
Improvement

### Referenced issue
https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/3701

### Background
The new replication approach is a good optimization because it avoids replicating each redirect rule individually.
However, it is not appropriate to grant authors/content admins broad access to all packages. In practice, this caused issues because editor/content-admin responsibilities are intentionally limited, and they typically do not have access to /etc/packages.

### Problem
Current implementation can require UI users (authors/content admins) to have package-related permissions, which violates least-privilege and role separation.

**Temporary workaround**
Grant limited permissions for package creation/replication to unblock operations.
**Proposed long-term solution**
Implement package management via a dedicated system user + Sling subservice that performs package creation and replication on behalf of users:
- UI users trigger replication without direct /etc/packages access.
- Backend service performs package operations using service credentials.
- Keep user permissions scoped to redirect configuration only.

**Acceptance Criteria**
1. Authors/content admins can replicate redirect maps without direct permissions to /etc/packages.
2. Package creation/assembly/replication runs through a dedicated subservice and system user.
3. Service user permissions are minimal and documented (least privilege).
4. Existing redirect replication functionality remains unchanged from a user perspective.
5. Audit/logging clearly shows actions executed by the service user.
6. Tests cover success and failure paths for service resolver acquisition and replication flow.

koha26:
Fixes https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/3710

### Type
Improvement

### Referenced issue
https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/3701

### Background
The new replication approach is a good optimization because it avoids replicating each redirect rule individually.
However, it is not appropriate to grant authors/content admins broad access to all packages. In practice, this caused issues because editor/content-admin responsibilities are intentionally limited, and they typically do not have access to /etc/packages.

### Problem
Current implementation can require UI users (authors/content admins) to have package-related permissions, which violates least-privilege and role separation.

**Temporary workaround**
Grant limited permissions for package creation/replication to unblock operations.
**Proposed long-term solution**
Implement package management via a dedicated system user + Sling subservice that performs package creation and replication on behalf of users:

- UI users trigger replication without direct /etc/packages access.
- Backend service performs package operations using service credentials.
- Keep user permissions scoped to redirect configuration only.

Acceptance Criteria

1. Authors/content admins can replicate redirect maps without direct permissions to /etc/packages.
2. Package creation/assembly/replication runs through a dedicated subservice and system user.
3. Service user permissions are minimal and documented (least privilege).
4. Existing redirect replication functionality remains unchanged from a user perspective.
5. Audit/logging clearly shows actions executed by the service user.
6. Tests cover success and failure paths for service resolver acquisition and replication flow.


YegorKozlov (changes requested):
Hi @koha26 
there are build errors. The baseline is asking to bump up the POM version. We can do that, but IMO the easiest fix would be to declare the reference in the old-school way:

```
    @Reference
    private transient Replicator replicator;
    @Reference
    private transient PackageHelper packageHelper;
    @Reference
    private transient Packaging packaging;
    @Reference
    private transient ResourceResolverFactory resourceResolverFactory;
```

Also, move to the repoinit changes to config.author. We only need it on Author.

ui.config/src/main/content/jcr_root/apps/acs-commons/config.author/org.apache.sling.jcr.repoinit.RepositoryInitializer-acs-commons-author.config
```diff
@@ -77,5 +77,13 @@ end
 
 # Copy Publish URL - AEM Sites Page Editor overlay
 create path (nt:unstructured) /apps/wcm(nt:folder)/core(nt:folder)/content(sling:Folder)/editor(cq:Page)/jcr:content/content/items(sling:OrderedFolder)/content(sling:OrderedFolder)/header/items/headerbar/items/pageinfopopover/items/list/items
+
+# Redirects Package Replicator
+create path /etc/packages(sling:Folder)
```
YegorKozlov: Replication via service user fails on cold start, when the  '/etc/packages/com.adobe.acs.commons.redirects` node does not exist.
The easiest fix is to pre-create it in repo-init
```
create path /etc/packages(sling:Folder)/com.adobe.acs.commons.redirects(sling:Folder)
```

YegorKozlov (changes requested):
@koha26 almost there. replication fails if `/etc/packages/com.adobe.acs.commons.redirects` does not exist. I think the easiest fix would be to pre-create it in the repoinit script .

ui.config/src/main/content/jcr_root/apps/acs-commons/config.author/org.apache.sling.jcr.repoinit.RepositoryInitializer-acs-commons-author.config
koha26: I tested it with an already existing folder. Didn't consider cold start. Fixed and pushed. Thank you.

YegorKozlov (approved):

YegorKozlov:
I tested on a vanilla 2026.2.24464.20260214T050318Z-260100 and 6.5.1.LTS and replication worked as expected.

koha26:
Thank you, @YegorKozlov !

YegorKozlov:
@davidjgonzalez can we please include this one?

kwin:
@koha26 Thanks for the PR, I am just wondering how to explicitly restrict users via ACLs to be able to manage redirects only (on the author) without giving permissions to replicate. Would be good to also have a check on a certain node for `crx:replicate` privileges with the current user before going ahead with the service user replication in the servlet.

koha26:
> @koha26 Thanks for the PR, I am just wondering how to explicitly restrict users via ACLs to be able to manage redirects only (on the author) without giving permissions to replicate. Would be good to also have a check on a certain node for `crx:replicate` privileges with the current user before going ahead with the service user replication in the servlet.

Hi @kwin, sorry, I missed your comment.

Your point is valid. I see the following solution:

1. Check the `crx:replicate` permission for the current user before passing the action over to the service user. Since we manage redirects under a specific path, such as `/conf/global/settings/redirects` or a project-specific path, we only need to check the ACE for that path. If the user does not have the required permissions, I will return `403`. I am going to handle this as part of the current issue.

2. Additionally, I am going to hide the **Publish Redirect Configurations** button on the **Manage Redirects** screen if the current user does not have the relevant permissions. This will prevent unauthorized publication.

3. I also see room for improvement for other action buttons on the screen, such as **+ Redirect Configuration**, **Import Redirect Map**, and **Save**. I am going to create a separate improvement issue to display these buttons only if the user has write permissions.

YegorKozlov:
@koha26 CRX package is only a technical means to copy the rules to publish. The ReplicateRedirectMapServlet servlet should check permissions before creating/replicating the package, e.g.

```java
@Override
    protected void doPost(SlingHttpServletRequest slingRequest, SlingHttpServletResponse slingResponse)
            throws ServletException, IOException {
  String path = slingRequest.getResource().getPath();
  if(!canReplicate(path, slingRequest.getResourceResolver()){
     throw new AccessDeniedException("no permissions to replicate" + path);
 }
}
```
this way the ACLs on the redirect root define what the user can do: CRUD + Replicate

As to #2 and #3, I'd lean towards a sensible error message instead of hiding the UI elements. There are subtle differences in the Granite UI in AEMaaCS and 6.5 and making it work everywhere might be challenging.

kwin:
@davidjgonzalez Any reason why you merged without my comment on this PR being solved?

kwin:
@davidjgonzalez Any reason why you merged without my comment on this PR being solved?

kwin:
@davidjgonzalez Any reason why you merged without my comment on this PR being solved?

Issue #3360, opened 2024-06-13
Page Root Provider Fallback for Experience Fragments


HitmanInWis:
The original `PageRootProvider` was written before Experience Fragments were prominent.  When the `PageRootProvider` attempts to get a root page for an Experience Fragment, it cannot find it.  This leads to the following downsides:

- Shared Component Properties render from the XF folder root rather than the rendered site.  They are also stored there when authoring XF.  This creates duplicate, potentially differing values from component instances in the site page tree and component instances within XF.
- Contextual Pathbrowser (from ACS AEM Commons) cannot be effectively used in components with XF, as the page root cannot be determined.
- Any other site functionality that wants to get the site homepage is unable to do so reliably with PageRootProvider

We should update the current `PageRootProvider` to check for a `fallbackPageRoot` property in the resource inherited properties.  This would allow a property of `fallbackPageRoot="/content/mysite/en"` in the folder root of XF, which would then cause all uses of `PageRootProvider` to return the website homepage.

Since this would only work in cases where the XF folder defines the property, it would be backward compatible with legacy implementations.  It would just give a new option for people that want their site to store/use Shared Component Properties and get the "homepage" using `PageRootProvider` from their site directly.

PR #3715, opened 2026-03-05
Page Root Provider Fallbacks

HitmanInWis:
This PR addresses [Issue](https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/3360) and other scenarios that need a fallback such as Launch content trees and viewing a historical version of a page/XF.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderConfig.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderConfig.java
@@ -63,15 +63,21 @@ public class PageRootProviderConfig {
 
     @Property(
             label = "Experience Fragment Root Path Method",
-            description = "Method to determine the root path for experience fragments. If not set, the page root path method will be used. Supported values are: [ site ]",
+            description = "Method to determine the root path for experience fragments. If not set (default), the page root path method will be used. Supported values are: [ site ].",
             value = { "" })
     static final String XF_ROOT_PATH_METHOD = "xf.root.path.method";
 
     @Property(
             label = "History Viewer Fallback",
-            description = "Enable this feature to use the corresponding live path when determining the page root. Note that for values reliant on page root (e.g. Shared Component Properties) the history viewer will reflect current values rather than historical values.",
+            description = "Enable this feature to use the corresponding live path when determining the page root. Note that for values reliant on page root (e.g. Shared Component Properties) the history viewer will reflect current values rather than historical values. Default false.",
             boolValue = { false })
     static final String HISTORY_VIEWER_FALLBACK = "history.viewer.fallback";
+
+    @Property(
+            label = "Content Launch Fallback",
+            description = "Enable this feature to have a content launch use the corresponding live path when determining the page root. This is generally desirable, as values reliant on page root (e.g. Shared Component Properties) will reflect the values that will apply when this page is promoted (except in the rare case where the launch also includes the page root and that page root is also being promoted with new values). Default false.",
+            boolValue = { false })
+    static final String LAUNCH_FALLBACK = "launch.fallback";
 
     private static final Logger log = LoggerFactory.getLogger(PageRootProviderConfig.class);
 
@@ -80,6 +86,7 @@ public class PageRootProviderConfig {
     private String xfRootPathMethod = null;
 
     private boolean historyViewerFallback = false;
+    private boolean launchFallback = false;
 
     /**
      * Retrieves the configured patterns.
@@ -110,6 +117,15 @@ public class PageRootProviderConfig {
         return this.historyViewerFallback;
     }
 
+    /**
+     * Get launch fallback setting
+     *
+     * @return true if launches should get root from current/live site paths
+     */
+    public boolean getLaunchFallback() {
+        return this.launchFallback;
+    }
+
     @Activate
     protected void activate(Map<String, Object> props) {
         List<Pattern> patterns = new ArrayList<Pattern>();
@@ -128,6 +144,7 @@ public class PageRootProviderConfig {
         this.pageRootPatterns = Collections.unmodifiableList(patterns);
         this.xfRootPathMethod = (String) props.get(XF_ROOT_PATH_METHOD);
         this.historyViewerFallback = PropertiesUtil.toBoolean(props.get(HISTORY_VIEWER_FALLBACK), false);
+        this.launchFallback = PropertiesUtil.toBoolean(props.get(LAUNCH_FALLBACK), false);
     }
 
     @Deactivate
--- a/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderConfig.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderConfig.java
@@ -61,9 +61,17 @@ public class PageRootProviderConfig {
     /* Page root property. */
     static final String PAGE_ROOT_PATH = "page.root.path";
 
+    @Property(
+            label = "Experience Fragment Root Path Method",
+            description = "Method to determine the root path for experience fragments. If not set, the page root path method will be used. Supported values are: [ site ]",
+            value = { "" })
+    static final String XF_ROOT_PATH_METHOD = "xf.root.path.method";
+
     private static final Logger log = LoggerFactory.getLogger(PageRootProviderConfig.class);
 
     private List<Pattern> pageRootPatterns = null;
+
+    private String xfRootPathMethod = null;
 
     /**
      * Retrieves the configured patterns.
@@ -74,6 +82,15 @@ public class PageRootProviderConfig {
         return Optional.ofNullable(this.pageRootPatterns)
                 .map(Collections::unmodifiableList)
                 .orElse(null);
+    }
+
+    /**
+     * Get XF root path method
+     *
+     * @return XF root path method
+     */
+    public String getXfRootPathMethod() {
+        return this.xfRootPathMethod;
     }
 
     @Activate
@@ -92,6 +109,7 @@ public class PageRootProviderConfig {
         }
 
         this.pageRootPatterns = Collections.unmodifiableList(patterns);
+        this.xfRootPathMethod = (String) props.get(XF_ROOT_PATH_METHOD);
     }
 
     @Deactivate
--- a/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderConfig.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderConfig.java
@@ -67,11 +67,19 @@ public class PageRootProviderConfig {
             value = { "" })
     static final String XF_ROOT_PATH_METHOD = "xf.root.path.method";
 
+    @Property(
+            label = "History Viewer Fallback",
+            description = "Enable this feature to use the corresponding live path when determining the page root. Note that for values reliant on page root (e.g. Shared Component Properties) the history viewer will reflect current values rather than historical values.",
+            boolValue = { false })
+    static final String HISTORY_VIEWER_FALLBACK = "history.viewer.fallback";
+
     private static final Logger log = LoggerFactory.getLogger(PageRootProviderConfig.class);
 
     private List<Pattern> pageRootPatterns = null;
 
     private String xfRootPathMethod = null;
+
+    private boolean historyViewerFallback = false;
 
     /**
      * Retrieves the configured patterns.
@@ -93,6 +101,15 @@ public class PageRootProviderConfig {
         return this.xfRootPathMethod;
     }
 
+    /**
+     * Get history viewer fallback setting
+     *
+     * @return true if history viewer should get root from current/live site/XF paths
+     */
+    public boolean getHistoryViewerFallback() {
+        return this.historyViewerFallback;
+    }
+
     @Activate
     protected void activate(Map<String, Object> props) {
         List<Pattern> patterns = new ArrayList<Pattern>();
@@ -110,6 +127,7 @@ public class PageRootProviderConfig {
 
         this.pageRootPatterns = Collections.unmodifiableList(patterns);
         this.xfRootPathMethod = (String) props.get(XF_ROOT_PATH_METHOD);
+        this.historyViewerFallback = PropertiesUtil.toBoolean(props.get(HISTORY_VIEWER_FALLBACK), false);
     }
 
     @Deactivate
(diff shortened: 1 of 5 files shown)


kwin:
@HitmanInWis Would it be possible to convert to OSGi annotations in PageRootProvider at the same time? We are trying to get rid of legacy SCR annotations slowly....

HitmanInWis:
Sure thing, done!

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderConfig.java
```diff
@@ -20,50 +20,41 @@
...
-)
-@Service(PageRootProviderConfig.class)
 /**
  * Configuration instance for Page Root Provider.
- * Use service.ranking to guarantee priority between conflicting configurations.
+ * Use `service.ranking` to guarantee priority between conflicting configurations.
  *
  * @see PageRootProviderMultiImpl
  */
+@Component(
```
santhanbattina, line 42: Nice to see the move from the old Felix SCR annotations to the OSGi DS annotation. This aligns better with modern AEM practices.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderConfig.java
```diff
@@ -105,4 +126,36 @@ protected void deactivate() {
         }
     }
 
+    @ObjectClassDefinition(
```
santhanbattina, line 129: I like the move to a typed Config interface here. It makes the configuration clearer compared to the old @Property approach.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderConfig.java
```diff
@@ -76,10 +67,37 @@ public List<Pattern> getPageRootPatterns() {
...
+     */
+    public boolean getLaunchFallback() {
+        return this.launchFallback;
+    }
+
     @Activate
-    protected void activate(Map<String, Object> props) {
-        List<Pattern> patterns = new ArrayList<Pattern>();
-        String[] regexes = PropertiesUtil.toStringArray(props.get(PAGE_ROOT_PATH), new String[] { DEFAULT_PAGE_ROOT_PATH });
+    protected void activate(Config config) {
```
santhanbattina, line 98: Switching the activate method to use the typed config instead of the properties map looks much cleaner and easier to follow.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderConfig.java
```diff
@@ -105,4 +126,36 @@ protected void deactivate() {
...
+                name = "Root page path pattern",
+                description = "Regex(es) used to select the root page root path. Regex must contain at least one group (with index 1) which is used as page root. It is matched against the given path. Evaluates list top-down; first match wins. Defaults to [ " + DEFAULT_PAGE_ROOT_PATH + " ]"
+        )
+        String[] page_root_path() default { DEFAULT_PAGE_ROOT_PATH };
+
+        @AttributeDefinition(
+                name = "Experience Fragment Root Path Method",
+                description = "Method to determine the root path for experience fragments. If not set (default), the page root path method will be used. Supported values are: [ site ]."
+        )
+        String xf_root_path_method() default "";
```
santhanbattina, line 144: The xf_root_path_method option is interesting. Should we validate supported values here (like "site") to avoid unexpected config values?

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderConfig.java
```diff
@@ -105,4 +126,36 @@ protected void deactivate() {
...
+
+        @AttributeDefinition(
+                name = "History Viewer Fallback",
+                description = "Enable this feature to use the corresponding live path when determining the page root. Note that for values reliant on page root (e.g. Shared Component Properties) the history viewer will reflect current values rather than historical values. Default false."
+        )
+        boolean history_viewer_fallback() default false;
```
santhanbattina, line 150: The history viewer fallback option looks useful. Just wondering if there are any edge cases where it might conflict with the existing page root pattern logic.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderConfig.java
```diff
@@ -105,4 +126,36 @@ protected void deactivate() {
...
+
+        @AttributeDefinition(
+                name = "Content Launch Fallback",
+                description = "Enable this feature to have a content launch use the corresponding live path when determining the page root. This is generally desirable, as values reliant on page root (e.g. Shared Component Properties) will reflect the values that will apply when this page is promoted (except in the rare case where the launch also includes the page root and that page root is also being promoted with new values). Default false."
+        )
+        boolean launch_fallback() default false;
```
santhanbattina, line 156: The launch fallback flag makes sense. The description helps explain when it would be useful

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderMultiImpl.java
```diff
@@ -38,20 +35,21 @@
...
 
-@Component(metatype=false)
-@Service(PageRootProvider.class)
 /**
  * Service to fetch the site root page (i.e. home page) for a given resource.
  * Supports multiple (independent) configurations.
  *
  * @see PageRootProviderConfig
  */
+@Component(service = PageRootProvider.class)
```
santhanbattina, line 44: Nice to see the move from the Felix SCR annotations to the OSGi DS annotations here. This keeps things aligned with newer AEM practices.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderMultiImpl.java
```diff
@@ -38,20 +35,21 @@
...
 public class PageRootProviderMultiImpl implements PageRootProvider {
 
-    private static final Logger log = LoggerFactory.getLogger(PageRootProviderMultiImpl.class);
+    private static final Pattern VERSION_HISTORY_PATTERN = Pattern.compile("/tmp/versionhistory/[0-9a-f]+/[0-9a-f-]+/(.*)");
+    private static final Pattern XF_PATH_PATTERN = Pattern.compile("/content/experience-fragments/(.*)");
+    private static final Pattern LAUNCH_PATH_PATTERN = Pattern.compile("/content/launches/.*?/content/(.*)");
+    private static final Logger LOG = LoggerFactory.getLogger(PageRootProviderMultiImpl.class);
```
santhanbattina, line 50: Good call standardizing the logger name to LOG. That matches the usual convention for static loggers.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderMultiImpl.java
```diff
@@ -38,20 +35,21 @@
...
  * Service to fetch the site root page (i.e. home page) for a given resource.
  * Supports multiple (independent) configurations.
  *
  * @see PageRootProviderConfig
  */
+@Component(service = PageRootProvider.class)
 public class PageRootProviderMultiImpl implements PageRootProvider {
 
-    private static final Logger log = LoggerFactory.getLogger(PageRootProviderMultiImpl.class);
+    private static final Pattern VERSION_HISTORY_PATTERN = Pattern.compile("/tmp/versionhistory/[0-9a-f]+/[0-9a-f-]+/(.*)");
```
santhanbattina, line 47: Defining the regex patterns as constants here makes the logic easier to follow.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderMultiImpl.java
```diff
@@ -76,21 +74,39 @@ public Page getRootPage(Resource resource) {
     @Override
     public String getRootPagePath(String resourcePath) {
         for (PageRootProviderConfig config : this.configList) {
+            String pathToSearch = resourcePath;
+
+            // If page/XF history viewer should use the corresponding live content tree to determine the root...
+            if (config.getHistoryViewerFallback()) {
+                pathToSearch = VERSION_HISTORY_PATTERN.matcher(pathToSearch).replaceFirst("/content/$1");
```
santhanbattina, line 81: I noticed the "/content/$1" replacement string is used in a few places here. Maybe it would be cleaner to extract it as a constant instead of repeating the literal.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderMultiImpl.java
```diff
@@ -76,21 +74,39 @@ public Page getRootPage(Resource resource) {
     @Override
     public String getRootPagePath(String resourcePath) {
         for (PageRootProviderConfig config : this.configList) {
+            String pathToSearch = resourcePath;
+
+            // If page/XF history viewer should use the corresponding live content tree to determine the root...
+            if (config.getHistoryViewerFallback()) {
```
santhanbattina, line 80: The history viewer fallback logic looks helpful. Just wondering if this could change behavior for existing configs that rely on the original resource path.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderMultiImpl.java
```diff
@@ -76,21 +74,39 @@ public Page getRootPage(Resource resource) {
...
+                pathToSearch = VERSION_HISTORY_PATTERN.matcher(pathToSearch).replaceFirst("/content/$1");
+            }
+
+            // If launch content should use the corresponding live content tree to determine the root...
+            if (config.getLaunchFallback()) {
```
santhanbattina, line 85: The launch fallback option makes sense here. Nice addition to support launch content paths.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderMultiImpl.java
```diff
@@ -76,21 +74,39 @@ public Page getRootPage(Resource resource) {
...
+                pathToSearch = LAUNCH_PATH_PATTERN.matcher(pathToSearch).replaceFirst("/content/$1");
+            }
+
+            // If XF should use the corresponding site content tree to determine the root...
+            if ("site".equals(config.getXfRootPathMethod())) {
```
santhanbattina, line 90: Using "site" as the XF root path method makes sense. Should we validate allowed values somewhere to avoid unexpected configuration values?

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderMultiImpl.java
```diff
@@ -76,21 +74,39 @@ public Page getRootPage(Resource resource) {
...
                 }
             }
         }
 
-        log.debug("Resource path does not include the configured page root path.");
+        LOG.debug("Resource path does not include the configured page root path.");
         return null;
     }
 
+    @Reference(name = "config", service = PageRootProviderConfig.class, cardinality = ReferenceCardinality.AT_LEAST_ONE, policy = ReferencePolicy.DYNAMIC)
```
santhanbattina, line 109: Switching to the DS @Reference style looks cleaner than the previous reference configuration.

santhanbattina (commented):
Overall this looks like a nice improvement. The migration to OSGi DS annotations simplifies the implementation, and the new fallback options for history viewer, launches, and XF paths add useful flexibility when determining the page root.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderConfig.java
HitmanInWis: I didnt bother with validation since it will otherwise fall back to the default method - having the service simply not start on a bad value would be another option, and I can change it to do that if required, just didn't seem like a big value add.

Only reason I made this one a string versus checkbox like the others is b/c I could see potentially alternate methods - e.g. "fallback" where it first checks if the XF has a root of its own, then uses the "site" method if not.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderConfig.java
HitmanInWis: Could be.  But the option is opt-in only so buyer beware :).

As of now SCP dont work on history viewer at all, so this is at least a step in the right direction IMO.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderMultiImpl.java
HitmanInWis: closing this comment as duplicate of the other that asks the same

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderMultiImpl.java
HitmanInWis: similar comment as the other above, regarding the history viewer fallback.  Definitely could be some implementations where the Page Root Provider is used for features other than Shared Component Properties, and in those cases it might not make sense for those implementations to opt-in if the new logic breaks them.  Hence why this option is off by default.

bundle/src/main/java/com/adobe/acs/commons/wcm/impl/PageRootProviderMultiImpl.java
HitmanInWis: I saw that recommendation from the automated review, but I sorta fundamentally disagree with it. There are 3 fundamentally different regexps that are using it as a replacement string, and the fact that they can all use exactly the same replacement slug feels more fortuitous than a reliable pattern.  Given there's no performance benefit in modern java to consolidate these strings, it feels like each using the literal value in its own use case is appropriate.

That said, if others agree this should be refactored to a constant, I can certainly do so.

HitmanInWis:
@kwin is this good to go?  or is it waiting on anything from your perspective?

kwin (approved):
LGTM, thanks.

HitmanInWis:
@kwin I've tested this myself, and also updated unit tests, but do we think anyone else needs to test before merging?  Or is this good for me to hit he merge button?

PR #3729, opened 2026-04-10
#3696 improve handling rules with extensions

### LINKED ISSUE #3696: Manage Redirects - extension ignored
ACS Commons v6.11.0

We are using Manage Redirects to create redirects in AEM and sync to Dispatcher.

If we configure a redirect of /na.html -> /na/en.html in Manage Redirects, AEM will do the following:
- Redirect /na.sitemap-index.xml -> /na/en.html.xml
- Redirect /na.sitemap.en-sitemap.xml -> /na/en.html.xml

Note that the *dispatcher* does not have this bug, only AEM.  When the request gets to AEM it sees that the URL (unshortened) refers to `/content/mysite/na` and even though the extension is `xml` and the redirect was configured with an extension of `html` the redirect is applied when it shouldn't be.  This makes our sitemaps inaccessible, and thus we needed to remove these redirects.

A potential workaround is to use the checkbox `Evaluate request URI` but this option is off by default (authors dont often select it) and it seems that extension should be honored regardless if specified.

YegorKozlov:
fixes #3696

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectConfiguration.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectConfiguration.java
@@ -159,10 +159,16 @@ public class RedirectConfiguration {
         String normalizedPath = normalizePath(resourcePath);
         RedirectMatch match = null;
         RedirectRule rule = getPathRule(normalizedPath, contextPrefix);
+        if (rule != null && !isExtensionCompatible(rule, request)) {
+            rule = null;
+        }
         if(rule == null && hasNonRegexRequestURIRules()){
             // there are request URI rules. Check is any mathes
             String pathToEvaluate = determinePathToEvaluate(normalizedPath, true, request);
             rule = getPathRule(pathToEvaluate, contextPrefix);
+            if (rule != null && !isExtensionCompatible(rule, request)) {
+                rule = null;
+            }
         }
         if (rule != null) {
             match = new RedirectMatch(rule, null);
@@ -240,6 +246,32 @@ public class RedirectConfiguration {
 
     private boolean hasNonRegexRequestURIRules() {
         return this.nonRegexRequestURIRules;
+    }
+
+    /**
+     * Verify that a path rule whose source carries an explicit extension (e.g. {@code /foo.html})
+     * only matches requests whose extension is compatible with the rule's source extension.
+     * Rules authored without an extension stay permissive and match any request extension
+     * (preserving the current default behaviour). A request without an explicit extension is
+     * also treated as compatible (AEM serves {@code /foo} as html by default).
+     *
+     * @param rule    the candidate path rule
+     * @param request the current sling request, may be {@code null} for non-request callers
+     * @return {@code true} if the rule's source extension is absent or compatible with the request extension
+     */
+    private static boolean isExtensionCompatible(RedirectRule rule, SlingHttpServletRequest request) {
+        String ruleExt = rule.getSourceExtension();
+        if (ruleExt == null) {
+            return true;
+        }
+        if (request == null) {
+            return true;
+        }
+        String requestExt = request.getRequestPathInfo().getExtension();
+        if (requestExt == null) {
+            return true;
+        }
+        return ruleExt.equalsIgnoreCase(requestExt);
     }
 
 }
--- a/bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectConfiguration.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectConfiguration.java
@@ -159,14 +159,14 @@ public class RedirectConfiguration {
         String normalizedPath = normalizePath(resourcePath);
         RedirectMatch match = null;
         RedirectRule rule = getPathRule(normalizedPath, contextPrefix);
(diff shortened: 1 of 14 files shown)


bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectConfiguration.java
```diff
@@ -242,4 +248,30 @@ private boolean hasNonRegexRequestURIRules() {
...
+     * only matches requests whose extension is compatible with the rule's source extension.
+     * Rules authored without an extension stay permissive and match any request extension
+     * (preserving the current default behaviour). A request without an explicit extension is
+     * also treated as compatible (AEM serves {@code /foo} as html by default).
+     *
+     * @param rule    the candidate path rule
+     * @param request the current sling request, may be {@code null} for non-request callers
+     * @return {@code true} if the rule's source extension is absent or compatible with the request extension
+     */
+    private static boolean isExtensionCompatible(RedirectRule rule, SlingHttpServletRequest request) {
```
kwin: I think the naming is confusing, what about `matchesExtension(....)`

bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectConfiguration.java
```diff
@@ -242,4 +248,30 @@ private boolean hasNonRegexRequestURIRules() {
...
+        String ruleExt = rule.getSourceExtension();
```
kwin, line 263: What about rules which should only match requests not having an extension at all?

kwin (changes requested):
I think adding more special logic for redirect rules with extensions makes it even more complex and hard to debug. What about selectors then being contained in rules?

I think the redirect should really be applied on top of the request url (before any Sling URL decomposition), otherwise you observe different behaviour depending on whether the source resource does exist or not and if you export to dispatcher or not.

bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectConfiguration.java
YegorKozlov: Extensionless source rule is permissive - matches any extension or no extension, e.g.
`/content/mysite/na  =>  /content/mysite/na/en`  will match `html`, `xml` and `null`

but if the source has extension then this PR enforces it's followed.

bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectConfiguration.java
YegorKozlov: ok

YegorKozlov:
> I think adding more special logic for redirect rules with extensions makes it even more complex and hard to debug. What about selectors then being contained in rules?
> 
> I think the redirect should really be applied on top of the request url (before any Sling URL decomposition), otherwise you observe different behaviour depending on whether the source resource does exist or not and if you export to dispatcher or not.

We have to balance between keeping the tool easy to use on one hand, and ability to handle edge cases on the other. When I designed it I did wanted to use Sling URL decomposition and evaluate on resource path only. All the features that made it so complicated came later. 

There is the Evaluate Request URI mode which passes the full URI including path, query and fragment (`request.getRequestURI()`). It's powerful, but confusing to plain content authors who don't know or don't care about Sling URL decomposition. 

The idea to evaluate the raw request URI instead of RequestPathInfo is lucrative, and I like it. But it's quite a maneuver I must say.

bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectConfiguration.java
```diff
@@ -242,4 +248,30 @@ private boolean hasNonRegexRequestURIRules() {
...
+     * only matches requests whose extension is compatible with the rule's source extension.
+     * Rules authored without an extension stay permissive and match any request extension
+     * (preserving the current default behaviour). A request without an explicit extension is
+     * also treated as compatible (AEM serves {@code /foo} as html by default).
+     *
+     * @param rule    the candidate path rule
+     * @param request the current sling request, may be {@code null} for non-request callers
+     * @return {@code true} if the rule's source extension is absent or compatible with the request extension
+     */
+    private static boolean matchesExtension(RedirectRule rule, SlingHttpServletRequest request) {
```
kwin, line 262: Let us consider the source type introduced in #3730 and only use this fuzzy matching for the default type.

kwin:
@YegorKozlov Any progress with that? That is biting us quite hard in the context of redirecting extensionless requests to one with `.html` extension.

bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectConfiguration.java
kwin, line 262: What if I want to match a rule only to extensionless request paths?

kwin:
Some other test cases currently not working (next to extensionless redirect):

1. Source: `/content/mysite/.../test.txt` -> Target: `/content/mysite/.../test.html` (Evaluate Request URI set to `true`), redirects incorrectly to `/content/mysite/.../test.txt.html`
1. Source: `/content/mysite/.../test.html` -> Target: `/content/mysite/.../test.pdf` (Evaluate Request URI set to `true`), redirects incorrectly to `/content/mysite/.../test.pdf.html`

PR #3733, opened 2026-04-14
Also consider "ignoreContextPrefix" for source matching

### LINKED ISSUE #3732: RedirectManager: Evaluate "contextPrefixIgnored" for both source and target instead of just for target
Currently the option to ignore a context prefix for a specific rule via checkbox  in "Advanced -> Ignore Context Prefix" is only evaluated for the target path at https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/21ac1e6f1b77cc8e26007af5a6f63d406080c282/bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java#L666. However it should also be considered for the source path matching in 
https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/21ac1e6f1b77cc8e26007af5a6f63d406080c282/bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectConfiguration.java#L217 and https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/21ac1e6f1b77cc8e26007af5a6f63d406080c282/bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectConfiguration.java#L190.

The latter is particularly important due to issues outlined in #3663

kwin:
Previously it was only evaluated for calculating the path. This closes #3732


bundle/src/test/java/com/adobe/acs/commons/redirects/filter/RedirectFilterTest.java
```diff
@@ -1207,14 +1207,35 @@ public void testContextPrefixWithAbsoluteUrl() throws Exception {
     public void testIgnoredContextPrefix() throws Exception {
         withRules(
             new RedirectResourceBuilder(context)
-                    .setSource("/en/one")
+                    .setSource("/content/geometrixx/en/one")
```
kwin, line 1210: I had to modify this IT to explicitly use the context in `source`, otherwise it would no longer match due to the flag set in line 1213. Do you consider this change backwards-incompatible @YegorKozlov?

davidjgonzalez:
This looks like a dupe of #3745  - do we still need this?

Issue #3734, opened 2026-04-14
RedirectManager: Support regular expression in source without capturing groups


kwin:
Due to the condition inside https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/21ac1e6f1b77cc8e26007af5a6f63d406080c282/bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectRule.java#L300-L303 any regular expressions not containing capturing groups (i.e. not containing `(` and `)`) are not supported. This limitation is not documented and should be lifted.

For example I consider the source `/content/path/.*` totally valid and a capturing group is not necessary if everything matching that regex should redirect to a fix page (without leveraging backreferences).

Issue #3730, opened 2026-04-14
RedirectManager: Trailing "*" in source path being converted to "(.*)" might destroy regular expression


kwin:
The magic replacing a trailing `*` by `(.*)` in https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/21ac1e6f1b77cc8e26007af5a6f63d406080c282/bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectRule.java#L159-L161 might destroy regular expressions.

For example a source path like `/content/we-retail/something-[a-z]*` is converted to `/content/we-retail/something-[a-z](.*)` which changes the matching semantics incorrectly.

The path `/content/we-retail/something-aaa1` doesn't match the original source regex but incorrectly matches the modified one.

kwin:
Not sure how to fix this in a backwards compatible manner. Any idea @YegorKozlov?

YegorKozlov:
> Not sure how to fix this in a backwards compatible manner. Any idea [@YegorKozlov](https://github.com/YegorKozlov)?

I'd move it to https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/21ac1e6f1b77cc8e26007af5a6f63d406080c282/bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectRule.java#L292

it will try to convert the source to regex first and then try the `* => (.*)` replacement in the `'catch' block. This way the "true" regex wins, and this syntax sugar is used as a fallback only.

kwin:
Not sure this approach works, as most paths ending with just a star are valid regular expressions as well (with the star referring to the character before). For example look at https://regex101.com/?regex=%2Fmy%2Fpath%2F*&testString=&flags=gm&flavor=java&delimiter=%22.

YegorKozlov:
@kwin I see no other way but to introduce heuristics to distinguish the "glob wildcard" sugar from true regex quantifiers. Something like checking that a wildcard preceded by ] (end of character class) or ) (end of group) is unambiguously a regex quantifier. 

the tests to add:
| Source | Before fix | After fix |
|--------|-----------|-----------|
| `/content/en/*` | converted to `/content/en/(.*) ✅` | converted to `/content/en/(.*) ✅ (unchanged)` |
| `/content/[a-z]*` | wrongly → `/content/[a-z](.*)` | kept as-is, used as 0-group pattern rule ✅ |
[...]

PR #3743, opened 2026-04-24
#3730 Trailing "*" in source path being converted to "(.*)" might destroy regular expression

YegorKozlov:
The PR addresses #3730 and #3734

The redirect configuration now provides an option to set the match type:

image


bundle/src/main/java/com/adobe/acs/commons/redirects/models/RedirectRule.java
```diff
@@ -347,6 +360,31 @@ public String getPreserveQueryString() {
...
+    /**
+     * Compile src as a regex without requiring capturing groups.
+     * Used when matchType is {@link MatchType#REGEX}.
+     */
+    static Pattern toRegexForced(String src, boolean nc) {
+        try {
+            int flags = nc ? Pattern.CASE_INSENSITIVE : 0;
+            return Pattern.compile(src, flags);
+        } catch (PatternSyntaxException e) {
+            log.error("invalid regex: {}", src);
```
kwin, line 383: Shouldn't this somehow ending up in the UI visible for the user?

bundle/src/main/java/com/adobe/acs/commons/redirects/models/MatchType.java
```diff
@@ -0,0 +1,43 @@
...
+package com.adobe.acs.commons.redirects.models;
+
+/**
+ * Controls how the source path of a redirect rule is interpreted during matching.
+ */
+public enum MatchType {
+    /**
+     * Default behaviour: a trailing {@code *} is replaced with {@code (.*)},
+     * and the result is compiled as a regex only when it contains at least one
+     * capturing group; otherwise the source is treated as an exact path.
```
kwin, line 27:
```suggestion
     * capturing group; otherwise the source is treated as plain text.
```

kwin (approved):
Thanks looks good, however maybe error handling can be improved. Also for invalid paths used for plain text source matches.

kwin:
The source match type allows to be stronger with regards to extension handling (i.e. treating `.html`specially). I think this should only be the case for the default type (for backwards compatibility)

PR #3759, opened 2026-05-17
Redirect Manager: catch IllegalArgumentException on malformed request URLs

YegorKozlov:
FIxes #3735


CLAUDE.md
kwin, line 1: Let's add this in a separate PR/commit

bundle/src/main/java/com/adobe/acs/commons/redirects/filter/RedirectFilter.java
```diff
@@ -646,6 +646,9 @@ RedirectMatch match(SlingHttpServletRequest slingRequest) {
         } catch (ExecutionException e){
             log.error("failed to load redirect rules from {}", configPath, e);
             return null;
+        } catch (IllegalArgumentException e) {
+            log.debug("Skipping mapped URL match, invalid URI", e);
```
kwin: Doesn't this deserve a WARN?

kwin (changes requested):

davidjgonzalez (approved):

PR #3763, opened 2026-05-29
fix: URL Asset Importer did not use stored Credentials

Skiftstar:
The Credentials you input into the URL Asset Importer did not get applied when the URLs were being called.

I noticed this while trying to import Assets from one AEM Author to another. The Importer would just be created files with 9KB, which, if inspected, turned out to be HTML files showing the AEM "Resource not found" page.

This Pull request should fix it.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/FileOrRendition.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/FileOrRendition.java
@@ -32,16 +32,21 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLConnection;
-import java.nio.charset.StandardCharsets;
-import java.util.Base64;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.TreeMap;
 import java.util.stream.Stream;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.http.HttpHost;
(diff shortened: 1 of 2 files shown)


bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/FileOrRendition.java
```diff
@@ -246,6 +249,13 @@ private HttpResponse initiateDownload() throws IOException {
             if (connection == null) {
                 try {
                     lastRequest = new HttpGet(url);
+                    if (StringUtils.isNotBlank(clientProvider.getUsername())) {
```
kwin, line 258: Rather use https://stackoverflow.com/a/42842916 as manually passed authentication headers may otherwise end up on redirect targets (see https://hc.apache.org/security.html: 
> For this reason, you should use a CredentialsProvider instead of adding security headers manually. Please see the ClientAuthentication example)

kwin (changes requested):

bundle/src/main/java/com/adobe/acs/commons/mcp/impl/processes/asset/FileOrRendition.java
Skiftstar: thanks for the info! I adjusted it

kwin:
Looks good now. Can you add a changelog entry as well?

Skiftstar:
> Looks good now. Can you add a changelog entry as well?

Added

kwin (approved):

PR #3768, opened 2026-06-07
Update plugins to latest versions

kwin:


kwin:
Fails due to https://github.com/adobe/aemanalyser-maven-plugin/issues/354.

PR #3789, opened 2026-08-05
Implement logging for empty agent list in AgentIdsAgentFilter

### LINKED ISSUE #3788: [Enhancement] Clarify and log "allow all" fallback behavior for empty lists in AgentIdsAgentFilter
Currently, passing an empty list of agent IDs causes `isIncluded(Agent)` to return `true` (allowing all agents). While this prevents silent replication failures, it can be counterintuitive for developers expecting an empty allowlist to return false. I'd like to submit a PR to add explicit JavaDocs to the constructor and class level explaining this fallback behavior.

If a developer dynamically generates a list of target agents, and that list unexpectedly resolves to empty, the Replicator will execute a massive "publish all" event. Because the class lacks explicit JavaDoc and logging around this specific fallback, this behavior is difficult to anticipate and trace during debugging.

bkkothari2255:
Related Issue
Closes #3788

it improves the observability and developer experience of the `AgentIdsAgentFilter` when it is initialized with an empty or null list.

Previously, if an empty list was passed, the filter would silently return true (allowing replication to all agents). While this is the intended fallback to AEM's default behavior, it violates the principle of least astonishment for an allowlist and lacked documentation, risking accidental "publish all" events.

Changes Included
- Added explicit JavaDoc: Documented the constructor to clearly state that an empty/null list bypasses the filter and allows all agents.

- Added Constructor Logging: Introduced a `log.debug` statement in the constructor that triggers only when an empty list is provided.

Note: The log is intentionally placed in the constructor rather than isIncluded() to prevent log spam and disk I/O overhead during high-volume replication events.

- Added unit test cases in `AgentIdsAgentFilterTest` for empty and null agent ID list fallback scenarios.

How to Test
1. Instantiate AgentIdsAgentFilter with an empty ArrayList.
2. Verify that the debug log is emitted: "Initialized with an empty agent list. Default AEM behavior (allow all agents) will apply..."
3. Verify isIncluded() still returns true.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/replication/AgentIdsAgentFilter.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/replication/AgentIdsAgentFilter.java
@@ -38,6 +38,7 @@ import org.slf4j.LoggerFactory;
  */
 public class AgentIdsAgentFilter implements AgentFilter {
 
+    private static final Logger log = LoggerFactory.getLogger(AgentIdsAgentFilter.class);
     private final List<String> agentIds;
 
     public AgentIdsAgentFilter(List<String> agentIds) {
--- a/bundle/src/main/java/com/adobe/acs/commons/replication/AgentIdsAgentFilter.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/replication/AgentIdsAgentFilter.java
@@ -24,7 +24,18 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+/**
+ * An AEM {@link AgentFilter} implementation that filters replication agents 
+ * based on a specified list of Agent IDs.
+ * <p>
+ * <strong>Note on Empty Lists:</strong> If the provided list of agent IDs is 
+ * null or empty, {@link #isIncluded(Agent)} will return {@code true} for all agents. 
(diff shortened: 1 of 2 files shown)


bundle/src/main/java/com/adobe/acs/commons/replication/AgentIdsAgentFilter.java
```diff
@@ -33,6 +44,10 @@ public AgentIdsAgentFilter(List<String> agentIds) {
         this.agentIds = Optional.ofNullable(agentIds)
                 .map(list -> (List<String>) new ArrayList<>(list))
                 .orElse(Collections.emptyList());
+        
+        if (this.agentIds.isEmpty()) {
+            log.debug("Initialized with an empty agent list. Default AEM behavior (allow all agents) will apply to this replication event.");
```
kwin, line 50: There is no field with name `log` yet in this class.

kwin (changes requested):
Build is failing due to issue mentioned above. Please build locally first before marking the PR as ready.

bundle/src/main/java/com/adobe/acs/commons/replication/AgentIdsAgentFilter.java
bkkothari2255: @kwin Good catch, thanks!! I’ve added the missing log field, also went ahead and added unit tests for the null and empty list scenarios while I was at it. Ready for another look!

davidjgonzalez (approved):

PR #3793, opened 2026-08-27
Support asset URN references in Asset Packager filter resolution

markdaugherty:
## Summary
- The Asset Packager's filter resolution previously only handled plain DAM asset paths; references authored as `urn:aaid:aem:<uuid>` (AEM asset ID/URN references) were silently skipped.
- `AssetPackageUtil#addFilter` now detects the `urn:aaid:aem:<uuid>` pattern, resolves the underlying asset by its `jcr:uuid`, and adds the resolved path to the package filter set — falling back to the existing path-based resolution otherwise.

## Test plan
- [x] Added `AssetPackageUtilTest` with AemContext-backed content covering both plain-path and URN-based `fileReference` values, verifying the expected filter paths are produced.
- [x] `mvn -pl bundle test -Dtest=AssetPackageUtilTest`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/packaging/util/AssetPackageUtil.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/packaging/util/AssetPackageUtil.java
@@ -28,6 +28,12 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 
+import com.day.cq.dam.api.DamConstants;
+import com.day.cq.dam.commons.util.DamUtil;
+import com.day.cq.wcm.api.NameConstants;
+import com.day.cq.wcm.api.Page;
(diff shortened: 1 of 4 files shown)


YegorKozlov:
What's the use case of this PR? To my knowledge, `urn:aaid:aem:*` references are inserted by Content Advisor and typically point to a remote Asset Delivery repo, resolved via ASSET_DELIVERY_REPOSITORY_ID plus the asset reference, not a local node lookup. 
The PR assumes the asset actually exists on *this* instance. On an author that's a consumer rather than producer of the delivery repo, that lookup would just return null and the reference would be skipped anyway .

 Code looks good, just trying to understand the intention and what case this fixes.

davidjgonzalez:
@markdaugherty As @YegorKozlov notes - how does this get installed somewhere? The package entries are just UUIDs right - how does AEM know where to put them on install?

Issue #2165, opened 2020-01-17
ResourceResolverMapTransformer decoding lead to wrong URLs


sblano4ka:
## Required Information

* [ ] Adobe Experience Manager, Version 6.3.3.5 
* [ ] ACS AEM Commons Version: 4.1.0
* [ ] Reproducible on Latest? yes/no

### Expected Behavior

_When I open a link  that was previously encoded on purpose and that was rewritten by ResourceResolverMapTransformer, I open encoded link and link is correctly mapped and i get correct escaped '&' further handling_

### Actual Behavior

_When I open a link  that was previously encoded on purpose and that was rewritten by ResourceResolverMapTransformer, link is correctly mapped but not encoded anymore and tracked incorrectly cause value "&" is considered to be a query param separator._

### Steps to Reproduce

-  Setup ResourceResolverMapTransformer to rewrite a:href.
-  In sightly remplate add next lines
`<sly data-sly-use.model="com.my.model.Tags">
    <a href="${'{0}.html?tagId={1}' @ format=[currentPage.parent.path, model.tag]}"/>
</sly>`
-  add com.my.model.Tags :
`@Model(adaptables = SlingHttpServletRequest.class)
public class Tags {
    public String getTag() throws UnsupportedEncodingException {
        return URLEncoder.encode("technologie-&-innovationen", StandardCharsets.UTF_8.name());
    }
}`

### Links

In version 3.9.0 this case works fine,
After we upgrade ACS Commons version to  4.1.0 issue is reproduced.
The reason: adding decode functionality in ResourceResolverMapTransformer introduced after 
https://github.com/Adobe-Consulting-Services/acs-aem-commons/pull/1492 and https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/1464
According to the description, resourseResolver.map() has to encode link back. But it doesn't happened, cause it seems to encode only spaces

sblano4ka:
Initial issue is resolved.
We incorrectly use ResourceResolverMapTransformer for making short urls for all a:href. 
It was done because linkchecker didn't work for us cause extension to a:href was added by custom Transformer after linkchecker. When order of linkchecker and custom transformer was changed, issue with long lnks was resolved so no more need to  use ResourceResolverMapTransformer.

But please take into account that there is still a case when on purpose encoded query parameters will be decoded by ResourceResolverMapTransformer.

davidjgonzalez:
@sblano4ka thanks for the update.

[...]

PR #3794, opened 2026-09-02
#2165 Fix XSS via ResourceResolverMapTransformerFactory decoding query strings/fragments

henrykuijpers:
ResourceResolverMapTransformerFactory#rebuildAttributes fully URL-decoded the entire attribute value (path + query string + fragment) before calling ResourceResolver#map. That decoding was only ever needed to correctly handle pre-encoded path segments (e.g. a thumbnail rendition URI containing an encoded "jcr:content"), since ResourceResolver#map does not decode a percent-encoded path itself and would otherwise double-encode it (#1464).

The side effect was that any percent-encoded characters in the query string or fragment (e.g. %22 for a double quote) were also decoded, and since ResourceResolver#map passes the query string/fragment through unchanged, those decoded characters ended up unescaped in the rewritten HTML attribute. That allowed an attacker-controlled, percent-encoded query parameter to break out of the HTML attribute and inject arbitrary markup.

Fix: split the attribute value at the first "?" or "#" and only decode and map the path portion. The query string/fragment is now always passed through untouched, so it can never be decoded into unescaped, breakout- capable characters.

Also rewrote the unit test to use a real JCR_OAK-backed ResourceResolver (via SlingContext) instead of a hand-rolled mock, since ResourceResolver#map's own encoding behaviour is exactly what this fix depends on and needs to be verified against. The factory is now instantiated through context.registerInjectActivateService instead of manual construction plus activate(), and reflection-based access to internals was removed in favor of asserting observable behaviour. Test coverage of the class is now 100%.

### DIFF
--- a/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/ResourceResolverMapTransformerFactory.java
+++ b/bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/ResourceResolverMapTransformerFactory.java
@@ -95,30 +95,46 @@ public final class ResourceResolverMapTransformerFactory implements TransformerF
 
         for (int i = 0; i < len; i++) {
             final String attrName = newAttrs.getLocalName(i);
-            if (ArrayUtils.contains(modifiableAttributes, attrName)) {
-                final String attrValue = newAttrs.getValue(i);
-                if (StringUtils.startsWith(attrValue, "/") && !StringUtils.startsWith(attrValue, "//")) {
-                    // Only map absolute paths (starting w /), avoid relative-scheme URLs starting w //
-                    final int suffixIndex = indexOfQueryOrFragment(attrValue);
-                    final String path = suffixIndex == -1 ? attrValue : attrValue.substring(0, suffixIndex);
-                    // Never decode (or otherwise touch) the query string / fragment: ResourceResolver#map
-                    // passes it through unchanged, so decoding it here would let it carry unescaped
-                    // characters (eg. a double quote) straight into the rewritten HTML attribute.
-                    final String suffix = suffixIndex == -1 ? "" : attrValue.substring(suffixIndex);
-                    try {
-                        // The path may already contain percent-encoded characters (eg. a thumbnail
-                        // rendition path with an encoded "jcr:content"); decode it first so that
-                        // ResourceResolver#map, which encodes the path it is given, doesn't double-encode it.
-                        final String pathDecoded = new URLCodec().decode(path);
-                        newAttrs.setValue(i, slingRequest.getResourceResolver().map(slingRequest, pathDecoded) + suffix);
-                    } catch (DecoderException e) {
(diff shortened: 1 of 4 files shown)


henrykuijpers:
@kwin  @davidjgonzalez @jsedding  @rbotha78 Can we review this and get this through? :)

henrykuijpers:
@stefanseifert @cziegeler @justinedelson @joerghoh Can we review this and get this through?

bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/ResourceResolverMapTransformerFactory.java
```diff
@@ -95,23 +95,66 @@ protected Attributes rebuildAttributes(final SlingHttpServletRequest slingReques
...
-                    // Only map absolute paths (starting w /), avoid relative-scheme URLs starting w //
-                    try {
-                        final String attrValueDecoded = new URLCodec().decode(attrValue);
-                        newAttrs.setValue(i, slingRequest.getResourceResolver().map(slingRequest, attrValueDecoded));
-                    } catch (DecoderException e) {
-                        log.error("Could not decode the attribute value", e);
-                        newAttrs.setValue(i, slingRequest.getResourceResolver().map(slingRequest, attrValue));
-                    }
-                }
+            if (!ArrayUtils.contains(modifiableAttributes, attrName)) {
```
cziegeler: Why the switch to use continue?

henrykuijpers: Fair enough, initially I did this to make the indenting lesser, but since mapAttributeValue is now extracted to a method, this would not be needed anymore.

bundle/src/main/java/com/adobe/acs/commons/rewriter/impl/ResourceResolverMapTransformerFactory.java
```diff
@@ -97,21 +97,63 @@ protected Attributes rebuildAttributes(final SlingHttpServletRequest slingReques
...
+    }
+
+    private static String mapAttributeValue(final SlingHttpServletRequest slingRequest, final String attrValue) {
+        final int suffixIndex = indexOfQueryOrFragment(attrValue);
+        final String path = suffixIndex == -1 ? attrValue : attrValue.substring(0, suffixIndex);
+        // Never decode (or otherwise touch) the query string / fragment: ResourceResolver#map
+        // passes it through unchanged, so decoding it here would let it carry unescaped
+        // characters (e.g., a double quote) straight into the rewritten HTML attribute.
+        final String suffix = suffixIndex == -1 ? "" : attrValue.substring(suffixIndex);
+        return mapPath(slingRequest, path) + suffix;
```
cziegeler, line 123: I personally like String.concat() better,

cziegeler (approved):

